PR #11, opened 2015-01-16
Avoid unnecessary recompilation with two identically named files in sub folders

tomasnilefrost:
If a/foo.vhd and b/foo.vhd are added to the same library one of them will always be recompiled since their vunit_hash-files collide.

This is a problem when using vunit in a Microsemi environment since all generated FIFO IP-cores are supposed to be in the same library and each FIFO contains identically named vhd-files.


kraigher:
Thank you.

PR #20, opened 2015-02-17
Remove tcl flag. Closes #16 for Microsemi 10.3a

### LINKED ISSUE #16: The flag -modelsimini is necessary on Modelsim Microsemi v10.3a
The fix for Issue #12 (5917f349810b968dc8aac307eabf8429a40c1d01) breaks vunit for "Model Technology ModelSim Microsemi vsim 10.3a Simulator 2014.04 Apr 15 2014"

After the fix it is impossible to load a design since vsim cannot find the compiled libaries (cannot find the modelsim.ini-file)

By reintroducing the flag -modelsimini in vunit_load() the majority of the acceptance tests can be run successfully.
However, this breaks issue #12 since -modelsimini cannot be specified correctly with spaces in the path.

tomasnilefrost:
The fix (0bcfe8f26587) does not remove the MODELSIM flag from the tcl scripts.
With Microsemi 10.3a this leads to MODELSIM being unset instead of correctly set from the init()-call.

### DIFF
--- a/vunit/modelsim_interface.py
+++ b/vunit/modelsim_interface.py
@@ -135,8 +135,11 @@ proc vunit_load {{}} {{
     {set_generic_str}
 
     # Workaround -modelsimini flag not respected in some versions of modelsim
-    global env
-    set env(MODELSIM) "{modelsimini}"
+    # however, Microsemi 10.3a corrupts the enviromnent variable (see dvt64978)
+    if {{[string first "Microsemi vsim 10.3a" [vsimVersionString]] eq -1}} {{
+        global env
+        set env(MODELSIM) "{modelsimini}"
+    }}
     vsim -wlf "{wlf_file_name}" -quiet -t ps {pli_str} {set_generic_name_str} {library_name}.{entity_name}{architecture_suffix}
     set no_finished_signal [catch {{examine -internal {{/vunit_finished}}}}]
     set no_test_runner_exit [catch {{examine -internal {{/run_base_pkg/runner.exit_without_errors}}}}]


kraigher:
@tomasnilefrost 
I am a bit reluctant to merge this since the user.do file will no longer work as-is.
I do not understand why your version of modelsim would not work in this case? I thought it ignored the environment variable when set from the tcl-script? If was not ignored it would be set to the same value as the parent python process. Are you saying it actually corrupts the value?

tomasnilefrost:
My initial testing yielded that the MODELSIM variable was unset when also set from the tcl-script.

I can fix some output tomorrow, but it seems that set in the tcl script is corrupting the environment variable.

kraigher:
@tomasnilefrost I wonder what convoluted code they must have in Modelsim to cause so many strange bugs arround modelsimini. In the worst case we might need to add some special case handling of this when microsemi modelsim 10.3a is detected based on the "vsimVersionString" tcl-command.

tomasnilefrost:
It's quite wierd to say the least.

I modified the tcl-script as follows:
printenv MODELSIM
global env
set env(MODELSIM) "{modelsimini}"
printenv MODELSIM

The first printenv yields the correctly set value from the python process and the second printenv yields: # can't read "env(MODELSIM)": no such variable.

So it seems broken in some way.

However, the real kicker is this:
printenv
global env
set env(MODELSIM) "{modelsimini}"
printenv

Still yields the same result, the second printenv doesn't contain the MODELSIM variable, but somehow the first set MODELSIM-variable is still used (without being visible after the second call).

So I would say that it is something fundamentallly wrong with the Microsemi implementation.
Replacing the first printenv with printenv MODELSIM does not seem to work the same way and the variable is completely lost.

It's an ugly hack to slap on a printenv before setting the environment variable but does it also work in other Modelsim environments?

kraigher:
@tomasnilefrost My recommended solution is to add an if-statement arround the setting of the MODELSIM environment variable in the tcl script based on the result of the "vsimVersionString" command.

tomasnilefrost:
I'll look into this tomorrow again and update my pull request.

tomasnilefrost:
@kraigher Would you mind running the acceptance tests on your version of Modelsim?

kraigher:
Nice work. Thank you.

PR #29, opened 2015-02-25
Add component instantiations to the dependency graph

KevinKes:
The compile order is now also using dependencies of any instantiated component

### DIFF
--- a/vunit/project.py
+++ b/vunit/project.py
@@ -77,7 +77,7 @@ class SourceFile:
             design_file = VHDLDesignFile.parse(code)
             self.design_units = self._find_design_units(design_file)
             self.dependencies = self._find_dependencies(design_file)
-            self.depending_components = design_file.components
+            self.depending_components = design_file.component_instantiations
         else:
             self.design_units = []
             self.dependencies = []
--- a/vunit/project.py
+++ b/vunit/project.py
@@ -240,7 +240,7 @@ class Project:
                 try:
                     primary_unit = library.primary_design_units[unit_name]
                 except KeyError:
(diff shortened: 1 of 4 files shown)


kraigher:
@KevinKes There are a number of problems with the pull request:

1. I has merge conflicts with master. You need to rebase.
2. I also prefer if you would squash the changes into a single commit such that the master history is cleaner and does not contain to many intermediate development stages.
3. I do not like the fix for the comment I made about the warnings for components not found in library.
I would prefer if a debug message was triggered only when the component has not been found in *any* library rather than for *each* library that it has not been found. 
     * NOTE: Even though in theory a warning would be more correct than a debug message, in many practical cases components are used for stuff within pre-compiled black-box libraries and it would be annoying to get a warning in those cases.
4. The indentation is wrong so I cannot even run it:
```shell
  File "examples/user_guide/run.py", line 14, in <module>
    from vunit import VUnit
  File "examples/user_guide/../../vunit/__init__.py", line 7, in <module>
    from vunit.ui import VUnit
  File "examples/user_guide/../../vunit/ui.py", line 21, in <module>
    from vunit.project import Project
  File "examples/user_guide/../../vunit/project.py", line 96
    for component in self.depending_components:
                                              ^
TabError: inconsistent use of tabs and spaces in indentation
```

KevinKes:
@kraigher Auch, sorry for my inexperience with git and python. I have to figure out how I can 'rebase' and how I can 'squash the changes into a single commit'. I will fix item 3 and 4.

KevinKes:
I took the 'easy' way and just deleted the fork with a backup of the changes. Now they are squashed into one commit :) . I will create a new pull request.

PR #30, opened 2015-02-26
Add component instantiations to the dependency graph

KevinKes:
This one does not have merge conflicts and should run OK.

### DIFF
--- a/vunit/project.py
+++ b/vunit/project.py
@@ -76,12 +76,14 @@ class SourceFile:
 
         self.design_units = []
         self.dependencies = []
+        self.depending_components = []
 
         if self.file_type == 'vhdl':
             try:
                 design_file = VHDLDesignFile.parse(code)
                 self.design_units = self._find_design_units(design_file)
                 self.dependencies = self._find_dependencies(design_file)
+                self.depending_components = design_file.component_instantiations
             except:
                 traceback.print_exc()
                 logger.error("Failed to parse %s", name)
@@ -93,6 +95,12 @@ class SourceFile:
                 logger.debug('Adding secondary design unit (package body) for package %s', design_unit.primary_design_unit)
             else:
                 logger.debug('Adding secondary design unit (%s) %s', design_unit.unit_type, design_unit.name)
+        
+        if len(self.depending_components) != 0:
+            logger.debug("The file '%s' has the following components:", self.name)
+            for component in self.depending_components:
+                logger.debug(component)
+        else: logger.debug("The file '%s' has no components", self.name)
 
     def _find_dependencies(self, design_file):
         # Find dependencies introduced by the use clause
@@ -107,6 +115,7 @@ class SourceFile:
             if library_name == "work": # Work means same library as current file
                 library_name = self.library.name
             result.append((library_name, entity))
+                        
         return result
 
     def _find_design_units(self, design_file):
@@ -165,10 +174,11 @@ class DesignUnit:
 
 
 class Project:
-    def __init__(self):
+    def __init__(self, depend_on_components=False):
         self._libraries = {}
         self._source_files = {}
         self._source_files_in_order = []
+        self._depend_on_components = depend_on_components
 
     def _validate_library_name(self, library_name):
         if library_name == "work":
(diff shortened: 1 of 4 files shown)


kraigher:
@KevinKes One thing struck me when I reviewed this change. Component binding occurs at elaboration time and it does not matter in which order the files are compiled thus the extra dependencies injected by this change will be unnecessary. What do you think?

vunit/project.py
```diff
@@ -229,6 +238,24 @@ def _find_other_design_unit_dependencies(self, source_file):
             else:
                 yield primary_unit.source_file
 
+    def _find_component_design_unit_dependencies(self, source_file):
+        
+        for unit_name in source_file.depending_components:
+            foundComponentEntity = False
```
kraigher: camelCase variable name violates PEP8 coding style which we try to follow.

vunit/test/test_project.py
```diff
@@ -391,6 +442,16 @@ def assert_has_architecture(self, source_file_name, name, entity_name):
                                       "architecture",
                                       name, False, entity_name)
         self.assertIsNotNone(unit)
+		
+    def assert_has_component(self, source_file_name, component_name):
+        foundComp = False
```
kraigher: camelCase variable name violates PEP8 coding style which we try to follow.

KevinKes:
@kraigher I think you are right. Some background on the reason I was implementing this.

Using Sigasi as a VHDL editor, I was looking for an easy method to do a Modelsim compile of the sources automatically in a nightly build. Secondly I would like to dynamically add all synthesis files to my synthesis project (which also needs to be in the right compilation order). As the synthesis tool is in most cases not able to synthesize the test benches, I would like to filter the source tree from a specific (synthesis) toplevel entity.

An average FPGA project contains a lot (~200) of files and I don't want to manually create the compile order. Sigasi offers the possibility to export a csv file with the compilation order of a project, but as a developer I can forget to update his file when I commit changes. Also automating Sigasi to export this file is not a nice solution. 

In my search for unit testing and VHDL I stumbled upon VUnit, providing a Python script to find the compilation order! Exactly what I was looking to automate the nightly, without updating a csv file with a compilation order.

I combined the VUnit VHDLparser and project with a parser for the Sigasi '.library_mapping.xml' file to get a complete FPGA project compilation for Modelsim, which works perfectly. (maybe this is interesting for your Sigasi integration?)

For the synthesis project I constructed a function to use the VUnit dependency functionality to get a compilation order based on a given top entity. Based on this top entity, the function loops trough all dependencies recursively to find all depending files. As some of the design files where using component instantations, these where not taken into account and I ended up adding a function to add components as a dependency.

Not sure if anything of this can be useful for you/VUnit?

kraigher:
@KevinKes What synthesis tool are you using that needs you to specify the compile order? Both Xilinx ISE and Vivado just lets you add the files to the project via tcl and it figures out the compile order automatically. In my project we use Vivado and we generate the synthesis project from Python filtering out the testbenches which are always located in a test-subfolder and we do not need to know the compile order to do this. We also have Sigasi integration already but we generate the Sigasi project from the VUnit project.py rather than parsing the Sigasi-file. 
Maybe it would be useful for project.py to have this feature you have implemented for your situation even though it is not necessary for simulation. It could be enabled/disabled by a parameter to the __init__ constructor for instance.

kraigher:
@KevinKes Just to clarify: do you use the Project-class directly or do you use it through the VUnit-class?

kraigher:
@KevinKes My suggestion is to add this feature to project.py but keep it disabled by default so that no unecessary recompilation is performed when just simulating. You can enable it by parameter to the project constructor when you use it for synthesis . Works for you?

KevinKes:
@kraigher Using Synopsis SynplifyPro here, it is able to determine the compile order, but you have to specify at least the top entity and filter out the test bench files. Interesting approach to generate the Sigasi project. 

As a developer the Sigasi project is the 'cock pit' of my design work. The build server then uses this project as input to perform a nightly build / simulation run. I also set up an external tool configuration (the green play button in the sigasi/eclipse toolbar) to update the synthesis project file, based on the Sigasi project.

I currently do import "vunit.project", so assume that I use it directly? :)

KevinKes:
@kraigher Yes this should work perfect for me.

kraigher:
@KevinKes If you add the constructor parameter to Project to enable this that is off by default and fix the PEP8 violations I will merge it. Don't forget to rerun the unit tests after adding the parameter.

kraigher:
@KevinKes We infer the vhdl library structure from the folder structure thus no vhdl file is individually mentioned in any file list. Typically we have src/library_name/*.vhd and src/library_name/test/*.vhd

KevinKes:
@kraigher I updated the commit with the init constructor and fixed the PEP8 violations.

We are also thinking of a directory structure similar to yours. But i'm wondering: why do you generate the sigasi project file? Is it only generated once, at the start of your design?

kraigher:
@KevinKes We re-generate the sigasi project as soon as we add any new file.

kraigher:
@KevinKes I merged the pull request now. Thank you for your contribution.

PR #34, opened 2015-03-20
Removed some unused declarations.

philippefaes:
I've removed some unused declarations (dead code) from the "vunit_lib" code. I'm not touching "osvvm", because I suppose you will want to track the upstream releases of osvvm.

### DIFF
--- a/examples/array/src/test/tb_sobel_x.vhd
+++ b/examples/array/src/test/tb_sobel_x.vhd
@@ -19,7 +19,6 @@ use osvvm.RandomPkg.all;
 entity tb_sobel_x is
   generic (
     runner_cfg : runner_cfg_t;
-    output_path : string;
     tb_path : string);
 end entity;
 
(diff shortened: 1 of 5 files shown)


kraigher:
I notice that you actually did touch "osvvm". As you said we want to just redistribute osvvm without modification so It might be best to leave it alone even if it contains unused variables and such. 
 vhdl/osvvm/AlertLogPkg.vhd
vhdl/osvvm/CoveragePkg.vhd

kraigher:
@philippefaes Any update on this pull request? I do not want to merge it with the changes to the osvvm-files. If you remove those changes I would gladly merge it.

philippefaes:
I understand and I agree that you need a better patch. I'm just tied up
with other work right now. I might get around to submitting a new patch
later.


-- 
Philippe Faes
Founder and CEO, Sigasi
http://www.sigasi.com
T: +32 9 265 71 38
M: +1 (408) 627 7691

PR #35, opened 2015-03-20
Removed references to std_logic_unsigned.

philippefaes:
I've removed references to std_logic_unsigned.
    
This library is not officially standardized and has known problems. This library does not really use std_logic_unsigned, so safe to remove.

### DIFF
--- a/vhdl/check/src/check.vhd
+++ b/vhdl/check/src/check.vhd
@@ -8,7 +8,6 @@
 
 library ieee;
 use ieee.std_logic_1164.all;
-use ieee.std_logic_unsigned.all;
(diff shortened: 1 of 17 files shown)


kraigher:
@philippefaes I should let you know that we have a test as part of the VUnit-regression suite that checks the license header in every file in particular the year range. When a file only had copyright in year 2014 and it was changed again in 2015 then range should be 2014-2015. The git log is used to check this. This test failed after this merge but I fixed it for you. The test is now part of the unit test suite which can be run with 'python -m unittest discover vunit/' in the repo root. This check should be run with everything checked in and without any unstaged files.

Issue #23, opened 2015-02-18
Add communication mechanism


LarsAsplund:
When I presented VUnit at FPGA Forum i Norway last week I showed a slide (2:40 into [this presentation](https://www.youtube.com/watch?v=D8s_VLD91tw)) with a VUnit testbench example. The example showed a testbench with mixed levels of complexity. The unit under test was partly controlled by pin-wiggling and partly controlled with more high-level transactions. The point was to show that VUnit does not care about the complexity of your testbench, it can be simple or advanced. However, after the presentation there was a number of question around the underlaying message passing mechanism I used for the transactions. Is it part of VUnit? Is it for free? There were also discussions around how streams of transactions should be organized in a testbench. Should all stream be controlled from a central process (I did that using the test runner as the central process) or should there be (many) smaller parallel processes controlling one stream/interface each. Regardless of the answer there was an interest for a reusable communication mechanism that could support transactions but also other communication design patterns. This would not be a part of the "VUnit core" but more of a good to have utility for testbench development. Unit testing is at the core of the VUnit project but our project mission is open to best practises in general so I think such a mechanism has a place. One way forward would be to simply push the code I currently have but I know that there are more similar ideas within the community so I would like to make an experiment and take one step back and let the community design and implement this functionality. I'm thinking an approach similar to this

1. Decide __WHAT__ the community wants from a generic communication solution. We do this with user stories
2. The user stories will give us a vocabulary of objects and how they relate to each other. Sender, receiver, packet, transaction, message are examples of possible objects.
3. With the _domain model_ above we are ready to refine our model into sequences of method/procedure/function calls that build our scenarios and we can set a number of VUnit test cases that we want to meet.
4. At this point we can start to get into the details figuring out __HOW__ things can be done to fulfill these test cases. So I'm thinking test-driven design where we set the test cases first.

I realize that it's sometimes easier to exemplify how things can be done in order to describe what you want but let's try to at least put the initial emphasis on __WHAT__. Sounds like a plan?

I think we should start the discussion in this issue to get some attention but at some point, maybe after the user stories, we should move the design to a wiki page.

I will start with a few user stories and let you fill in and then we will see what happens.

I'm not sure about your experiences with user stories and I guess there are tons of rights and wrongs (probably overlapping). However, I'm used to the following format

As a \<user\>, I want \<something\> so that \<the reason\>.

[...]

PR #42, opened 2015-05-05
Added a communication mechanism as discussed in issue #23

LarsAsplund:
Added placeholder for communication package and an almost empty testbench.

Added empty create_actor function and a few failing test cases.

Added test cases for find and destroy.

Implemented create, find, and destroy

Refactoring of communication mechanism.

Added test cases for doing a find without deferred creation in case of missing actor and for asking for the number of deferred creations. These functions can reduce the risk associated with trying o find a misspelled actor.

Implemented num_of_deferred_creations.

Removed unused deferred_creation function.

Added test cases for normal client/server operation.

Implemented client/server test cases.

Added test case for the publisher/subscriber pattern.

Added more test cases and implementation for the client/server communication pattern

Merged all status types into one.

Added example for the communication mechanism. Unit under test is not implemented, just the unit tests.

Moved receive status into the message to reduce a variable declaration for a variable often not used

Added support for synchronous request/reply

Added support for a bounded actor inbox. Also changed behavior when an already existing actor is created again. Instead of returning the existing actor a null actor is returned.

Added support for encoding and decoding of VHDL built-in types as well as ieee standardized types. Also added support functionality for writing custom codecs.

Added support for automatic generation of codecs for enumeration and record types

Added codec generation for array types

Performance enhancements on memory allocations, notify mechanism, and codec grouping

Created more dense and less readable encodings to increase performance

Updated run scripts to use add_com() in order to enable the com functionality

Separated codec templates into its own file.

Made it possible to decode message type with a single command when several message type types are present.

Packing two bits per character when encoding std_ulogic_vector to improve performance.

Refactoring

Made a behavioral model for card shuffler to make example test bench pass

Updated license test to ignore generated codecs.

Added com tests to acceptance test suite

### DIFF
--- a/vunit/ui.py
+++ b/vunit/ui.py
@@ -37,7 +37,9 @@ from vunit.check_preprocessor import CheckPreprocessor
 from vunit.vhdl_parser import CachedVHDLParser
 from vunit.builtins import (add_builtins,
                             add_array_util,
-                            add_osvvm)
+                            add_osvvm,
+                            add_com)
+from vunit.com import codec_generator
 
 import logging
 LOGGER = logging.getLogger(__name__)
@@ -84,6 +86,7 @@ class VUnit(object):  # pylint: disable=too-many-instance-attributes
 
         return cls(output_path=args.output_path,
                    clean=args.clean,
+                   use_debug_codecs=args.use_debug_codecs,
                    no_color=args.no_color,
                    verbose=args.verbose,
                    xunit_xml=args.xunit_xml,
@@ -135,6 +138,10 @@ class VUnit(object):  # pylint: disable=too-many-instance-attributes
                             default=False,
                             help='Remove output path first')
 
+        parser.add_argument('--use-debug-codecs', action='store_true',
+                            default=False,
+                            help='Run with debug features enabled')
+
         parser.add_argument('-o', '--output-path',
                             default=join(abspath(getcwd()), "vunit_out"),
                             help='Output path for compilation and simulation artifacts')
@@ -171,6 +178,7 @@ class VUnit(object):  # pylint: disable=too-many-instance-attributes
     def __init__(self,  # pylint: disable=too-many-locals, too-many-arguments
                  output_path,
                  clean=False,
+                 use_debug_codecs=False,
                  no_color=False,
                  verbose=False,
                  xunit_xml=None,
@@ -210,6 +218,7 @@ class VUnit(object):  # pylint: disable=too-many-instance-attributes
         self._external_preprocessors = []
         self._location_preprocessor = None
         self._check_preprocessor = None
+        self._use_debug_codecs = use_debug_codecs
 
         if simulator_name is not None:
             self._simulator_name = simulator_name
@@ -466,6 +475,14 @@ class VUnit(object):  # pylint: disable=too-many-instance-attributes
     def _preprocessed_path(self):
         return join(self._output_path, "preprocessed")
 
+    @property
+    def codecs_path(self):
+        return join(self._output_path, "codecs")
+
+    @property
+    def use_debug_codecs(self):
+        return self._use_debug_codecs
+
     def _create_tests(self, simulator_if):
         """
         Create the test suites by scanning the project
@@ -508,6 +525,19 @@ class VUnit(object):  # pylint: disable=too-many-instance-attributes
         library = self.add_library(library_name)
         add_builtins(library, self._vhdl_standard, mock_lang, mock_log)
 
+    def add_com(self, library_name="vunit_lib", use_debug_codecs=None):
+        """
+        Add communication package
+        """
+        if not self._project.has_library(library_name):
+            library = self.add_library(library_name)
+        else:
+            library = self.library(library_name)
+        if use_debug_codecs is not None:
+            self._use_debug_codecs = use_debug_codecs
+        add_com(library, self._vhdl_standard,
+                use_debug_codecs=self._use_debug_codecs)
+
     def add_array_util(self, library_name="vunit_lib"):
         """
         Add array utility package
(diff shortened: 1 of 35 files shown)


vunit/ui.py
```diff
@@ -548,6 +578,31 @@ def set_pli(self, value):
     def add_source_files(self, pattern, preprocessors=None):
         self._parent.add_source_files(pattern, self._library_name, preprocessors)
 
+    def generate_codecs(self, package_name, codec_package_name=None, used_packages=None, output_file_name=None):
```
kraigher: It would be nicer if this followed the same object oriented style as  is achieved with the `EntityFacade` where you can write `lib.entity("foo").something()` such as:
```python
lib.package("foo").generate_codecs(..)
```

vunit/com.py
```diff
@@ -0,0 +1,350 @@
...
+"""
+Module for generating VHDL com codecs.
+"""
+
+from vunit.vhdl_parser import VHDLEnumerationType, VHDLPackage
+from vunit.ostools import read_file, write_file
+from vunit.codec_templates import StdCodecTemplate, DebugCodecTemplate, CodecTemplate
+
+
+class CodecGenerator(object):
```
kraigher: This is a pointless class. It is basically just a function in a class.

kraigher:
I also have a few general comments:

1) The commit message could be shortened. It is pointless to include the commit message of every commit you squashed especially "Refactoring". It just makes for an unclean repository history

2) The codec_templates.py and com.py would be better put in a sub-folder named "com" (dont forget to add an __init__.py). com.py could also be renamed to something more meaningfull such as "codec_generator.py". A good Python to convention is to have every major class name ClassName in a separate file called class_name.py.

3) I am not that happy about the new command line argument for debug codecs since it is an optional feature that is supposed to be modular. I would like it more if the it was not there when not using the com package. Maybe we need to add some kind of pre-argparse configuration for this to be possible. Anyway I wont argue about it if you insist on having it.

kraigher:
@LarsAsplund I have pushed two fixes of the comments that I have made to the `com-fixes` branch.
First I added a PackageFacade to ui.py to match the style of the Entity-related interfaces. Secondly I moved the com python files to a separate com folder and renamed com.py codec_generator.py.

kraigher:
@LarsAsplund In generall a few explanatory comments in the VHDL code would not hurt, especially the example. One thing that could be worth mentioning is how memory leaks are prevented. I first thought you had a memory leak in the `tb_card_shuffler.vhd` but then realized that messages was automatically deleted on the next receive call. This is convenitient but if someone uses a procedure to implement the body of the receive-loop and uses a local variable in that procedure then an explicit delete would have to be used since the varaible is not the same for each receive call.

kraigher:
@LarsAsplund I think I am finished with the review now. If you cherry-pick the com-fixes and add some VHDL comments / documentation about the com package and examples I really have nothing to complain about. Nice work!

kraigher:
@LarsAsplund FYI to update a pull request just push or force push to the branch which you set as the source of the pull request and GitHub will know it was updated.

vunit/ui.py
LarsAsplund: Agree, I'll have a look at that.

vunit/com.py
LarsAsplund: Agree, it collapsed in the last refactoring. I'll remove it.

LarsAsplund:
@kraigher 
* I'll clean up the commit message. What you see is just the default one created by the squash
* I thought about the submodule as well. Parts in the derived CodecVHDLPackage (which I think should be named something like DatatypeVHDLPackage) should actually be in separate classes derived from VHDLArrayType, VHDLEnumerationType, ... That would result in even more files
* I will think about the alternatives to --use-debug-codecs. I'd like to see how the debug codecs will be used, not at all, when searching for bugs or always. Depending on that the option will be more or less useful.
* Yes, more comments are needed (there's no linting tool that forces me...). I'm planning for a user guide which will be supported by the examples. That will give some of the documentation you're asking for
* I'll get back to you when the pull request is updated

vunit/ui.py
kraigher: I have already fixed it for you in the `com-fixes` branch. Use that and save some work.

LarsAsplund: Yes I saw that and picked  those commits

tomasnilefrost:
@LarsAsplund Great work! Will be quite fun to start using this feature when it reaches develop.

One sidenote, do you think that you can run a beautifier on the vhdl-source-code and choose whether to align ":" of the arguments/port-signals or not. As of now it is quite mixed, differs between functions as well as within functions. It should be easier to perform this on the feature branch rather than when the feature is merged.

kraigher:
@tomasnilefrost @LarsAsplund There is also a lot of trailing whitespace in the vhdl code. I have a save hook in emacs that removes trailing whitespace and a plugin which highlights it in red.

LarsAsplund:
@kraigher  @tomasnilefrost  I'll do the beautify before next pull request. Olof, could you point out the emacs plugins for me?

kraigher:
@LarsAsplund Put this in your `.emacs` file:
```elisp
; Remove trailing whitespace on saving file.
(add-hook 'before-save-hook 'delete-trailing-whitespace)

; Highlight trailing whitespace and long lines
(require 'whitespace)
(setq whitespace-style '(face empty tabs lines-tail trailing))
(setq whitespace-line-column 119)
(global-whitespace-mode t)

```

kraigher:
@LarsAsplund I have successfully run all the tests on Linux using Python 3 and Python 2.7.

Btw I ran GHDL on it as well, after removing all contexts and it finds some issues with the code. Mainly reading of output parameters of procedures notify, reply and decode. Also it thinks that the max_timeout_c constant of 1000 hr is overflowed. Also the procedure clear_reply_stash had an "in" keyword in the body but not the definition causing an error.

Anyway I think the GHDL errors are not enough to stop the pull request so go ahead. I am satisfied now. We can fix the GHDL errors in later commits.

LarsAsplund:
@kraigher Ok, I will look at those issues but make the merge before. It feels better to get the feature off its marathon branch and move it into maintenance.

PR #47, opened 2015-05-29
Add description_from_user to VUnitCLI

LukasVik:
If your "run.py" is more than a simple interface to VUnit its a bit silly if it prints only ”VUnit command line tool.” when you do --help.

If you agree you can take my patch as it is, or feel free to implement it in your own way.


kraigher:
I cherry-picked the commit manually since I wanted to make a slight change. 5152d2f240b3e093ef1c72ad8f272ef2d8020cef

PR #64, opened 2015-07-29
tells the destinaton of the compilation

vermaete:
A small one.  Before VUnit, we had one Makefile for every library.  Now, it's all in one run.py script.  And without this small modification it was hard to find out from the logging into which library the VHDL was compiled into.


kraigher:
I manually merged this also adding the same print to the ghdl_interface.py. Thank you.

Issue #67, opened 2015-08-27
Dependency generation and output


joshrsmith:
VUnit already has baked in support for managing VHDL projects (collections of files), as well as parsing VHDL enough to generate dependency graphs. It would be useful to be able to obtain a compile-order dependency tree from VUnit, without taking further action (compilation, simulation, etc). It would be reasonable to assume that you don't need a simulator at all to do this. All the functionality is already baked into VUnit.

Possible use cases:
1. Output *all* project files in compile order
2. Output only project files that a specified file / design unit / node depends on, in compile order

Output could be directed to a user specified file.
Outputting the library that each file belongs to would also be helpful.

This ordered output file would be generic enough for someone to use automation for other purposes, other than unit testing via VUnit.

I recognize this might run orthogonal to the core goals of VUnit, so perhaps it should only be considered as an ancillary utility (alternate entry point) that uses vunit library functionality.

LarsAsplund:
I've thought about this as well. Seeing a dependency graph can be a way to understand the structure of the code.  That can help you understand what needs to be improved, dependencies that can/should be removed and so on. What is your story? I'm interested too see if that story is something that VUnit should support rather than just the intermediate step.

kraigher:
I can agree that having a flag to dump the project hierarchy and dependencies as a means of debugging via the main VUnit command line interface would be useful. For more advanced uses I such as further scripting based on the dependency graph I would prefer users to use VUnit as a library. Using it as a library is already possible, the relevant module is vunit.project which has a Project class. I know of at least one guy #27 that already uses project.py/Project as a library.

I will just note that our current dependency mechanism can determine the dependencies of just a single file but not, at least in theory, the dependencies of a single design unit. Since we do not create complete parse trees but rather just extract fragments such as entity instantiations we do for instance not currently keep track of which architecture body in a file (out of many) that an entity instantiation belongs to. This is a simplification that works for our needs. A good design practice is to just have one design unit per file though and then this simplification will not have any effect.

LarsAsplund:
Maybe we can make the dump using some highly readable but yet defined graph format. An example is the [trivial graph format](https://en.wikipedia.org/wiki/Trivial_Graph_Format) looking like this.

    1 vhdl/foo/src/bar.vhd
    2 vhdl/spam/src/eggs.vhd
    3 vhdl/baz/src/qux.vhd
    #
    1 2 vhdl/foo/src/bar.vhd depends on vhdl/spam/src/eggs.vhd
[...]

PR #68, opened 2015-08-28
Example of dependency generation

joshrsmith:
Here is an implementation that I think is close to what we were discussing earlier in issue #67. I am (still) not 100% sold on how to roll this functionality into the public interface of VUnit. You can see what I did below as an example. I created a trivial example as well.

While I was working on this, I think I also found a minor bug in the dependency_graph.py. An edge case if there were no dependencies to return. Check it out.

### DIFF
--- a/examples/vhdl/compile_order/compile_order.vhd
+++ b/examples/vhdl/compile_order/compile_order.vhd
@@ -1,0 +1,12 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this file,
+-- You can obtain one at http://mozilla.org/MPL/2.0/.
+--
+-- Copyright (c) 2014-2015, Lars Asplund lars.anders.asplund@gmail.com
+
+entity compile_order is
+end entity compile_order;
+
+architecture compile_order_arch of compile_order is
+begin
+end architecture compile_order_arch;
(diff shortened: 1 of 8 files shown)


kraigher:
I took you pull request and fixed some problems while also making some modifications. You can review my changes in the [pull68](https://github.com/LarsAsplund/vunit/commits/pull68) branch.

What I did:
* There were some trailing whitespace problems which failed the lint checks we have. 
* The dependency method you added would only include direct dependencies and not also indirect dependencies. I fixed this and added tests for it as well.
* I re-structured the compile_order example by moving the project creation logic to a separate file and also created a separate `print_compile_order.py` script with the sole purpose of printing the compile order and let the `run.py` file avoid doing this and retain its original purpose. The project creation could be re-used between those two scripts. This illustrates that ancillary functions can be easily moved to separate scripts without parasiting on `run.py` while avoiding code duplication. 

I will let you have a look and see if you have any comments and then I will integrate it into the `master` branch.

joshrsmith:
I think I understand most of the changes you made.
I knew I was going to have problems with pep/lint because I was having trouble getting it running on my machine. I think I have some files in side my Vunit project directory that was confusing things. If I contribute anything else, I will obviously have to fix that.

So it seems like run.py still does what run typically does, and now there is a separate script, which serves as an example for printing out this new information. Great.

I won't get a chance to check through this thoroughly until later today.

kraigher:
@joshrsmith I pushed the changes to master now. Thank you for your contribution.

PR #70, opened 2015-08-31
Allow HDL files with various case extensions (upper, lower, mixed)

joshrsmith:
I have run into a problem using VUnit where some of the files are using different cases than directly supported. For example, *.VHD instead of .vhd, or more horrifically, .VHDL  (who does this?)

Unfortunately, it is not as easy as renaming the file. Some CM tools (e.g. Perforce) when hosted on a server that is case insensitive (e.g. Windows) does not allow changing the case. However the file was first checked in will be how it will forever stay. Sure, you can change it in your local workspace, but next time you checkout a fresh copy... bad news.

This fix simply forces the extension compatibility check to use a lower case representation of the input file extension. This should allow any case of extensions to be supported.

I wasn't sure how you may want to refactor the extensions out from being hard-coded in a function, so I took a stab.

I couldn't find any other cases where the case may matter, but I may have missed somewhere.
I added a couple unittest cases to check for various capitalization. I also added a testcase for a "bad" file extension. I am not sure if this was covered somewhere else already, but I figured it can't hurt.

### DIFF
--- a/vunit/test/acceptance/test_ui.py
+++ b/vunit/test/acceptance/test_ui.py
@@ -10,13 +10,13 @@ Acceptance test of the VUnit public interface class
 
 
 import unittest
-from string import Template
+from string import Template, upper, lower
 from tempfile import NamedTemporaryFile
 from os.path import join, dirname, basename, exists
 from os import remove
 import re
 from re import MULTILINE
-from vunit.ui import VUnit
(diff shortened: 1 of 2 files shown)


kraigher:
I have manually merged this into master with some very minor modifications. Thank you for your contribution.

PR #132, opened 2016-02-22
Update xunit export

go2sh:
Update the xunit report generation to support the classname attribute in a testcase. Therefore testbenches with no runs will get the testname "all".

### DIFF
--- a/vunit/test/acceptance/artificial/vhdl/tb_magic_paths.vhd
+++ b/vunit/test/acceptance/artificial/vhdl/tb_magic_paths.vhd
@@ -2,7 +2,7 @@
 -- License, v. 2.0. If a copy of the MPL was not distributed with this file,
 -- You can obtain one at http://mozilla.org/MPL/2.0/.
 --
--- Copyright (c) 2015, Lars Asplund lars.anders.asplund@gmail.com
+-- Copyright (c) 2015-2016, Lars Asplund lars.anders.asplund@gmail.com
 
 library vunit_lib;
 context vunit_lib.vunit_context;
@@ -30,7 +30,7 @@ begin
   begin
     test_runner_setup(runner, runner_cfg);
     check_has_suffix(tb_path, "acceptance/artificial/vhdl/");
-    check_has_suffix(output_path, "acceptance/artificial out/tests/lib.tb_magic_paths/");
+    check_has_suffix(output_path, "acceptance/artificial out/tests/lib.tb_magic_paths.all/");
     test_runner_cleanup(runner);
     wait;
   end process;
--- a/vunit/test/acceptance/test_artificial.py
+++ b/vunit/test/acceptance/test_artificial.py
@@ -44,7 +44,7 @@ class TestVunitArtificial(unittest.TestCase):
 
         elab_expected_report = []
         for status, name in EXPECTED_REPORT:
-            if name in ("lib.tb_elab_fail",):
+            if name in ("lib.tb_elab_fail.all",):
                 status = "failed"
             else:
                 status = "passed"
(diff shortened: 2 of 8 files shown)


kraigher:
I am not sure we want to change how the actual tests are named on the VUnit CLI rather than just the JUnit XML. If we add this *all* suffix to those test benches it will change the names of many tests for people using VUnit. I need to think about it some more.

If merging this pull request we also have to update the user guide on the website.

go2sh:
Yeah, I also thought about that. It came up when I changed all the tests. The coolest thing would be to create a TestCase class rather then just some strings. Then the required information for every task could be extracted.

kraigher:
@LarsAsplund What do you think. Should we add a `.all` to test benches where there are no explicit tests just to be compatible with XUnit reports?

LarsAsplund:
@kraigher The Jenkins test result certainly looks nicer with the proposed change but since it affects many users I'd like a second opinion. I will contact two VUnit + Jenkins users and see what they think.

LarsAsplund:
@kraigher I've spoken with one of two projects I know that uses the [Parallel Test Executor plugin](https://wiki.jenkins-ci.org/display/JENKINS/Parallel+Test+Executor+Plugin) for Jenkins. They will test but thinks it will probably be easier since this plugin requires a class name and today they do post processing of the VUnit XML do achieve that

TimmyBrolin:
The Parallel Test Executor plugin has several problems. It is obvious that it was written only with java testing in mind.
We have extended the plugin with more flexibility to support other tests, such as VUnit, with no processing of any files needed. It is available here:
https://github.com/jenkinsci/parallel-test-executor-plugin/pull/15
It is supposed to be merged with the official plugin. Not sure why it is taking so long.
@LarsAsplund Perhaps you can forward this link to those who are currently doing post processing. They will probably find it useful. And additional fieldtesting of our modifications would not hurt.

Regarding the suggested addition of test classes, I believe it should work OK for us. As I understand it, the currrent "test names" would basically become "test classes" instead?

go2sh:
Not exactly. The library + testbench + (archictecture) becomes the classname and the testname becomes the name.
The problem with the testname is that only test benches with run strings in it have dedicated testnames, testbenches without run strings have no testname. This PR adds a generic testname ("all") to thoses testbenches and sets the xunit export as described earlier.

LarsAsplund:
@TimmyBrolin Here is an example

``` console
d:\Programming\github\vunit\examples\vhdl\uart>python run.py -l
tb_uart_lib.tb_uart_rx
tb_uart_lib.tb_uart_tx.test_send_one_byte
tb_uart_lib.tb_uart_tx.test_send_two_bytes
tb_uart_lib.tb_uart_tx.test_send_many_bytes
Listed 4 tests
```

image

kraigher:
After contemplating a bit I think we should make this change. It aligns us better with other XUnit tools.

LarsAsplund:
@kraigher Agree

kraigher:
I will fix the docs also.

TimmyBrolin:
There might be a compatibility problem with the listing format...
The Parallell Test Executor plugin requires a list of tests in the granularity they are supposed to be parallelized.
VUNit does not support running only one testcase within a testbench, right?

So "python run.py -l" needs to return a list of classnames.

kraigher:
Yes it supports running only one testcase within a test bench.

kraigher:
The changed report format neither removes nor provides additional information on what can be executed in parallel. You can compute the old equivalent name = classname + name

kraigher:
If someone misses the old format of the JUnit.xml we can add a `--xunit-flat` flag

PR #150, opened 2016-03-15
"Adopters guide" documentation

joshrsmith:
Also fixes lint error.

### DIFF
--- a/docs/user_guide.rst
+++ b/docs/user_guide.rst
@@ -133,6 +133,37 @@ A warning will be given if the name *does* match the above pattern but
 lacks a ``runner_cfg`` generic or parameter preventing it to be run
 by VUnit.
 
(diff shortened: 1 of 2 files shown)


LarsAsplund:
I would prefer if this is part of the VHDL Test Benches chapter. When placed close to the original source there is less need to repeat information and you can focus on the major differences, for example that the end-point has to be replaced.

I think a before and after example can be good as well. Something simple like a flip-flop testbench testing that reset and normal operation works.

VUnit-ify in steps, first without test cases and then with test cases. The default behaviour to run test cases in isolation should be emphasized. In case the original code has dependent "test cases" then these dependencies has to be removed or `-- vunit_pragma run_all_in_same_sim` inserted.

joshrsmith:
I'll take another shot at this a later time.

PR #152, opened 2016-03-23
Fixes for Aldec simulators

joshrsmith:
Been working with @LarsAsplund to make these fixes.

Two issues were present:
1. Issue with Aldec simulator interpretation of VHDL when a comparison with the result of a function was an argument to `check`
2. Issue with Aldec simulator script generation where a non-numeric exit status from VUnit (e.g. "Z") would cause the script to fail because of a TCL syntax error.


LarsAsplund:
I added the same TCL fixes to the ModelSim interface and merged it.

PR #159, opened 2016-04-08
Make sure vcover is executed correctly when using coverage

ehliar:
I'm having an issue where --coverage doesn't work. The problem is that execve is trying to execute the command  "vcover merge -strip 1..." with no arguments rather than "vcover" with the arguments "merge" "-strip" "1" etc.

This patch makes it work for me at least, but I'm not a Python wizard so I may have missed something important.


kraigher:
Looks good.

Issue #161, opened 2016-04-15
Running VUnit with ModelSim DE 10.1a results in SIGSEGV


suoto:
I'm running the example from [examples/vhdl/uart] (https://github.com/VUnit/vunit/tree/master/examples/vhdl/uart) folder and ModelSim DE 10.1a is terminating with SIGSEGV when launching the vsim command.

I have both ModelSim DE 10.1a and ModelSim Altera 10.3c and this happens **only** with ModelSim DE 10.1a. (I have cleaned all generated files when switching between the two).

The shell output is as follows:

```
~/dev/vunit/examples/vhdl/uart $ python run.py
Compiling src/test/uart_model_pkg.vhd into tb_uart_lib ...
Compiling src/uart_tx.vhd into uart_lib ...
Compiling src/uart_rx.vhd into uart_lib ...
Compiling src/test/tb_uart_tx.vhd into tb_uart_lib ...
###### /home/asouto/dev/vunit/examples/vhdl/uart/src/test/tb_uart_tx.vhd(111):   end process;
** Warning: [2] /home/asouto/dev/vunit/examples/vhdl/uart/src/test/tb_uart_tx.vhd(111): (vcom-1090) Possible infinite loop: Process contains no WAIT statement.
Compiling src/test/tb_uart_rx.vhd into tb_uart_lib ...
Starting tb_uart_lib.tb_uart_rx.test_tvalid_low_at_start
# vsim -L uart_lib -L vunit_lib -L tb_uart_lib -L osvvm -modelsimini /home/asouto/dev/vunit/examples/vhdl/uart/vunit_out/modelsim/modelsim.ini -onfinish stop -quiet -t ps -wlf /home/asouto/dev/vunit/exam
ples/vhdl/uart/vunit_out/tests/tb_uart_lib.tb_uart_rx.test_tvalid_low_at_start/modelsim/vsim.wlf {-g/tb_uart_rx/runner_cfg="enabled_test_cases : test_tvalid_low_at_start,output path : /home/asouto/dev/vu
nit/examples/vhdl/uart/vunit_out/tests/tb_uart_lib.tb_uart_rx.test_tvalid_low_at_start/,active python runner : true"} tb_uart_lib.tb_uart_rx(tb) 
# //  ModelSim DE 10.1a Feb 21 2012 Linux 3.16.0-0.bpo.4-amd64
# //
# //  Copyright 1991-2012 Mentor Graphics Corporation
# //  All Rights Reserved.
# //
[...]

PR #163, opened 2016-04-19
Possible fix for #161

suoto:
This has fixed issue #161 for ModelSim DE 10.1a on my setup.

In file `vunit/vhdl/string_ops/src/string_ops.vhd` it seems that ModelSim was getting confused with `string_length_after_replace` being created inside the `replace` function.

I'm not sure why this was coded this way since `string_length_after_replace` doesn't need to be dynamic to this point, but I guess it's personal style. Either way, I made the fewest changes as possible.

Let me know what you guys think.

### DIFF
--- a/vunit/vhdl/string_ops/src/string_ops.vhd
+++ b/vunit/vhdl/string_ops/src/string_ops.vhd
@@ -306,15 +306,15 @@ package body string_ops is
     constant cnt : in natural := natural'high)
     return string is
     constant n_occurences : natural := count(s, old_segment);
-    function string_length_after_replace return natural is
-      variable n_replacements : natural := n_occurences;
+    function string_length_after_replace (constant n_occurences_i : natural) return natural is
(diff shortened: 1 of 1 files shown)


LarsAsplund:
Thanks. I'll have a look at it. What did you mean with _doesn't need to be dynamic to this point_?

suoto:
I was thinking of when you can't define a function until you actually run and reach the point you have all info you need. 

Actually I didn't know you could define nested functions in VHDL, so it kind of surprised me. :)

LarsAsplund:
Yes, you can have subprograms locally and it's very useful if you want to encapsulate the details to where it belongs/is used. If I made that function global the code would be less readable since it would be less obvious where it is used. Several places within the `string_ops` package? By `string_ops` package users?

I added a comment on the workaround so it isn't removed by mistake considering we don't have a 10.1a to test with. Thanks for contributing.

PR #166, opened 2016-04-21
Print out remaining number of tests when pressing ctrl-c

ehliar:
If you start running a number of tests in VUnit and press Ctrl-C after a number of tests have passed and no tests have failed you may get the impression that all tests passed.

This is an attempt to test this by printing out a warning that the number of tests that were scheduled to run is not equal to the number of tests that were actually finished.

Caveat emptor: I'm still not a Python wizard, this works for me but I'm not saying that this is the best way to solve this problem. It should at least be able to serve as a starting point for a discussion whether this is a good feature to have in the first place.

### DIFF
--- a/vunit/test_report.py
+++ b/vunit/test_report.py
@@ -27,12 +27,19 @@ class TestReport(object):
         self._test_names_in_order = []
         self._printer = printer
         self._real_total_time = 0.0
+        self._expected_num_tests = 0
 
     def set_real_total_time(self, real_total_time):
         """
         Set the real total execution time
         """
         self._real_total_time = real_total_time
+
+    def set_expected_num_tests(self, expected_num_tests):
+        """
+        Set the number of tests that we expect to run
+        """
+        self._expected_num_tests = expected_num_tests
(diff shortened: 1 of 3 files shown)


kraigher:
It could be an OK solution. We do have a concept for "skipped" tests in the test report. Another way of doing this would be to add all tests not run when pressing Ctrl-C as skipped tests.

vunit/test_report.py
```diff
@@ -151,6 +158,12 @@ def print_str(self):
             self._printer.write("All passed!", fg='gi')
         self._printer.write("\n")
 
+        assert len(all_tests) <= self._expected_num_tests
+        if len(all_tests) < self._expected_num_tests:
+            self._printer.write("WARNING: Expected to run %d tests, but only ran %d tests"
```
kraigher: I think this should use our logging system to make it a "proper" warning.

LarsAsplund: Upgrading our logging system to use color printing would be nice.

Issue #175, opened 2016-06-21
Adding colors to logging output


suoto:
I've been testing colorized VUnit logging so that the output log shows different colors depending on the severity level. The goal is to make easier to spot interesting stuff on large logs.

Below is a capture of how the logs look like:

![vunit_color](https://cloud.githubusercontent.com/assets/6818668/16216626/a546f792-3742-11e6-9285-73b6cf549fbe.png)

Since the code is roughly done, I would like to ask feedback on some issues before diving into fixing the unit tests that are currently failing.
- The coloring scheme looks good enough for me, but you might want to do some fine tuning. I'm not sure how to implement a way for the user to change colors
- User should be able to enable/disable colored logs via run script. Is it OK to user the `--no-color` to control this as well?
- On ModelSim (and probably other simulators), the raw control chars are shown at the transcript window as shown below. To avoid this, using `--gui` must imply `--no-color`. Does this sound OK?
  ![vunit_color_gui](https://cloud.githubusercontent.com/assets/6818668/16216630/b122ca1e-3742-11e6-98bc-35a14402069b.png)

cmarqu:
(User input here.) 
I quite like your screenshot. It looks like it could be made to also work nicely in a terminal with bright background (like I am using).
However, color rendition is rather different between terminals (tried xterm, kterm, gnome-terminal at some point), and depends on advertised capabilities as well (i.e. the same basic colors look different in my gnome-terminal with and without "TERM=xterm-256color" set), so I can imagine the usefulness of full configurability to make it look nice everywhere.
(As for implementing that... Python config options forwarded as string VHDL generics/Verilog parameters maybe?)

suoto:
@cmarqu Thanks for the input. :)
Indeed, different terminal capabilities should be handled and tested. I'm thinking on how to communicate that to VHDL; I guess Python could write/modify something like an 'environment info' package with that. I'll give it a try to check if that could actually work.

LarsAsplund:
Wouldn't it be better to capture the output from the simulator, look for warning/error/failure keywords and highlight those lines? Then we would avoid the lack of such functionality in the simulator transcript windows and we would reuse the color printing we already have for pass/fail status/summary. I'm hoping to get some input from @kraigher on the implementation details  on this since he wrote the relevant functionality.

suoto:
@LarsAsplund, adding colors outside the HDL seems rather interesting. Only that instead of capturing the simulator output directly, I think it's more reliable to use a CSV-like output.

[...]

PR #193, opened 2016-08-16
Adding Python color overlay for coloring VUnit output

suoto:
This PR implements a Python coloring overlay to VUnit as suggested on https://github.com/VUnit/vunit/issues/175#issuecomment-234397294

Below is a list of improvements/limitations:
- Only the first line of multiline reports such as ModelSim's are highlighted. I couldn't come up with a solution to this that didn't involved mixing the color overlay class and the simulator interface class
- The color overlay class does not handle changes on the report format made by the user on `modelsim.ini`, 
- Simulation assertions from Riviera Pro and compilation and simulation messages from Active-HDL are not included as I don't have them for testing


kraigher:
@suoto Thank you for your contribution! It shows that you have really made an effort to build this with quality. As a main author of the Python code in VUnit I see some things in this pull request which I think are mal-placed concerns as well as some architectural issues. It could be hard for someone else to notice these things or it could be just a matter of personal taste. Regardless I will try to find the time in the not so distant future to refactor those parts of your work in such a way that I can accept in into the master branch and give you a return on investment on the time you have spent contributing.

LarsAsplund:
@suoto I had a look at the work you've done and verified the different use cases (compiler/simulator problems, VUnit log messages and VHDL asserts). Still having problems to see all coloring unless I run with `-v`. I see compiler warnings/errors and the colors in the summary report but nothing generated by VUnit logs or VHDL asserts. The pass/fail status printed after each test case also lacks colors regardless if `-v` is active or not.

suoto:
@LarsAsplund Unfortunately I couldn't reproduce these issues you mentioned. Which simulator are you using? Is it possible that you might be using an older commit?

kraigher:
I read Windows 10 will support ANSI color codes. It is probably better to skip all the Python side complexity and just use ANSI color codes instead. It will work on Linux and modern Windows.

kraigher:
I have decided that this is best solved using ANSI escape codes for color in VHDL. I will investigate integrating VUnit with colorama python package which enables support for ANSI escape codes on Windows. This way we avoid any special handling of our VHDL log package messages which I would really like to avoid. @suoto I know you did some code regarding adding ANSI color codes to the VHDL log package. Maybe you could dig that up again? When can then also add a configuration to the runner_cfg to disable ANSI color codes when running in a simulator GUI which will not handle the color codes.

PR #196, opened 2016-08-31
Specify testname in .ucdb file. Resolve problem of conflicting attrib…

mbzeouich:
…utes when merging multiple coverage records.

### DIFF
--- a/vunit/modelsim_interface.py
+++ b/vunit/modelsim_interface.py
@@ -276,7 +276,8 @@ class ModelSimInterface(SimulatorInterface):  # pylint: disable=too-many-instanc
         else:
             coverage_file = join(output_path, "coverage.ucdb")
             self._coverage_files.add(coverage_file)
-            coverage_save_cmd = "coverage save -onexit -assert -directive -cvg -codeAll {%s}" % fix_path(coverage_file)
+            coverage_save_cmd = "coverage save -onexit -assert -directive -cvg -codeAll {%s}" \
+			% (os.path.basename(os.path.dirname(output_path)), fix_path(coverage_file))
             coverage_args = "-coverage=" + to_coverage_args(self._coverage)
 
         vsim_flags = ["-wlf {%s}" % fix_path(join(output_path, "vsim.wlf")),
--- a/vunit/modelsim_interface.py
+++ b/vunit/modelsim_interface.py
@@ -8,7 +8,6 @@
 Interface towards Mentor Graphics ModelSim
 """
 
-
 from __future__ import print_function
 
 import logging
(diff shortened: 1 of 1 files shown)


vunit/modelsim_interface.py
```diff
@@ -276,7 +276,8 @@ def _create_load_function(self,  # pylint: disable=too-many-arguments,too-many-l
         else:
             coverage_file = join(output_path, "coverage.ucdb")
             self._coverage_files.add(coverage_file)
-            coverage_save_cmd = "coverage save -onexit -assert -directive -cvg -codeAll {%s}" % fix_path(coverage_file)
+            coverage_save_cmd = "coverage save -onexit -assert -directive -cvg -codeAll {%s}" \
+			% (os.path.basename(os.path.dirname(output_path)), fix_path(coverage_file))
```
kraigher: I am missing a comment or intermediate variable name to document the intention that dirname of output path is the test name.

mbzeouich: In vunit the outpath always contains the testname. Hereafter the attributes of the generated coverage report with the testname:
`
# coverage read -dataset coverage vunit_out_rtl/tests/tb_lib.tb_mybloc.rtl.testcase_bfm/modelsim/coverage.ucdb
# vunit_out_rtl/tests/tb_lib.tb_mybloc.rtl.testcase_bfm/modelsim/coverage.ucdb opened as coverage dataset "coverage"
coverage attribute
coverage attribute
# TESTNAME = tb_lib.tb_mybloc.rtl.testcase_bfm
#     SIMTIME             = 1182250.000000
#     TIMEUNIT            = ps
`

kraigher: Yes but it is not obvious from the code. As I said I am missing a comment in the code or intermediate variable to make it obvious that it is the test name that is extracted.

In the future the output path migth not even contain the test name since we migth need to compress it due to Windows path length limitations. Then this code would have to be modified, by us, or you will notice it.

mbzeouich: I agree with you that the best way is to use a test name variable, but this parameter is not passed to the SimulationInterface or ModelsimInterface class. And I don't have enough experience with python and Vunit that will allows me to do the necessary updated to add this variable to these classes.
Maybe someone else could help with adding this variable so to correct the merge error.

But I think even if the output directory changes, in any case it will be unique for a test, isn't it?

kraigher:
This pull request is no longer compatible with master since we changed the output_path to not contain the test name but rather a hash of the test name to avoid problems with special characters in test names as well as long path names on Windows. Being a nice guy I implemented the change you wanted on master in a compatible way.

mbzeouich:
Thank you for the update :)
No doubt you are a nice guy

Issue #167, opened 2016-04-27
Problems running run.py in Windows 7 with python 3


RomelTorres:
I am running the user guide vhdl example, and it does not work. I get an error while it tries to write the modelsim_ini: 

self._create_modelsim_ini()
  File "C:\Anaconda3\lib\site-packages\vunit\modelsim_interface.py", line 154, in _create_modelsim_ini
    write_file(self._modelsim_ini, read_file(join(self._prefix, "..", "modelsim.ini")))
  File "C:\Anaconda3\lib\site-packages\vunit\ostools.py", line 270, in write_file
    file_to_write.write(contents)
  File "C:\Anaconda3\lib\encodings\cp1252.py", line 19, in encode return codecs.charmap_encode(input,self.errors,encoding_table)[0] UnicodeEncodeError: 'charmap' codec can't encode characters in position 3273-3274: character maps to <undefined>

The workaround I found was to manually copy the modelsim.ini file into the vunit_out/modelsim directory.

kraigher:
Are you using Python 2.7 or 3.x?

RomelTorres:
Hi I am using Python 3.5

So far what I did was to change this =>  
_cfg.read(self._modelsim_ini)_ for 
_cfg.read_file(self._modelsim_ini, encoding = "utf-8") _
in lines number: 233 and 243 in the vunit/vunit/modelsim_interface.py. It is running so far, basically the problem is that the default encoding (sys.getdefaultencoding()) in windows is cp1252. I think in Python 3 you got to specify the encoding type.

DennisLipschinski:
I have the same problem. Also on the latest release, the potential fix, moves the error from _create_modelsim_ini to configparser.py

my suggestion is:
remove the fix and add a ", errors='replace'" to the open function in write_file

kraigher:
This should be fixed now.

PR #205, opened 2016-11-11
Fix for #167

DennisLipschinski:


DennisLipschinski:
Seems to work on windows only :(

kraigher:
I am a bit reluctant to add a complete ignore of encoding errors to such a general function such as reading and writing files. Will this really work? I would guess that some information that something needs must be lost if there really was an encoding error. It could lead to hard to debug errors if some character was missing from the file.

I will think about if I can do this differently without ignoring the encoding errors without warning. If cant think of anything better than this I will approve the pull request.

DennisLipschinski:
I understand your concerns. 
But what do you thing should happen if the modelsim.ini contains decodable characters (like it does in newer version)? -> chrashing with a exception, like at the moment, is not a good user feedback
 
One idea: add a new parameter "ignore_decoding_errors" to the "read_file" function so the return value is always writeable ...

kraigher:
Well the most correct thing to do would be to open the modelsim.ini file with the correct encoding, maybe it is utf-8, such that there are no encoding errors.

Just ignoring encoding/decoding errors for modelsim.ini only like you suggest at least not as bad ignoring it for all files read and written. 

Could you send me your modelsim.ini file so that I can see what encoding it is really supposed to have?
Maybe you can try to change the encoding to utf-8 in Python and see if it solves the problem for you?

DennisLipschinski:
To be honest, i didn't look deeper to the modelsim.ini file. I just tried some encodings (also utf-8) without success.
[modelsim.ini.gz](https://github.com/VUnit/vunit/files/591260/modelsim.ini.gz)

kraigher:
I tried decoding the file you sent using utf-8 and it worked.

kraigher:
I will make a fix where the modelsim.ini file is always read using utf-8. If it fails to read it with utf-8 it will issue a warning and re-read ignoring the decoding/encoding errors.

DennisLipschinski:
Ok but, for me utf-8 doesn't work (with python 3.5.1):
reader with utf-8:
`UnicodeDecodeError: 'utf-8' codec can't decode byte 0xfc in position 496: invalid start byte`
writer with utf-8:
`UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d in position 3324: character maps to <undefined>`

kraigher:
I tried this on the file you sent, it worked with utf-8. Strange.

``` python
Python 3.5.1 (v3.5.1:37a07cee5969, Dec  6 2015, 01:38:48) [MSC v.1900 32 bit (Intel)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> len(open("modelsim.ini").read())
  File "<stdin>", line 1, in <module>
  File "C:\Python35\lib\encodings\cp1252.py", line 23, in decode
    return codecs.charmap_decode(input,self.errors,decoding_table)[0]
UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d in position 3239: character maps to <undefined>
>>> len(open("modelsim.ini", encoding="utf-8").read())
92427
>>>
```

DennisLipschinski:
OK i did a deeper look. The problem was that I changed the read_file function encode generally and this caused a decode error on a VHDL file later in the flow... And I was just looking to the last line of output ... sorry for the confusion.
So your proposed solution should work fine for the problem.

kraigher:
This should be fixed by the latest commit.

kraigher:
@DennisLipschinski Are you satisfied with the solution?

DennisLipschinski:
@kraigher -> cloned the current repo and now Modelsim 10.5b runs without a problem! Thanks!

PR #206, opened 2016-11-12
Use tox as backend for Travis testing

joshrsmith:
I noticed that the tox.ini has been broken for awhile. Specifically the "warning as error" flag to Sphinx was causing the documentation test to fail out (https://github.com/VUnit/vunit/commit/2da6ca7121f2f3a3465422a83fb556207652c19b). While looking at this, I realized we really have two ways of testing, each with some of their own configuration. I decided to try and reconcile this into one method.

1. I reworked the tox.ini so that it could be compatible with Travis (working around the limitation of Travis not being able to check the years in the license info).
2. I eliminated travis.py, since this has been replaced by using tox and tox.ini
3. Travis now directly calls tox, and the configurations for testing are mostly handled by tox. The travis.yml file is still needed for some bootstrap operations with Travis.

This should keep tox and travis testing configurations in sync.

There are also a couple ninja-changes related to lint checking. I ran into a issue where git checked out and changed line-endings which caused a mismatch with what was expected. Using universal newline conversion fixed this.


kraigher:
Thanks, nice to clean up redudancy

kraigher:
Oops wrong button

PR #213, opened 2016-11-25
Add verbose list option

go2sh:
This pull-request adds support for an advanced version of the list generation command. This list contains also the file and the line, where the test could be found.

Therefor I added some information to the testcase classes and changed some properties. Additionally the run_string parser is line based now und returns the run_string with the line number.

I developed this to support editor integration and I might think about some more commits to make the output more easy to parse.

### DIFF
--- a/vunit/test/unit/test_test_scanner.py
+++ b/vunit/test/unit/test_test_scanner.py
@@ -338,7 +338,8 @@ if run("Test_2")
             if isinstance(test2, tuple):
                 name, test_cases = test2
                 self.assertEqual(test1.name, name)
-                self.assertEqual(test1.test_cases, list(test_cases))
+                self.assertEqual(
+                    [test_case.name for test_case in test1.test_cases], list(test_cases))
             else:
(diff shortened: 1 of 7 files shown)


go2sh:
Argh, some I missed the integration tests...

kraigher:
It has potential though it is missing unit tests for the new functionality. Probably some tests in test_test_scanner.py could be extended to also check the new run_string line number is parsed. Be sure to do it for both VHDL and Verilog. Also the discriptor (descriptor?) method needs unit tests. Currently there is a failure for SameSimTestCase:s where the method is named descriptor but ui.py calls discriptor. Shouldnt be that much extra work and we care about quality.

go2sh:
Yeah, for sure. I fix the failing tests and add test cases for added code. I''ll find some time tomorrow.

kraigher:
Has this been abandoned?

kraigher:
I have made a big refactor which makes this code hard to merge. The refactor makes it easier to add this feature though

Issue #218, opened 2016-12-06
VUnit is not aware of the libraries added via the simulator config


KevinKes:
The modelsim installation comes with pre compiled libraries (e.g. ieee, std, vital2000, etc) and it can also contain other pre compiled libraries (in my case all xilinx libs).

I don't like to use `add_external_library` from the VUnit api to make the simulator libraries known to VUnit, because then I don't like to duplicate the modelsim library administration by hand and synchronize all the paths on all build PCs.

When not using the `add_external_library`, VUnit prints a warning in the console because it cannot find the depending library. This is not a big deal but it adds a lot of confusing warning messages to the console.

I suggest that VUnit adds the known simulator libraries to the internal VUnit project. The `ModelSimInterface` already knows which simulator libraries are configured via the ini file. It seems that [ModelSimInterface.setup_library_mapping](https://github.com/VUnit/vunit/blob/master/vunit/modelsim_interface.py#L174) can be updated to add the simulator libraries to the internal VUnit project.

What do you think?

joshrsmith:
This doesn't answer your problem, and in fact I think it suggests what you are trying to avoid.

I have taken the practice of assuming only libraries that are available with a simulator by default installation (ieee, std, etc) are available when setting up a project. Every project is then responsible for mapping their required libraries into their own simulation environment with no assumption of a pre-configured environment (other than the simulator tool is available).

This allows different projects with different library version dependencies to use the same base simulation environment, by pushing the work of "Bring your own dependencies" to all the projects that use the environment.  Since I also version control my simulation library dependencies, getting the simulation environment for a project setup at build/simulation time is a straightforward process of mapping libraries in version control.

That said, even with this approach, I think your suggestion may still be important to consider. If there are libraries that have been mapped into the global library list of the simulator environment (whatever the reason), then VUnit should be able to find out about it, and not emit a warning if those global libraries are not explicitly added later (regardless if they are standard libraries such as `ieee` or not).

kraigher:
@KevinKes So your problem is only that VUnit warns about missing library references? The library is still found when compiling?

There are multiple possible actions here where more than one can be performed:
1. We can add a method to the public API to ask for the simulator installation prefix which VUnit already knows about. This makes it easier for the run.py script to add installed libraries relrative to the installation prefix.

2. We can add a public method of library names to ignore when issuing a missing reference warning like with external libraries.

3. We can let the simulator interfaces automatically detect what libraries are already installed and automatically disable warnings for those.

[...]

PR #219, opened 2016-12-23
Make VUnit aware of ModelSim libraries

KevinKes:
Add external libraries that are available in ModelSim also to the
internal VUnit project to prevent warnings being issued to the console
about missing libraries. Discussed in issue #218


Piressss:
I would like to know if the Vunit is aware of ModelSim Libraries and others externals libraries that are compiled and referenced in the default modelsim.ini.

kraigher:
@Piressss yes it is aware.

PR #221, opened 2017-01-23
Update activehdl_interface.py

sethmkaras:
removed the -sv2k12 from for compiling verilog.  It is not an available flag/option in Active-HDL 9.3-10.3


kraigher:
I fixed this on master and updated the failing unit test

PR #229, opened 2017-02-11
TCL's command pipe will always return failing error codes on anything that prints to stderr

### LINKED ISSUE #228: modelsim vunit_restart doesn't work
Output with no source file changes:
``` 
vunit_restart
# Re-compiling using command <my_path2_python>\python.exe -u run.py --compile lib.tb_xxx.yyy --gui
# Re-compile failed
```
Output with a source file change:
```
vunit_restart
vunit_restart
# Re-compiling using command <my_path2_python>\python.exe -u run.py --compile lib.tb_xxx.yyy --gui
# Compiling tb\tb_xxx.vhd into lib ...
# ** Warning: (vcom-14) Failed to open "modelsim.ini" specified by the MODELSIM environment variable.
# 
# The system cannot find the file specified. (GetLastError() = 2)
# Re-compile failed
```

modelsim 10.4a if that matters.

nathanaelhuffman:
Merge stderr to stdout from python so that TCL's catch will rely on the error code returned rather than being tricked by warnings printed to stderr.  Fixes #228

### DIFF
--- a/vunit/vsim_simulator_mixin.py
+++ b/vunit/vsim_simulator_mixin.py
@@ -56,7 +56,7 @@ class VsimSimulatorMixin(object):
         turn call the python command we actually wanted but in the
         correct working directory using subprocess.call
 
-        -u flag is needed for continous output
+        -u flag is needed for continuous output
         """
         recompile_command = [
             sys.executable,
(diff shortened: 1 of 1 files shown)


kraigher:
Great catch! Could you also add a code comment above that line explaining the problem like you do in a comment here. Better also refer to the issue in the code comment. That way the reasoning is not lost when someone in the future feels an itch to change the line. Regression of your error will not be caught by automated test.

nathanaelhuffman:
Done.  Thanks!

PR #242, opened 2017-05-19
Fix issue 225

noasic:
Coverage support for riviera-pro. Tested on my machine and the customer's one.


kraigher:
I have taken this pull request and fixed it up a bit removing the lint warnings and factoring out some common code. I also made the `--coverage` flag trigger re-compile of files so that you do not have to remember to do `--clean`. I pushed this commit to master and thus I close this pull request.

PR #243, opened 2017-05-25
don't use -dbg when collecting code coverage

noasic:
For faster simulation, Aldec recommends to not use the -dbg switch when collecting code coverage data.


kraigher:
This fails the unit test checking that --dbg was set.   You need to fix the unit test as well.

kraigher:
@noasic Do you have the intention to resolve the failing unit test of this pull request or will it be abandoned?

noasic:
Thanks for the reminder. I've comitted the corresponding fix.

PR #255, opened 2017-06-13
Fixed bug in array type parsing.

benreynwar:
The fixes a typo in the pattern matching of array subtype definitions.

The space between "array" and the range should be optional.

t_whatever is array (integer range <>) of t_thing;
vs
t_whatever is array(integer range <>) of t_thing;


kraigher:
This pull request is missing unit tests that ensures this works now and in the future

PR #256, opened 2017-06-13
Add a command-line argument that can turn off the catching of excepti…

benreynwar:
The commit adds a command line options --dont-catch-exceptions which prevents exceptions that are raised in the test from being caught.
I use this so that when I run "python -m pdb -c continue my_test.py" I drop into the REPL when I hit an exception and can easily debug.


benreynwar:
I haven't added a test, because I thought I'd check in to see if this addition was worth cluttering the code base with before going to the effort of making a test.

PR #259, opened 2017-07-03
Declare global variable 'aldec' in vunit_load, in order to fix the Matlab interface.

senst:
Attempting to use the interface between Riviera-PRO and Matlab from a VUnit-testbench currently leads to the following error:
VHPI: Loading library '$ALDEC/bin/aldec_matlab_cosim.dll'
VHPI: The target platform of the "$ALDEC/bin/aldec_matlab_cosim.dll" library cannot be identified. The library may be missing, corrupted, or its format is incorrect.

The problem is that the variable 'aldec' is not visible from within the function vunit_load. This is fixed by declaring the variable as global.

### DIFF
--- a/vunit/rivierapro_interface.py
+++ b/vunit/rivierapro_interface.py
@@ -247,6 +247,7 @@ class RivieraProInterface(VsimSimulatorMixin, SimulatorInterface):
 
         tcl = """
 proc vunit_load {{}} {{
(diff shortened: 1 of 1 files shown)


kraigher:
A comment in the code on why this is needed above the lie would be appreciated. Then we can remember why it is there in a couple of years.

PR #266, opened 2017-07-21
Add parsing of VHDL subtypes and constants.

benreynwar:


benreynwar:
Looks like my changes clash with some commits in the last couple of days.  I'll sort that out now.

benreynwar:
Looks like the acceptance-ghdl tests are failing, but it not obvious from looking at the logs why.

kraigher:
Look here for an example; https://travis-ci.org/VUnit/vunit/jobs/255923277#L5509
There are more failures if you search the log.

In general if the VHDL parser fails the `run.py` file will continue anyway and disregard dependency scanning of that file but print a stack trace. This is because a syntax error could trigger a parser bug and it is better to let the simulator compiler tell the user what is wrong with the file.

Also I should mention that I am not really that enthusiastic on adding more ad-hoc parsing options to the VUnit VHDL parser at it is in conflict with the goals that it should be robust and fast. The VUnit VHDL parser does not need to be a complete standard compliant parser but a good heuristic way to get the dependencies of the source files. 

For general purpose analysis of VHDL code the VUnit parser architecture we have is not the right tool for the job and adding more regex to it in the same manner as now without making it a real parser with tokenization and recursive grammar just makes it more likely to fail on corner cases found out in the wild where there are millions of lines of VHDL code going through the it. 

Also the VUnit VHDL parser is not part of the public API of VUnit and is not really intended to be used by others I would not recommend it but people have been known to do it. If you want real parsing of VHDL it is better to use a project which has those goals and a the right software architecture for it such as; https://github.com/Paebbels/pyVHDLParser made by @Paebbels

It is possible that we choose to integrate a full VHDL parser into VUnit but it will not be based in any way on the architecture of `vhdl_parser.py` and most likely be a separate project from VUnit.

kraigher:
By the way have you though of subtypes from unconstrained records which is supported in VHDL 2008?

kraigher:
I think if we merge this it must be turned off by default and only enabled explicitly for those user who use our VHDLParser as a free-standing class. (Which is not considered public API)

benreynwar:
I"m using the VHDL parser to be able to capture the interface of a module.
I need subtypes and constants since these are often used in the port
defintiions.  It sounds like for now maybe I should just keep this as my
own branch?

Another possibility would be to make vhdl_parser be configurable with
additional types of parsing.  So for example I could define the VHDLSubtype
and VHDLConstant in my package, and then register them into the vhdl_parser
module somehow.  That would prevent vhdl_parser from filling up with
user-specific cruft, but still allow users to take advantage of and
customize the parser.

benreynwar:
Thought about it some more, and I think making the parser configurable
would be messy and not worth it.  The real solution is a proper fast parser
as you suggested.  For now I'll just keep my changes on my own branch, and
put looking into better parsers on my todo list :).

benreynwar:
Out of curiosity, do you know if anyone has experimented with the
performance of a VHDL parser generated using antlr4?

kraigher:
No I have not seen any performance data on that.

Paebbels:
@benreynwar As far as I know the antlr grammar is not complete for VHDL. Do you have a link for a current grammar? The problem with VHDL is that I know no parser and grammar that is able to parse VHDL because of some special rules in the language. Many tools use hand written parsers.

Are you interested in playing with pyVHDLParser?

benreynwar:
@paebbels Thanks Patrick.  I'll check out pyVHDLParser this weekend.

PR #272, opened 2017-08-09
Set initial values on control signals in axi_lite_master.vhd

LukasVik:
Some Xilinx AXI IP can't handle e.g. awvalid being 'X'.


kraigher:
I manually integrated this commit.

Issue #32, opened 2015-03-09
Check almost equal


barri:
A neat feature when comparing different implementations of the same algorithm would be fault tolerant checkers. The checker would compare two values and allow for a user specified error tolerance. This would allow for easy comparison between different implementations vs some reference data or between each other. A bonus would be getting a report of minimum, maximum and average differences between the two value sets.

LarsAsplund:
The pure checker should be doable. It is a more generalised form of check_equal. However, I feel that collecting stats is something that should be done elsewhere.  Checks should be concerned with pass/fail. Stats would make it concerned with degrees of pass and fail. The question is how important this is to the user.  If that is important he/she would do the abs (measurement - value) calculations anyway and lot of the added convenience with a new checker would be lost. If you already have the diff you can do check_relation (diff < tolerance ). What do you think?

barri:
I think the pure checker would be useful to have for the cases where you don't need the stats. For the use cases where the stats are needed you can use the check_relation in the manner you suggested.

LarsAsplund:
Another comment. Even though a statistics module doesn't belong to the checker functionality it may still have a place in VUnit. Unit testing support is at focus of VUnit but we have not excluded other reusable components for testbench development. The communication mechanism is one such example. Given that testbenches based on a reference model is a common design pattern I would expect that there are reusable components in that area. Suggestions are appreciated.

LarsAsplund:
@barri  Right now I will focus on getting the communication mechanism ready. Please comment the latest additions!

However, let's start with some inputs for the "within tolerances checker"
- Would tolerances be symmetrical, i.e is a == b in ref-a <= measurement <= ref+b?
- What datatypes are you considering?
- I haven't given it any thought but would the addition of a tolerance parameter to check_equal be an acceptable solution? Default of that parameter would be zero. @kraigher Yor thoughts on this?

barri:
Supporting the signed and unsigned datatypes with symmetrical tolerances would probably cover the basic need. Adding a tolerance parameter to check_equal would be a a simple solution but it might be clearer to have a new function.

(Will try to catch up on the communication mechanism tonight.)

kraigher:
[...]

PR #279, opened 2017-10-19
Add check_equal for real data type

ludli505:
Test cases and overloaded functions are pretty much copy-pasted from those used for time checking.

Cheers!

### DIFF
--- a/vunit/vhdl/check/src/check.vhd
+++ b/vunit/vhdl/check/src/check.vhd
@@ -4021,6 +4021,103 @@ package body check_pkg is
     return pass;
   end;
 
+  procedure check_equal(
+    constant got             : in real;
+    constant expected        : in real;
+    constant msg             : in string := check_result_tag_c;
+    constant level           : in log_level_t := dflt;
+    constant line_num        : in natural     := 0;
+    constant file_name       : in string      := "") is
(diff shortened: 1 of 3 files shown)


LarsAsplund:
Are you sure that this is what you want/need considering that

``` vhdl
check_equal(0.15 + 0.15, 0.2 + 0.1);
```

results in

```
ERROR: Equality check failed - Got 3.000000e-01. Expected 3.000000e-01.
```

Wouldn't it be better to implement #32? If you really want an exact equality you can set `eps` to zero. Such a statement would better indicate "This is not a bug, I know what I'm doing"

kraigher:
This will cause merge conflicts with the bfm branch where the modified files have been removed in favor of code generation.

ludli505:
@larsasplund: Yep, your suggestion seems better.

@kraigher: I guess such adding #32 would cause conflicts as well,
considering I'd have to edit at least check_api.vhd?

kraigher:
@ludli505 There is now a check_equal on master for reals that compares the absolute difference of got and expected to a max_diff argument that is 0 by default.

ludli505:
Super, thanks!

Den 23 feb. 2018 21:05 skrev "kraigher" <notifications@github.com>:

PR #288, opened 2017-12-04
Fixup for modelsim coverage argument

ludli505:
Using Questasim 10.6, the "-coverage=<coveragetypes>" argument given by VUnit to vsim is not valid. The argument should only be given as "-coverage". With Questasim 10.5, the faulty argument is simply ignored.

A Mentor FAE has confirmed that the argument to vsim should be "-coverage" for all versions of Modelsim and Questasim.

It would of course be great if someone using the coverage functionality with Modelsim could this as well (I guess VUnit haven't been granted Modelsim licenses?).

Cheers!
/Ludde


LarsAsplund:
No, we have yet to get some developer licenses for ModelSim. Haven't tried in a while though.

The acceptance tests for your pull request fails due to pytest dropping support for Python 3.3 which has reach end of life. We will drop that too and I just pushed that change to master.

ludli505:
Ok, thanks. I've pulled your changes to the fork - hopefully that was the correct way to do it..

Issue #188, opened 2016-07-25
Include free verilog compiler


leoheck:
Like ``icarus verilog``

LarsAsplund:
Some work has been done by @ehliar. Not sure about current status. Join https://gitter.im/VUnit/vunit to find out

AlexDaniel:
Any news?

LarsAsplund:
No, not really. This is an area where we would appreciate contributions. Have you been able to work on this @ehliar?

feddischson:
Did someone get a response from @ehliar about his initial steps? (https://gitter.im/VUnit/vunit/archives/2016/04/27).

I am interested in having the `icarus verilog` support and I am willing to help.

svenka3:
I made some attempt. I've a crude version running. However it is only skeleton as Icarus has issues around SV string methods, copy etc. I would like to support especially if there is some spec on what it takes to run a unit test under VUnit - like there is setup, loop etc. I believe it is obvious for SW folks, not so popular for HW engineers I guess. I've been hacking the unit_pkg files with those APIs - making them skeleton and uncommenting line-by-line. I feel that spoils the core functionality at some point as there are work-arounds for Icarus. Is there a doc for the VUNIT API from a developer perspective?

Thanks

LarsAsplund:
@svenka3 Would it be possible to make your work public so that everyone interested can help out?

The starting point for our documentation is VUnit.github.io. There you'll find how to use VUnit but there is no documentation for the details needed to develop the tool. The only information we have is how to make sure that the things you've made doesn't break anything. http://vunit.github.io/contributing.html

svenka3:
Sure, will upload. Please give me few days to do so.

svenka3:
Quick update - I believe I now have the "loop" function ported to Icarus. Looks like Icarus support for string is minimal and hence the remaining 2 functions: setup() and search_replace() seem unlikely to run on it.

Can you please clarify what you would like to achieve via the setup()? I am sure we can find a pure Verilog way of doing it. Search-Replace - good for future maybe!

Thanks
Srini

svenka3:
Please see: https://github.com/svenka3/vunit_icarus 

Please do let me know if you see any issue in running it. As I cautioned earlier - this is heavily work in progress and not ready for release. Need some more work.

feddischson:
@svenka3 Nice, this is a starting point.
[...]

PR #306, opened 2018-01-21
Adds unit-test to demonstrate failing instance detection via Verilog's generate

feddischson:


kraigher:
I updated your pull request with another commit. Please try it before I merge on your code to see that is does not cause any other side-effect.

kraigher:
Your last commit will fail pep8 style check due to too many lines between two class methods. See http://vunit.github.io/contributing.html for how to run them locally.

kraigher:
I do not think the additional test with nesting offers any extra test coverage. What tricked it up was the block label. Maybe you can add a failing test for block label after end. That would fail now.

feddischson:
Block-labels after end seems to work.

kraigher:
The bug occurs when the instance comes immediately after a block label.

kraigher:
Does it work in your code-base now?

feddischson:
Yes, it works fine. Thanks for your support.
It helped me to realize the dependency-graph based compilation for the icarus support (mentioned [here](https://github.com/VUnit/vunit/issues/188#issuecomment-359199288))

PR #312, opened 2018-03-02
Wishbone BFM

slaweksiluk:
I think my implementations of Wishbone Slave and Master Bus Functional Models are ready to be reviewed. Currently only pipelined mode is supported. Read/Write cycles lengths can be 1 (Single Read/Write) or more (Block Read/Write).

### Wishbone Slave BFM
It is wrapper for Vunit memory VC. Currently memory is internal variable, but it should rather be passed as generic parameter. Manual tb with generic test cases is provided.

### Wishbone Master BFM
Receives bus write and bus read messages from main sequencer. Consecutive groups of bus write/reads messages are merged into block cycles. Cycle is finished when next message is the opposite action (e.g write command after read) or there is no more messages in actor inbox. Test bench with Master and Slave connected is provided. 

I didn't want to add new calls to bus_maser_pkg. However adding calls like: bus_write_block and bus_read_block might be good idea.

### DIFF
--- a/vunit/vhdl/verification_components/src/wishbone_master.vhd
+++ b/vunit/vhdl/verification_components/src/wishbone_master.vhd
@@ -14,6 +14,7 @@ use ieee.std_logic_1164.all;
 use work.queue_pkg.all;
 use work.bus_master_pkg.all;
 context work.com_context;
+use work.com_types_pkg.all;
 use work.logger_pkg.all;
 use work.check_pkg.all;
 
@@ -49,15 +50,15 @@ begin
     variable msg_type : msg_type_t;
     variable status : com_status_t;
   begin
-      -- Cannot use receive, as it deletes the message
-      --receive(net, bus_handle.p_actor, request_msg);
-      wait_for_message(net, bus_handle.p_actor, status);
-      get_message(net, bus_handle.p_actor, request_msg);
+      request_msg := null_msg;
+      receive(net, bus_handle.p_actor, request_msg);
       msg_type := message_type(request_msg);
       if msg_type = bus_read_msg then
         push(rd_request_queue, request_msg);
+        request_msg := null_msg;
       elsif msg_type = bus_write_msg then
         push(wr_request_queue, request_msg);
+        request_msg := null_msg;
       else
         unexpected_msg_type(msg_type);
       end if;
--- a/vunit/vhdl/verification_components/src/wishbone_master.vhd
+++ b/vunit/vhdl/verification_components/src/wishbone_master.vhd
@@ -48,11 +48,9 @@ begin
     stb <= '0';
     wait until rising_edge(clk);
     loop
-      trace(bus_handle.p_logger, "blocking on receive");
       receive(net, bus_handle.p_actor, request_msg);
       msg_type := message_type(request_msg);
       if msg_type = bus_read_msg then
-		trace(bus_handle.p_logger, "got bus read msg");      
         adr <= pop_std_ulogic_vector(request_msg);
         cyc <= '1';
         stb <= '1';
@@ -64,7 +62,6 @@ begin
         pending_acks := pending_acks +1;
 
       elsif msg_type = bus_write_msg then
-		trace(bus_handle.p_logger, "got bus write msg");            
         adr <= pop_std_ulogic_vector(request_msg);
         dat_o <= pop_std_ulogic_vector(request_msg);
         sel <= pop_std_ulogic_vector(request_msg);
@@ -80,7 +77,6 @@ begin
 
       elsif msg_type = bus_ack_msg then
         -- TODO bus errors detection
-        trace(bus_handle.p_logger, "Got bus ack msg");
         received_acks := received_acks +1;
         if pending_acks = received_acks then
           -- End of wb cycle
@@ -108,12 +104,10 @@ begin
       push_std_ulogic_vector(reply_msg, dat_i);
       reply(net, request_msg, reply_msg);
       delete(request_msg);
-      trace("Sent msg to bus_master_pkg");
     end if;
     -- Response main that ack is received
     ack_msg := new_msg(bus_ack_msg);
     send(net, bus_handle.p_actor, ack_msg);
-    trace("Sent msg to main");
   end loop;
   end process;
 end architecture;
--- a/vunit/vhdl/verification_components/src/wishbone_master.vhd
+++ b/vunit/vhdl/verification_components/src/wishbone_master.vhd
@@ -5,7 +5,7 @@
 -- Slawomir Siluk slaweksiluk@gazeta.pl 2018
 -- Wishbome Master BFM for pipelined block transfers
 -- TODO:
---  * Random strobe option
+--  * Random strobe?
 library ieee;
 use ieee.std_logic_1164.all;
 
@@ -71,7 +71,6 @@ begin
     variable received_acks : natural := 0;
     variable rd_cycle : boolean;
     variable wr_cycle : boolean;
-    variable ack_mock : boolean;
   begin
     cyc <= '0';
     stb <= '0';
@@ -118,7 +117,7 @@ begin
           received_acks := 0;
           rd_cycle := false;
           wr_cycle := false;
-	      wait until rising_edge(clk);
+          wait until rising_edge(clk);
         end if;
 
       -- No cycles, juest sleep
@@ -130,7 +129,6 @@ begin
 
   acknowladge : process
     variable request_msg, reply_msg, ack_msg : msg_t;
-    variable ack_mock : boolean;
   begin
     wait until ack = '1' and rising_edge(clk);
     info(bus_handle.p_logger, "ack");
@@ -142,7 +140,7 @@ begin
       reply(net, request_msg, reply_msg);
     end if;
     delete(request_msg);
-    -- Response main that ack is received
+    -- Response main sequencer that ack is received
     ack_msg := new_msg(bus_ack_msg);
     send(net, wb_master_ack_actor, ack_msg);
   end process;
--- a/vunit/vhdl/verification_components/src/wishbone_master.vhd
+++ b/vunit/vhdl/verification_components/src/wishbone_master.vhd
@@ -37,7 +37,7 @@ end entity;
 architecture a of wishbone_master is
   constant request_queue : queue_t := new_queue;
   constant bus_ack_msg   : msg_type_t := new_msg_type("wb master ack msg");
-  constant wb_master_ack_actor : actor_t := new_actor("wb master ack actor");  
+  constant wb_master_ack_actor : actor_t := new_actor("wb master ack actor");
 begin
   main : process
     variable request_msg, reply_msg : msg_t;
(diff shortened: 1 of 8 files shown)


vunit/vhdl/verification_components/test/tb_wishbone_master.vhd
```diff
@@ -0,0 +1,167 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this file,
+-- You can obtain one at http://mozilla.org/MPL/2.0/.
+--
```
kraigher: Copyright needs to be given to Lars. You can keep your name for crediting the contribution.

vunit/vhdl/verification_components/test/tb_wishbone_master.vhd
```diff
@@ -0,0 +1,167 @@
...
+    wait for 50 ns;
+    test_runner_cleanup(runner);
+    wait;
+  end process;
+  test_runner_watchdog(runner, 100 us);
+  set_format(display_handler, verbose, true);
+  show(tb_logger, display_handler, verbose);
+  show(default_logger, display_handler, verbose);
+  show(master_logger, display_handler, verbose);
+  show(com_logger, display_handler, verbose);
```
kraigher: All these calls (except test_runnner_watchdog) should be after test_runner_setup in a process. I do not recommend making these calls as concurrent statements.

vunit/vhdl/verification_components/test/tb_wishbone_slave.vhd
```diff
@@ -0,0 +1,126 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this file,
+-- You can obtain one at http://mozilla.org/MPL/2.0/.
+--
```
kraigher: Copyright needs to be given to Lars, you can keep your names for crediting the work but have to give up copyright.

vunit/vhdl/verification_components/test/tb_wishbone_slave.vhd
```diff
@@ -0,0 +1,126 @@
...
+
+    wait for 50 ns;
+    test_runner_cleanup(runner);
+    wait;
+  end process;
+  test_runner_watchdog(runner, 100 us);
+  set_format(display_handler, verbose, true);
+  show(tb_logger, display_handler, verbose);
+  show(default_logger, display_handler, verbose);
+  show(com_logger, display_handler, verbose);
```
kraigher: These calls (except test_runner_watchdog) should not be concurrent and should be located after test_runner_setup

vunit/vhdl/verification_components/src/wishbone_slave.vhd
```diff
@@ -0,0 +1,131 @@
...
+      send(net, ack_actor, rd_request_msg);
+    end if;
+  end process;
+
+  acknowledge : process
+    variable request_msg : msg_t;
+    variable msg_type : msg_type_t;
+    variable data : std_logic_vector(dat_i'range);
+    variable addr : natural;
+    variable memory : memory_t := new_memory;
```
kraigher: When using a memory_t from a VC you should convert it by using to_vc_interface. This enables permission checks on all read/write accesses which is not enabled by default as the test_bench should read and write freely but the VC should respect the permission settings.

vunit/vhdl/verification_components/src/wishbone_master.vhd
```diff
@@ -0,0 +1,147 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this file,
+-- You can obtain one at http://mozilla.org/MPL/2.0/.
+--
```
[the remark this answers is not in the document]
slaweksiluk: OK, no problem I will leave Copyright as in other vunit sources. I will also add line:
```
-- Author:  Slawomir Siluk slaweksiluk@gazeta.pl 
```

vunit/vhdl/verification_components/src/wishbone_master.vhd
```diff
@@ -0,0 +1,147 @@
...
+  constant wb_master_ack_actor : actor_t := new_actor("wb master ack actor");
+begin
+
+  main : process
+    variable request_msg : msg_t;
+    variable msg_type : msg_type_t;
+    variable status : com_status_t;
+  begin
+      -- Cannot use receive, as it deletes the message
+      --receive(net, bus_handle.p_actor, request_msg);
```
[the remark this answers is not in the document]
slaweksiluk: Do you mean:
```
request_msg := null_message
receive(net, bus_handle.p_actor, request_msg)
```
I sholud still manually delete the message, right?

vunit/vhdl/verification_components/src/wishbone_slave.vhd
slaweksiluk: I'm going to pass memory as entity generic. Should I still use to_vc_interface in that case? I will look for some examples in existing vunit VC's.

vunit/vhdl/verification_components/src/wishbone_master.vhd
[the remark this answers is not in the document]
kraigher: The receive procedure will delete the request_msg if it is not already null_message.
By setting it to null_message before you inhibit receive from deleting the message.
You take the responsibility to delete the message yourself when you are done using it.
Ideally you should set the request_msg to null_message right after you push it to a queue since that is the point where ownership is semantically handed over to the queue. The process that pops the queue takes ownership and should delete it once it is done with it.

vunit/vhdl/verification_components/src/wishbone_master.vhd
```diff
@@ -0,0 +1,147 @@
...
+      wait_for_message(net, bus_handle.p_actor, status);
+      get_message(net, bus_handle.p_actor, request_msg);
+      msg_type := message_type(request_msg);
+      if msg_type = bus_read_msg then
+        push(rd_request_queue, request_msg);
```
kraigher: set request_msg to null_message after push since ownership was transferred into the queue.

vunit/vhdl/verification_components/src/wishbone_master.vhd
```diff
@@ -0,0 +1,147 @@
...
+      elsif msg_type = bus_write_msg then
+        push(wr_request_queue, request_msg);
```
kraigher: set request_msg to null_message after push since ownership was transferred into the queue.

vunit/vhdl/verification_components/src/wishbone_slave.vhd
kraigher: to_vc_interface should be used privately within the VC to create a new memory object which is identical to the one that was passed into the VC by the user. This has the effect of enabling read/write permission for all read/write operations to the memory when using the new value returned by to_vc_interface. This is to make it easier to support the two use cases of reading/writing memory from the test bench without permission checks and having axi_slave etc. perform memory operations with permission checks.

kraigher: Also ideally you should create a wishbone_slave_t record type that is created with a new_wishbone_slave function like axi_slave. The new_wishbone_slave can take a memory_t as an argument and maybe also a logger_t. Internally in new_wishbone_slave and actor_t can be created that allows you to define command functions to the wishbone slave such as `enable_random_stall(net : net_t, slave : wishbone_slave_t)` or `set_random_stall_probability(net : net_t, slave : wishbone_slave_t, probability : real);`

vunit/vhdl/verification_components/src/wishbone_slave.vhd
```diff
@@ -0,0 +1,131 @@
...
+context work.com_context;
+
+use work.memory_pkg.all;
+
+library osvvm;
+use osvvm.RandomPkg.all;
+
+entity wishbone_slave is
+  generic (
+    max_ack_dly : natural := 0;
```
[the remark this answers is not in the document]
slaweksiluk: Ok, I will define wishbone_slave_t record in wishbone_pkg.vhd (please propose the other file name if necessary)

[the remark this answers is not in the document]
kraigher: Sounds good

slaweksiluk:
Ok, thank you for comments. I applied most of them, excluding random parameters set via msg. It is definitely better solution, however I need to have some deeper look into axi slaves to imperilment it.

kraigher:
Regarding setting random parameters via message. The first you could to is just add them as fields in the wishbone_slave_t record and then rename them to initial_rand_stall_probability. That way it is forward compatible to a future where the rand_stall_probability can be set at run-time as well. In the axi_slaves we wrap all generics in the axi_slave_t record.

vunit/vhdl/verification_components/src/wishbone_slave.vhd
```diff
@@ -0,0 +1,126 @@
...
+use work.wishbone_pkg.all;
+
+library osvvm;
+use osvvm.RandomPkg.all;
+
+entity wishbone_slave is
+  generic (
+    wishbone_slave : wishbone_slave_t;
+    max_ack_dly : natural := 0;
+    rand_stall  : boolean := false
```
kraigher: Rename max_ack_dly and rand_stall to initial_max_ack_delay and initial_rand_stall_probability : real and move them into wishbone_slave_t to make it forward compatible with changing them at runtime and more similar to axi_slave_t.

vunit/vhdl/verification_components/src/wishbone_pkg.vhd
```diff
@@ -0,0 +1,45 @@
...
+  impure function new_wishbone_slave(
+    memory : memory_t;
+    logger : logger_t := wishbone_slave_logger)
+    return wishbone_slave_t;
+
+end package;
+package body wishbone_pkg is
+
+  impure function new_wishbone_slave(
+    memory : memory_t;
```
kraigher: Add max_ack_delay and random_stall_probability parameters here and to wishbone_slave_t to make it similar to axi_slave_t and forward compatible with setting them at run-time.

slaweksiluk: Ok, I will go that way. I will also change both parameters to be probability of having value 1, E.g:
```
init_stall_high_probability : real := 0.0;
init_ack_high_probability : real := 1.0;
```
It seems better defined this way. I also consider adding random stb signal stimulation option to master VC.

kraigher:
Great work. I am ready to accept this pull request. You just need to fix the broken checks found by Travis CI: https://travis-ci.org/VUnit/vunit/jobs/352852061

I think the only problem is the Python PEP8 code style checks that are broken by the changes in the verification_components run.py file.

You can run this check locally using `pytest vunit/test/lint/test_pep8.py`.
First you need to install pytest and pep8 using `pip install pytest pep8`

kraigher:
@slaweksiluk Do you need any help finishing this up?

slaweksiluk:
I fixed coding style errors. I was using designed bfm's in my project to verify it in real conditions. Wishbone slave looks good, but I haven't got time to try wishbone master. Moreover I remember I saw small bug in wishbone master regarding cyc signal one clock cycle late termination. However it only occurs at the end of block cycle, so it is not harmful. I will create issues for it when I trace it. I think you should mentioned somewhere in documentation that wishbone bfms are still in testing.

kraigher:
We write in the documentation that the entire verification component library is in beta so maybe the maturity of wishbone needs no additional disclaimer.

Issue #310, opened 2018-02-21
TLAST not required on every axi_stream_slave transaction


jklapel:
Currently, every transaction of the axi_stream_slave requires that TLAST be set to '1', or else a check in the axi_stream_slave entity will fail, causing a failure of the VUnit test. However, this TLAST is not a requirement of the AXIS protocol and should only come on the last data beat of a packetized data stream.

As a workaround, I commented out the check in axi_stream_slave, as well as added a pop_axi_stream procedure (similar to the push_axi_stream procedure) that returns TLAST so that I can check/act on it myself.

kraigher:
So we could add a last-concept to the generic stream master/slave VCI. 

Making the following changes to stream_master:
```vhdl
  -- NOTE: Added last parameter
  -- Push a data value to the stream
  procedure push_stream(signal net : inout network_t;
                        stream : stream_master_t;
                        data : std_logic_vector;
                        last : boolean := false);
```

Making the following additions/changes to stream_slave:
```vhdl
  -- NOTE: New procedure in addition to old
  procedure pop_stream(signal net : inout network_t;
                       stream : stream_slave_t;
                       variable data : out std_logic_vector;
                       variable last : out boolean);

  -- NOTE: New procedure in addition to old
  procedure await_pop_stream_reply(signal net : inout network_t;
                                   variable reference : inout stream_reference_t;
                                   variable data : out std_logic_vector;
                                   variable last : out boolean);

  -- NOTE: Added last parameter
  -- Blocking: read stream and check result against expected value
  procedure check_stream(signal net : inout network_t;
                         stream : stream_slave_t;
                         expected : std_logic_vector;
                         last : boolean := false;
                         msg : string := "");
```

jklapel:
That would work for my immediate needs.  (Having the rest of the AXI Stream signals would be useful as well, but that's likely a separate issue.)

Would this make the current push_axi_stream redundant then?

kraigher:
@jklapel Yes it would make the AXI-stream specific push-method redundant until we want to add AXI specific flags to the push/pop commands such as tstrb or tkeep or tuser.

kraigher:
This is a good first issue if you want to make a pull request to VUnit.

jklapel:
@kraigher I created a PR that adds in the suggestions you mentioned above.  I'd appreciate any feedback you have.

jklapel:
This has been resolved with PR #314

PR #314, opened 2018-03-06
Generic stream last

jklapel:
This adds in the last signal to the generic stream VCI, as discussed in https://github.com/VUnit/vunit/issues/310

### DIFF
--- a/examples/vhdl/uart/src/test/tb_uart_rx.vhd
+++ b/examples/vhdl/uart/src/test/tb_uart_rx.vhd
@@ -52,7 +52,7 @@ begin
 
       elsif run("test_receives_one_byte") then
         push_stream(net, uart_stream, x"77");
-        check_stream(net, axi_stream, x"77");
+        check_stream(net, axi_stream, x"77",true);
         wait until rising_edge(clk);
         check_equal(tvalid, '0');
         check_equal(num_overflows, 0);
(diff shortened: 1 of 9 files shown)


kraigher:
The uart rx/tx VC was not updated according with the generic stream interface and that is why the tests fail. See https://travis-ci.org/VUnit/vunit/jobs/349952989

you can run `examples/vhdl/uart/run.py` and `vunit/vhdl/verification_components/run.py` to re-create

kraigher:
It looks good. I think it could be of value to add a pop_stream and await_pop_stream_reply variant that does not need a last output variable for convenience. For example for an UART VC that uses the stream interface and has no use for last.

jklapel:
I added the pop_stream and await_pop_stream convenience procedures in my local repo, which I think solved the primary UART issue.  However, when I run the verification components testbench, it fails on the "vunit_lib.tb_axi_stream.test stream pop expects tlast" test (as well as on another test that I haven't looked at yet).  If I am understanding how the test was originally written, the new functionality should break this test and it should therefore be removed.  In its place, I should probably add a couple tests that use the new procedures that have been written.  Am I correcting in my understanding on this test?

kraigher:
Yes the test you mention is not relevant any more since tlast is not automatically expected any more.

Issue #324, opened 2018-04-18
Let VUnit execute simulators which are inside a container or VM, from outside


1138-4EB:
Because VUnit requires the simulator to be available in the path, and executable names (vsim, vcom, etc.) are 'hardcoded', VUnit must be installed in the same docker image as the simulator. In a use case, I have two images, one with GHDL and a different one with QuestaSim, and VUnit is installed in both of them. Indeed, because I have versions with and without VUnit, there are four images. Each time VUnit is updated, two of them have to be rebuilt.

My proposal is to make VUnit execute simulators which are inside a container, from outside:

- Have a local python installation, but do not install simulators in the host.
  - This would allow you not to compile GHDL in each test, but still test multiple python versions with the latest GHDL in travis.
- Build a single VUnit docker image that can interact with simulator containers.

In the latter, I would have three images instead of four, and I would have to rebuild a single one. Note how this would scale with more simulators. Furthermore, if the VUnit docker image was generated in the vunit travis-ci flow, I would consume it as an alternative to `ghdl/ext:vunit`.

- Is it possible to prepend (not append) args to every vsim, vcom, ghdl call?
- Is it possible to hook commands (to create/kill temporal containers) before and after each 'task'? By 'task' I mean a sequence a commands to be executed in the same container.

A workaround would be to 'hack' it at system level. That is, write some man-in-the-middle shell scripts to get VUnit's calls and forward them to containers. However, it is not clean, because it is not straighforward to handle the lifecycle of tests as VUnit does.

> @kraigher (Gitter):
> @1138-4EB Regarding your use case about customizing the command for running vsim etc. it has some overlap with the use case of running a batch server farm with bsub if you are familiar with that.
> Anyway such things are easier to discuss in an issue rather than in this chat.

1138-4EB:
@kraigher , I've checked bsub and, although similar, it is overkill. Because it is meant for HPC with multiple physical devices, an additional orchestration service (LSF) must be run. My proposal is more straighforward because all the containers are expected to be executed in the same machine, so no additional service is required.

Well, docker is the service, somehow, as `docker exec` works for both local containers and services in a swarm cluster. But I'll focus on the local case, which is easier to follow. Four functionally equivalent examples are shown [below](#examples):

1. All the commands at the same time with local tools.
2. All the commands at the same time in a temporal container (same as #323, but without VUnit).
3. All the commands at the same time in a named container running in the background.
4. Commands one-by-one in a named container running in the background.

[...]

PR #327, opened 2018-04-19
ensure quoted generics

1138-4EB:
Fixes a quoting issue when a script is used between VUnit and the executables. See [#324](https://github.com/VUnit/vunit/issues/324#issuecomment-382729779).


kraigher:
This does not work. You cannot add quotes since VUnit sets argv of the subprocess directly without involving any shell which is more robust. The argv that your bash script received is exactly the one that you need to set for the subprocess.

PR #329, opened 2018-04-23
fix delayed simulator selection

1138-4EB:
In [this](https://github.com/VUnit/vunit/commit/8cf71ea4389bb6c6c1eafca63e5b111964d5ee41) commit the simulator selection is delayed until VUnit class instantiation, instead of executing it during the package import. However, the `simulator_factory` is still created when the package is imported. Therefore, simulator detection is executed on the PATH available at import time, even if the selection is done later.

This PR moves content of then `__init__` function/method from `simulator_factory` to a new function named `detect_available_simulators`. The new function is called first in `__init__` and once again in `select_simulator`.

### DIFF
--- a/vunit/simulator_factory.py
+++ b/vunit/simulator_factory.py
@@ -101,14 +101,13 @@ class SimulatorFactory(object):
         Select simulator class, either from VUNIT_SIMULATOR environment variable
         or the first available
         """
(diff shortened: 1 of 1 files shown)


kraigher:
I made some changes before merging.

PR #335, opened 2018-05-21
Add JSON-for-VHDL

1138-4EB:
This PR adds [Paebbels/JSON-for-VHDL](https://github.com/Paebbels/JSON-for-VHDL) as a git submodule. As talked in Gitter, this is a proof of concept to possibly replace the `runner_cfg` enconding with JSON. Right now, it is just an alternative to complex generics that supports any number of depth in the content structure.

Method `add_json4vhdl` is added, in order to easily load the library. Also, `set_json_generic` is provided, to provide a cleaner syntax.

In Paebbels/JSON-for-VHDL#4 support for stringified generics was added (see also Paebbels/JSON-for-VHDL#3). Therefore, there are two options to pass data from Python to VHDL: a file and a string generic. A example is added, where both approaches are used. On the one hand, the filename is passed as a generic. On the other hand, the content is read in the `run.py` file (where it can be optionally modified), and it is stringified to a different generic.

Note that there is a very similar example in https://github.com/Paebbels/JSON-for-VHDL/tree/master/VUnit Therefore, I don't know whether it makes sense to keep both. Maybe we should keep one only.


vunit/ui.py
```diff
@@ -447,6 +447,26 @@ def set_generic(self, name, value, allow_empty=False):
         for test_bench in check_not_empty(test_benches, allow_empty, "No test benches found"):
             test_bench.set_generic(name.lower(), value)
 
+    def set_json_generic(self, name, value, allow_empty=False):
```
kraigher: I think it would be better to add a function such as encode_json to make this more orthogonal and avoid having a different variant for each set_generic/set_parameter on the TestBench, Library etc. . Also the add_config method has generic parameter which cannot make use of a set_json_generic.

LarsAsplund:
@1138-4EB I think you should add an example for the python list and dictionary use cases discussed in the post.

1138-4EB:
@LarsAsplund the current example JSON file includes an array named `Image`, which is equivalent to `image_resolution = [640, 480]` in https://vunit.github.io/posts/2017_06_03_enable_your_simulator_to_handle_complex_top_level_generics/post.html. I now added a `decode` function (VHDL) in order to fill a record. To show the versatility of the approach, `dump_debug_data` is added in `run.py` to the content read from the JSON file. See:

- https://github.com/1138-4EB/vunit/blob/json_generics/examples/vhdl/json4vhdl/src/test/tb_json_gens.vhd#L24-L37
- https://github.com/VUnit/vunit/blob/master/examples/vhdl/composite_generics/test/tb_composite_generics.vhd#L19-L32

However, I think that reading a list/array of integers as done with `split(<content>, ", ")` is not supported in JSON-for-VHDL, yet. This is because, AFAIK, the length of the arrays/objects is not saved when the JSON content is parsed. But I think that it might be useful to allow reading arrays of integers as a comma separated string. I tried reading `jsonGetString(Content, "Image")` and `jsonGetString(JSONContent, "Platform/KC705/IIC/0")`. The first one returns `640` (i.e., the first element in the array). The second one returns `ERROR`. @Paebbels might confirm this.

1138-4EB:
I added a workaround for arrays of integers. A recursive function in pyhton adds the length as the first element of the array. Then, a generic `decode_array` function in VHDL sequentially reads the elements from the compressed JSOn representation.

Also, I reorganized the content. `vunit/json4vhdl.py` contains `read_json`, `encode_json`, and `add_array_lens`. Dir `vunit/vhdl/json4vhdl` contains a `json4vhdl_context` and the `decode_array` function in a package named `json4vhdl_decode`.

vunit/json4vhdl.py
```diff
@@ -0,0 +1,62 @@
...
+
+    .. code-block:: python
+
+       generics = read_json(join(root, "src/test/data/data.json"))
+    """
+    return json.loads(open(filename, 'r').read())
+
+
+def add_array_lens(obj):
+    """
```
kraigher: add_array_lens should not be added.

vunit/vhdl/json4vhdl/json4vhdl_decode.vhd
```diff
@@ -0,0 +1,28 @@
...
+context vunit_lib.vunit_context;
+
+use vunit_lib.json.T_JSON;
+use vunit_lib.json.jsonGetString;
+
+package json4vhdl_decode is
+  impure function decode_array(Content : T_JSON; Path : string) return integer_vector;
+end;
+
+package body json4vhdl_decode is
```
kraigher: this package should not be added. It should be fixed in json for VHDL.

vunit/vhdl/json4vhdl/json4vhdl_context.vhd
```diff
@@ -0,0 +1,14 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this file,
+-- You can obtain one at http://mozilla.org/MPL/2.0/.
+--
+-- Copyright (c) 2014-2018, Lars Asplund lars.anders.asplund@gmail.com
+
+context json4vhdl_context is
```
kraigher: Is this context necessary if we remove the decode package.

1138-4EB: Yes. It is used anyway, in order to provide `T_JSON`, `jsonLoad`, `jsonGetString` and `jsonGetBoolean`, which are used in the `decode` function in the testbench (not the `decode_array` in the package).

vunit/json4vhdl.py
1138-4EB: Does this mean that it should be added to JSON-for-VHDL prior to merging this PR? Or will this go forward with a note explaining that the user must ensure that the first element in each array is the length of it?

vunit/vhdl/json4vhdl/json4vhdl_context.vhd
kraigher: But why should VUnit add a json4vhdl context on top of json4vhdl? If such a context is useful it should be added directly to json4vhdl.

1138-4EB: JSON-for-VHDL is compiled to library `vunit_lib`, in order to allow this syntax in the testbench:

1138-4EB: JSON-for-VHDL is compiled to library `vunit_lib`, in order to allow this syntax in the testbench:

vunit/json4vhdl.py
kraigher: I think either fix it in json4vhdl first or do not have an example that needs dynamically sized arrays.
I do not think we should add work-arounds to the Public API of VUnit just because a feature is missing in another open source library. 

If you still want an example of this now even before the feature is added to json4vhdl it could be a bit more clear not add the array length as an element of the array but as a separate field in parallel with the array such as.

```json
{
"length": 3,
"data": [1,2,3]
}
```

vunit/vhdl/json4vhdl/json4vhdl_context.vhd
kraigher: I do not think json4vhdl should be compiled into vunit_lib. Ideally each library should have a standard library name such that code can be written in a portable way. For example osvvm always named osvmm and vunit always named vunit_lib. json4vhdl would also be better served by having a standard library name. Thus the user can write code that does not have to change depending if json4vhdl is in vunit_lib or in a library called json4vhdl.

1138-4EB: JSON-for-VHDL is compiled to library `vunit_lib`, in order to allow this syntax in the testbench:

```
library vunit_lib;
context vunit_lib.json4vhdl_context;
```

If I move json4vhdl_context to JSON-for-VHDL, will using `work` as the library inside the context allow to keep the syntax above?

kraigher: json4vhdl is not a part of vunit and should thus not be in vunit_lib. Just like osvvm it should be a separate 3rd party library.

kraigher: I do not think "work" works within context clauses. That is why it is important to globally define a standard library name. Maybe when VHDL gets support for hierarchical libraries things can become more flexible.

kraigher: Have a look at OSVVM: https://github.com/OSVVM/OSVVM/blob/master/OsvvmContext.vhd

kraigher: A good library name for json4vhdl would be "json"

kraigher: But it is really up to the project maintainer of json4vhdl to define this standard name.

vunit/builtins.py
```diff
@@ -147,6 +148,20 @@ def _add_osvvm(self):
 
             library.add_source_files(file_name, preprocessors=[])
 
+    def _add_json4vhdl(self):
+        """
+        Add JSON-for-VHDL library
+        """
+        library_name = "vunit_lib"
```
kraigher: Do not add json4vhdl to vunit_lib, it is 3rd party

vunit/vhdl/json4vhdl/json4vhdl_context.vhd
1138-4EB: Paebbels/JSON-for-VHDL#6

1138-4EB:
- I moved the `json_context` to the upstream project: Paebbels/JSON-for-VHDL#6.
- JSON-for-VHDL is no longer compiled in `vunit_lib`, but in library `JSON`.
- I removed all the `integer_vector` related code from here.
  - I added the prototype of a future update, once `jsonGetIntArray` is properly implemented in the upstream, to a separte branch: https://github.com/1138-4EB/vunit/compare/json_generics...1138-4EB:json-IntArray
- 

@LarsAsplund, the dictionary/record example is ready to use. The list is not, yet. See comments above.

@kraigher this will fail until Paebbels/JSON-for-VHDL#6 is merged, due to the context file not being found in the current master branch.

> To be honest I think the array length problem should be solved upstream in the json4vhdl package and not in VUnit. I do not want to have layers on top of json4vhdl that would later be redundant. Then I would rather wait for it to be nicely supported directly by json4vhdl.
>
> Thus I do not think we should add any context or decode package for json4vhdl.

This was a misunderstanding on my part. I thought that the idea was to integrate JSON-for-VHDL as part of VUnit in the long term, since the current implementation of JSON-for-VHDL is optimized for synthesis and VUnit would take advantage of a variant optimized for simulation. Therefore, I conceived `json4vhdl` to be the simulation-friendly version, which would be part of VUnit, and JSON-for-VHDL would be presented to the user as limited alternative meant for hardware.

However, after the talk on gitter, I rethought it to always be an external dependecy. Even when/if it becomes a hard dependency for the core of VUnit, and not just an optional feature.

Yet, `json4vhdl` is still present in this PR as:

- The `add_json4vhdl` method, to load JSON-for-VHDL in `run.py`. This will not be required if/when JSON-for-VHDL is used for `runner_cfg`.
- The name of the example.
- A python script in the vunit codebase that implements `read_json` and `encode_json`.

Is it ok to use it like this? Or should we also move `json4vhdl.py` to the upstream project?

1138-4EB:
@kraigher, @LarsAsplund the PRs in the upstream project are merged now. So, both `json_ctx` and `jsonGetIntegerArray` are available. Therefore, I think that this PR meets all your requirements. However, let me know if you want any addition/modification.

`json4vhdl` is still present in this PR as:

- The `add_json4vhdl` method, to load JSON-for-VHDL in `run.py`. This will not be required if/when JSON-for-VHDL is used for `runner_cfg`.
- The name of the example.
- A python script in the vunit codebase that implements `read_json` and `encode_json`.

Is it ok to use it like this? Or should we also move `json4vhdl.py` to the upstream project?

PR #349, opened 2018-06-25
Adding csv mapping support for files and libraries

davidmedinasigasi:


vunit/ui.py
```diff
@@ -393,6 +394,36 @@ def add_external_library(self, library_name, path, vhdl_standard=None):
...
+                        libs_and_files[lib_name] += [file_name]
+                    else:
+                        libs_and_files[lib_name] = [file_name]
+                elif len(row) > 2:
+                    LOGGER.error("More than one library and one file in csv description")
+    
+            for lib_name in libs_and_files.keys():
+                files = libs_and_files[lib_name]
+                lib = self.add_library(lib_name)
+                lib.add_source_files(files)
```
kraigher: Other functions that add files return the list of files added.

vunit/ui.py
```diff
@@ -393,6 +394,36 @@ def add_external_library(self, library_name, path, vhdl_standard=None):
...
+            vhdl_standard = self._vhdl_standard
+        
+        libs_and_files = dict()
+
+        with open(project_csv_path) as f:
+            content = csv.reader(f)
+            for row in content:
+                if len(row) == 2: 
+                    lib_name = row[0].strip()
+                    file_name = row[1].strip()
```
kraigher: It would be reasonable to interpret the file name as relative to the location of the .csv file or an absolute path rather than relative to the current working directory when this code is executing,

vunit/ui.py
```diff
@@ -393,6 +394,36 @@ def add_external_library(self, library_name, path, vhdl_standard=None):
         self._project.add_library(library_name, abspath(path), vhdl_standard, is_external=True)
         return self.library(library_name)
 
+    def add_csv(self, project_csv_path, vhdl_standard=None):
```
kraigher: add_source_files_from_csv would be a better name

vunit/ui.py
```diff
@@ -393,6 +394,36 @@ def add_external_library(self, library_name, path, vhdl_standard=None):
...
+        """
+        Add a project configuration, mapping all the libraries and files
```
kraigher: The description does not describe the format of the csv file or the meaning of relative paths of the file names.

1138-4EB: I have some WIP related to a similar approach with JSON:

- https://github.com/1138-4EB/vunit/tree/json_cfg_run

where the 'generic' `run-py` contains:

- https://github.com/1138-4EB/vunit/blob/f67be68b6e8a114a69f1ac67dba92b6f79de11c2/examples/vhdl/array_axis_vcs/cfg_run.py#L16-L20
- https://github.com/1138-4EB/vunit/blob/f67be68b6e8a114a69f1ac67dba92b6f79de11c2/examples/vhdl/array_axis_vcs/cfg_run.py#L35-L48

and the example JSON itself is:

- https://github.com/1138-4EB/vunit/commit/f67be68b6e8a114a69f1ac67dba92b6f79de11c2#diff-c8cec296822f0068205fef2fce6117ac

I'd be glad if the format/approach was explicitly documented, in order to follow the same pattern when I return to it.

vunit/ui.py
```diff
@@ -422,7 +424,7 @@ def add_csv(self, project_csv_path, vhdl_standard=None):
             for lib_name in libs_and_files.keys():
                 files = libs_and_files[lib_name]
                 lib = self.add_library(lib_name)
-                lib.add_source_files(files)
+                return lib.add_source_files(files)
```
kraigher: bug: return inside for-loop only works for single library in the csv file

vunit/ui.py
kraigher: What I mean with the file name in this context is the file name that is written inside the csv file. 

For example if a have the following structure:
/directory/proj.csv
/directory/file.vhd
/run.py

If proj.csv contains "lib, file.vhd" and I run "python run.py" in the root directory it should refer to /directory/proj.csv making the semantics of the proj.csv file independent on the current working directory where the run.py script is executed.

davidmedinasigasi: I understand what you want me to do, I just don't know how, can you please explain a little more?

kraigher: You can achieve what I describe by. 
```python
from os.path import normpath, join, dirname
normpath(join(dirname(csv_file_name), file_name))

It will make file_name relative to the parent directory of the csv file unless file name is an absolute path already.
```

vunit/ui.py
```diff
@@ -405,26 +405,28 @@ def add_source_files_from_csv(self, project_csv_path, vhdl_standard=None):
...
                         libs_and_files[lib_name] = [file_name]
                 elif len(row) > 2:
                     LOGGER.error("More than one library and one file in csv description")
     
+            list_of_source_files = list()
             for lib_name in libs_and_files.keys():
                 files = libs_and_files[lib_name]
                 lib = self.add_library(lib_name)
-                return lib.add_source_files(files)
+                list_of_source_files.append(lib.add_source_files(files))
```
kraigher: This will create a list of lists and not a single list of all files added.

vunit/ui.py
```diff
@@ -426,7 +426,10 @@ def add_source_files_from_csv(self, project_csv_path, vhdl_standard=None):
                 files = libs_and_files[lib_name]
                 lib = self.add_library(lib_name)
                 list_of_source_files.append(lib.add_source_files(files))
-            return list_of_source_files
+            
+            list_of_source_files_flattened = [val for sublist in list_of_source_files for val in sublist]
```
kraigher: += would have been nicer. Also it  seems this should have a unit test since there were so many bugs.

vunit/test/unit/test_ui.py
```diff
@@ -255,6 +255,52 @@ def test_add_source_files(self):
...
+        
+        for i in range(len(list_of_files)):
+            self.create_file(list_of_files[i], str(i)) 
+        
+        self.create_csv_file('test_returns.csv', csv)
+        ui = self._create_ui()
+
+        source_files = ui.add_source_files_from_csv('test_returns.csv')
+        
+        for i in range(len(source_files)):
```
kraigher: This test is flawed. source_files could just be empty or miss the last element and the test would still pass. Use list_of_files as the iteration length source instead. Or even better just compare the lists directly with` self.assertEqual([source_file.name for source_file in source_files, list_of_files)`

vunit/ui.py
```diff
@@ -393,6 +394,41 @@ def add_external_library(self, library_name, path, vhdl_standard=None):
         self._project.add_library(library_name, abspath(path), vhdl_standard, is_external=True)
         return self.library(library_name)
 
+    def add_source_files_from_csv(self, project_csv_path, vhdl_standard=None):
+        """
+        Add a project configuration, mapping all the libraries and files
+        :param project_csv_path: path to csv project specification, each line contains the name
```
kraigher: The documentation should also say that relative file names are relative to the parent folder of the csv file.

vunit/ui.py
```diff
@@ -393,6 +394,41 @@ def add_external_library(self, library_name, path, vhdl_standard=None):
...
+                    no_normalized_file = row[1].strip()
+                    file_name = normpath(join(dirname(project_csv_path), no_normalized_file))
+                    if lib_name in libs_and_files:
+                        libs_and_files[lib_name] += [file_name]
+                    else:
+                        libs_and_files[lib_name] = [file_name]
+                elif len(row) > 2:
+                    LOGGER.error("More than one library and one file in csv description")
+    
+            list_of_source_files = list()
```
kraigher: = SourceFileList()

Otherwise the return value will be a list() and not a SourceFileList like the documentation claims.

kraigher:
Good now you just have to fix the lint violations to make the Travis CI check pass.
You can run pylint and pycodestyle locally as decribed here: http://vunit.github.io/contributing.html#running-the-tests
```
pytest vunit/test/lint/
```

vunit/test/unit/test_ui.py
```diff
@@ -255,6 +255,46 @@ def test_add_source_files(self):
...
+        self.create_csv_file('test_csv.csv', csv)
+        for file_name in files:
+            self.create_file(file_name)
+
+        ui = self._create_ui()
+        ui.add_source_files_from_csv('test_csv.csv')
+
+        for index, library_name in enumerate(libraries):
+            file_name = files[index]
+            file_name_from_ui = ui.add_source_files_from_csv(file_name, library_name)
```
kraigher: This line does not make sense. You add the file_name as a csv file but the file is an empty vhdl_file.
Did you mean to do get_source_file?

vunit/test/unit/test_ui.py
```diff
@@ -255,6 +255,46 @@ def test_add_source_files(self):
...
 
+    def test_add_source_files_from_csv(self):
+        csv = """
+        lib,  tb_example.vhdl
+        lib1 , tb_example1.vhd
+        lib2, tb_example2.vhd
+        """
+
+        libraries = ['lib', 'lib1', 'lib2']
+        files = ['tb_example.vhdl', 'tb_example1.vhd', 'tb_example2.vhd', ' tb,ex3.vhd']
```
kraigher: tb,ext3.vhd is unnessary?

vunit/ui.py
```diff
@@ -393,6 +394,44 @@ def add_external_library(self, library_name, path, vhdl_standard=None):
...
+            file_before = None
+            for row in content:
+                if len(row) == 2:
+                    lib_name = row[0].strip()
+                    no_normalized_file = row[1].strip()
+                    file_name_ = normpath(join(dirname(project_csv_path), no_normalized_file))
+                    lib = self.library(lib_name) if lib_name in libs else self.add_library(lib_name)
+                    libs.add(lib_name)
+                    file_ = lib.add_source_file(file_name_)
+                    if file_before is not None:
```
kraigher: This circumvents the dependency scanning performed by VUnit, why are you creating manual dependencies? VUnit will still can the files added for dependencies unless you add the no_parse argument to add_source_file.

vunit/ui.py
```diff
@@ -393,6 +394,46 @@ def add_external_library(self, library_name, path, vhdl_standard=None):
         self._project.add_library(library_name, abspath(path), vhdl_standard, is_external=True)
         return self.library(library_name)
 
+    def add_source_files_from_csv(self, project_csv_path, vhdl_standard=None, is_in_order=True):
```
kraigher: The default should not be to disable the VUnit dependency scanning.

vunit/ui.py
```diff
@@ -393,6 +394,46 @@ def add_external_library(self, library_name, path, vhdl_standard=None):
...
+        with open(project_csv_path) as csv_path_file:
+            file_before = None
+            for row in csv.reader(csv_path_file):
+                if len(row) == 2:
+                    lib_name = row[0].strip()
+                    no_normalized_file = row[1].strip()
+                    file_name_ = normpath(join(dirname(project_csv_path), no_normalized_file))
+                    lib = self.library(lib_name) if lib_name in libs else self.add_library(lib_name)
+                    libs.add(lib_name)
+                    file_ = lib.add_source_file(file_name_, no_parse=is_in_order)
```
kraigher: No parse will also inhibit any test_benches/tests from being detected from the source file.

davidmedinasigasi: When sigasi generates the csv file, we resolve all dependencies in a way that the first file is a dependency for the next one, for example
lib, example.vhd
lib2, example2.vhd  // This one depends on lib, example.vhd
lib, example3.vhd    // This one depends on lib2, example2.vhd
lib3, example4.vhd  // This one depends on lib, example3.vhd

kraigher: Dependencies are never linear they are a tree. To force them to be linear creates a lot of unnecessary re-compilation.

kraigher: Consider the case with 200 files with 20 groups of 10 files. Each group has a linear dependency but the groups are independent. If you can only express a linear order some group will be first and some group will be last. If a file in the first group is modified it causes 200 files to be recompiled instead of only 10.

kraigher: Sigasi for sure has a dependency tree internally. It is only when you need to generate a sequence of compile commands for the initial compile that it is useful to generate an ordered list of files. For subsequent *incremental* compiles which is a big feature of VUnit a new ordered list of compile commands has to generated based on the files changes and the dependency tree.

kraigher:
The function to get built in libraries should not be a part of this pull request. It should be reviewed separately. You need to use feature branches in your fork to separate the issues.

PR #350, opened 2018-06-28
Added vhdl_standard attribute at class initialization

FranzForstmayr:
This change allows to select a proper vhdl standard at initializion, for easier switching between multiple projects using multiple standards

### DIFF
--- a/vunit/ui.py
+++ b/vunit/ui.py
@@ -1805,7 +1805,7 @@ def select_vhdl_standard(vhdl_standard=None):
     """
     Select VHDL standard either from class initialization or according to environment variable VUNIT_VHDL_STANDARD
     """
-    if vhdl_standard:
+    if vhdl_standard is not None:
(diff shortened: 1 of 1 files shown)


vunit/ui.py
```diff
@@ -1801,12 +1801,16 @@ def add_dependency_on(self, source_file):
...
 
-def select_vhdl_standard():
+def select_vhdl_standard(vhdl_standard=None):
     """
-    Select VHDL standard according to environment variable VUNIT_VHDL_STANDARD
+    Select VHDL standard either from class initialization or according to environment variable VUNIT_VHDL_STANDARD
     """
-    vhdl_standard = os.environ.get('VUNIT_VHDL_STANDARD', '2008')
-    check_vhdl_standard(vhdl_standard, from_str="VUNIT_VHDL_STANDARD environment variable")
+    if vhdl_standard:
```
kraigher: Please use "if x is not None" rather than "if x".

PR #351, opened 2018-06-29
Adding functions to get all built in libraries

davidmedinasigasi:


vunit/ui.py
```diff
@@ -1023,6 +1023,9 @@ def get_implementation_subset(self, source_files):
         return SourceFileList([SourceFile(source_file, self._project, self)
                                for source_file in source_files])
 
+    def get_builtins(self):
```
kraigher: I do not want to make the Builtins class public. Just add one method to get the built_in_libraries directly.

vunit/builtins.py
```diff
@@ -232,3 +241,6 @@ def _add_check(self, name, args=None):
                 "Optional builtin %r added with arguments %r has already been added with arguments %r"
                 % (name, args, old_args))
         return True
+
```
kraigher: This function does not need to be public in the Builtins class (Add _ to make private) in fact this funciton is not necessary at all.

davidmedinasigasi: ` Access to a protected member _get_types of a client class (protected-access)` 
Can you explain why is not necessary?

kraigher: Ok they where different classes so I was wrong. It needs to be public.

vunit/ui.py
```diff
@@ -1023,6 +1023,9 @@ def get_implementation_subset(self, source_files):
         return SourceFileList([SourceFile(source_file, self._project, self)
                                for source_file in source_files])
 
+    def get_built_in_libraries(self):
```
kraigher: Functions on Public API should have documenation. Add documentation of the return value. Also since the function returns a set of names and not library object like other methods on VUnit class it should be named get_built_in_library_names.

kraigher: You added the documentation in the wrong place. It should be added on the VUnit method not the Builtins method. The Builtins method is not part of the public vunit api. You can keep the documentation of the Builtins method since it does not hurt but it should be also added to the VUnit class method.

kraigher:
Actually this does not do what you want it to. The type names that you use are not actually library names. All VUnit built-in libraries are added to vunit_lib even if they are optional. Then there are two third party libraries that we ship (osvvm, json4vhdl) that can optionally be added under separate library names.

kraigher:
As I said earlier array_util, com etc are not separate libraries even though they are optional they are still compiled into vunit_lib. Thus the method you have here does not work. You have to consider what you want to achieve in the end. Why do you need to know all the built-in libraries? Is not enough to simply know all files added in the run.py regardless if they are "build-in" or not. If you remove the files you provide to the run.py file from the .csv-file from the files reported with --files effectively get the non-sigasi managed files.

lievenlemiengre:
We're exploring your suggestion @kraigher and we're getting good results with it.

Basically anything in the csv file will be editable by Sigasi all other files are added as read-only & updated whenever the vunit python description changes.

@davidmedinasigasi you can close this pull request

PR #355, opened 2018-07-12
Show terminated test as 'skipped'

Witnessthis:
Use case:
User starts many simulations, one of the tests hang, the user terminates the tests using 'ctrl + c'.

This commit will show the terminated test as 'skipped' in the test report, making it easier to see which test that failed.

### DIFF
--- a/vunit/test_runner.py
+++ b/vunit/test_runner.py
@@ -21,7 +21,7 @@ import logging
 import string
 from contextlib import contextmanager
 import vunit.ostools as ostools
-from vunit.test_report import PASSED, FAILED
+from vunit.test_report import PASSED, FAILED, SKIPPED
 from vunit.hashing import hash_string
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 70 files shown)


vunit/test_runner.py
```diff
@@ -210,6 +210,8 @@ def read_output():
             results = test_suite.run(output_path=output_path,
                                      read_output=read_output)
         except KeyboardInterrupt:
+            results[test_suite.name] = SKIPPED
```
kraigher: It is not test_suites but rather test_cases within test_suites that have test report results. When running several tests within the same simulation there are multiple test_cases within a test_suite. See the code that prints the test status using a for loop over test cases within the test suite.

vunit/test_runner.py
```diff
@@ -210,6 +210,9 @@ def read_output():
             results = test_suite.run(output_path=output_path,
                                      read_output=read_output)
         except KeyboardInterrupt:
+            for name in test_suite.test_cases:
+                results[name] = SKIPPED
+                self._add_results(test_suite, results, start_time, num_tests, output_file_name)
```
kraigher: _add_result should be called once for the test suite and not once for every test case.

PR #359, opened 2018-07-23
Avalon MM BFM

slaweksiluk:
Adds basic Avalon MM Pipelined Master and Slave Verification Components based on the existing Wishbone Master and Slave. Burst transfer not yet supported (only fixed burst = 1).

Also included two changes in existing wishbone modules:
  - adds strobe randomization in Wishbone Master (as VC entity generic)
  - Wishbone Slave drives 'stall' signal independent from 'cyc' state


kraigher:
There where just some pylint warnings. Do you know how to fix them?

slaweksiluk:
I suppose I need to run pytest vunit/test/lint/test_pep8.py?

PR #364, opened 2018-08-01
Support for Avalon-MM burst transfers

slaweksiluk:
### Master
For support for burst transfers current bus_master had to be extended and some way of passing multiple data words had to be chosen. I decided to use VUnit queues as it seems the most natural for me. Please comment if somebody see the better options. Currently data words from queue are pushed to bus burst msg, however it is also possible to push queue reference instead of single data chunks.

In bursts transactions byteenable signals is always high - probably need to extend it.

### Slave
Avalon Slave was rewritten - write and read handling is separated.  New implementation does not prevent simultaneous write and read .The another event which should be forbidden is mixed write/read bursts. In general bus monitor should be added to capture and report  such a invalid transactions.

### DIFF
--- a/vunit/vhdl/verification_components/src/bus_master_pkg.vhd
+++ b/vunit/vhdl/verification_components/src/bus_master_pkg.vhd
@@ -71,6 +71,10 @@ package bus_master_pkg is
                       constant data : std_logic_vector;
                       -- default byte enable is all bytes
                       constant byte_enable : std_logic_vector := "");
+  -- Procedures for burst bus write: Caller is responsible for allocation and
+  -- deallocation of burstdata queue. Procedure cunsumes burst_length data words
+  -- from burstdata queue. If burstdata queue has less data words, all data
+  -- words are consumed and pop from empty queue error is raised.
   procedure write_bus(signal net : inout network_t;
                       constant bus_handle : bus_master_t;
                       constant address : std_logic_vector;
@@ -106,6 +110,9 @@ package bus_master_pkg is
   procedure await_read_bus_reply(signal net : inout network_t;
                                  variable reference : inout bus_reference_t;
                                  variable data : inout std_logic_vector);
+  -- Procedure for burst read reply: Caller is responsible for allocation and
+  -- deallocation of burstdata queue. Procedure pushes burst_length data words
+  -- into burstdata queue.
   procedure await_read_bus_reply(signal net : inout network_t;
                                  constant bus_handle : bus_master_t;
                                  constant burstdata : queue_t;
@@ -132,6 +139,9 @@ package bus_master_pkg is
                      constant bus_handle : bus_master_t;
                      constant address : natural;
                      variable data : inout std_logic_vector);
+  -- Procedure for burst bus read: Caller is responsible for allocation and
+  -- deallocation of burstdata queue. Procedure pushes burst_length data words
+  -- into burstdata queue.
   procedure read_bus(signal net : inout network_t;
                       constant bus_handle : bus_master_t;
                       constant address : std_logic_vector;
--- a/vunit/vhdl/verification_components/src/bus_master_pkg.vhd
+++ b/vunit/vhdl/verification_components/src/bus_master_pkg.vhd
@@ -74,12 +74,12 @@ package bus_master_pkg is
   procedure write_bus(signal net : inout network_t;
                       constant bus_handle : bus_master_t;
                       constant address : std_logic_vector;
-                      constant burstsize : positive;
+                      constant burst_length : positive;
                       constant burstdata : queue_t);
   procedure write_bus(signal net : inout network_t;
                       constant bus_handle : bus_master_t;
                       constant address : natural;
-                      constant burstsize : positive;
+                      constant burst_length : positive;
                       constant burstdata : queue_t);
 
   -- Non blocking: Read the bus returning a reference to the future reply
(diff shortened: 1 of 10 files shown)


vunit/vhdl/verification_components/src/bus_master_pkg.vhd
```diff
@@ -70,6 +71,16 @@ package bus_master_pkg is
                       constant data : std_logic_vector;
                       -- default byte enable is all bytes
                       constant byte_enable : std_logic_vector := "");
+  procedure write_bus(signal net : inout network_t;
+                      constant bus_handle : bus_master_t;
+                      constant address : std_logic_vector;
+                      constant burstsize : positive;
```
kraigher: At least in AXI terminology it would be called burst length and not burst size.

slaweksiluk: In Avalon it is burstcount. The another option is simple "burst". However, for me, "burstlength" sounds good.

vunit/vhdl/verification_components/src/bus_master_pkg.vhd
```diff
@@ -70,6 +71,16 @@ package bus_master_pkg is
...
+                      constant burstdata : queue_t);
```
kraigher: This comment is valid for all functions added. The ownership of the queues should be documented in the API. For example is ownership of the queue transferred into the function?, does the queue need to be deallocated afterwards?, is the queue consumed by the function?
This is important for the user to know.

vunit/vhdl/verification_components/src/bus_master_pkg.vhd
LarsAsplund: I think we should either keep Avalon naming or use our current naming conventions all the way with underscore separators. burstlength would be something in between.

LarsAsplund: Queues are very explicit about data order but not explicit about what data types that it accepts. I'm on my mobile and haven't looked at the code carefully  but I think an integer queue is assumed. A user may find it more natural to represent the data as a vector. Or create the burst out of order. Maybe we should list some pros cons when comparing with integer_array_t.

slaweksiluk: So lets stay at 'burst_length'?

vunit/vhdl/verification_components/src/bus_master_pkg.vhd
slaweksiluk: Yes, it need to be documented. I assumed that queue is allocated in tb main which is responsible for its deallocation. In that case single queue could be shared by multiple reads/writes in different verification components instantiations.

vunit/vhdl/verification_components/src/bus_master_pkg.vhd
kraigher: I think the queue is fine for now. There is nothing stopping other data-types from being supported in the future if the message format consists of the data by value.

slaweksiluk: Ok so it seems no more activity here. I will change to *burst_length*.

vunit/vhdl/verification_components/src/bus_master_pkg.vhd
slaweksiluk: Is it enough to document queue API in vhdl package code or also in general vunit doc?

vunit/vhdl/verification_components/src/bus_master_pkg.vhd
kraigher: Ok burst_length is a good name

vunit/vhdl/verification_components/src/bus_master_pkg.vhd
kraigher: It is enough to document it in a comment above the function in the package header. 

Note: For many VUnit packages we have separated the package header from the package body to include the package header in verbatim on the webpage as a simple form of API documentation lacking any better autodoc for VHDL like there is for Python.

slaweksiluk:
Failed Travis build seems not related to this pull request

slaweksiluk:
@kraigher do you have idea what happened to travis now?

kraigher:
@slaweksiluk Looks like an environment error. I restarted the build manually.

slaweksiluk:
@kraigher I have not idea how to fix it
```
vunit/test/acceptance/test_external_run_scripts.py::TestExternalRunScripts::test_verification_components_vhdl_2008 

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

The build has been terminated

```

kraigher:
The PR needs to be rebased and the CI checks must pass. Then it is probably ready to merge after a review.

slaweksiluk:
Still time-out on travis

kraigher:
Are the new test you added long running tests? Travis has a timeout for builds. Maybe we need to start splitting the running of all run.py files into several builds.

kraigher:
@slaweksiluk I just pushed a commit to master which moves the verification components to a separate build. Hopefully it will make it more robust.

kraigher:
I think this is ready to merge except for the names I commented on, what do you think about the names @slaweksiluk. Is it not more clear to have different names for burst read and normal read or do you have a different opinion?

slaweksiluk:
I had doubts about it and decided to use the same names as in non-burst transfers. It is impossible to confuse burst and non burst transfers as they have different interfaces. On the other side having the same names has no clear advantages. Its up to you, I can implement *burst_read/write_bus* naming if needed.

kraigher:
I prefer having different names to make the behavioral difference more obvious when reading the code.

slaweksiluk:
@kraigher I've updated names

PR #371, opened 2018-08-13
Handle sync messages in axi_stream_slave

hugh-jackson-itdev:
Most of these changes are similar to current features in the axi_stream_master.

Added checks that wait_for_time behaves correctly for both master and slave (wasn't currently checked for master).


kraigher:
It seems a change in the Travis environment has made master branch CI fail and as a consequence this pull request. I think there is nothing wrong with the pull request itself so you can just rebase on master after it has been fixed.

PR #374, opened 2018-08-16
Add Riviera's global libs as builtin libraries

pkorpine:


vunit/rivierapro_interface.py
```diff
@@ -106,8 +106,13 @@ def __init__(self, prefix, output_path, persistent=False, gui=False, coverage=No
 
     def add_simulator_specific(self, project):
         """
-        Add coverage flags
+        Add builtin (global) libraries and coverage flags
         """
+        mapped_libraries = self._get_mapped_libraries()
+        for library_name in mapped_libraries:
+            if not project.has_library(library_name):
```
kraigher: A spot a potential problem here. 

1. Consider if the user adds a library `lib2` to their `run.py` file and runs once. The library will be added to `library.cfg` and will not be considered a built-in library.

2. The user removes `lib2` from their `run.py` and runs again. `vlist` will read `library.cfg` and discover `lib2` which will be added as a built-in library. This means any references to `lib2` will be ignored by VUnit instead of flagged as errors which it should since the library is not there anymore.

The possible solution would maybe be a bit out of scope of this PR. But I think that the `RiviveraProInterface` should not depend on the state of the `library.cfg` file in the form of reading it but rather always overwrite it regardless of what was there before. The built-in libraries can be discovered by running `vlist` without any `library.cfg` which will just show the built-in libraries.

PR #383, opened 2018-08-24
adds avalon streaming packet signals

Aaron-Hartwig:
This commit adds Start of Packet (SOP) and End of Packet (EOP) signals to the Avalon streaming components and Avalon transaction type.

### DIFF
--- a/vunit/vhdl/verification_components/src/avalon_source.vhd
+++ b/vunit/vhdl/verification_components/src/avalon_source.vhd
@@ -44,12 +44,12 @@ begin
 
     handle_sync_message(net, msg_type, msg);
 
-    if msg_type = stream_push_msg or msg_type = avalon_stream_transaction_msg then
+    if msg_type = stream_push_msg or msg_type = push_avalon_stream_msg then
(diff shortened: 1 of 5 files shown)


vunit/vhdl/verification_components/test/tb_avalon_stream.vhd
```diff
@@ -89,6 +87,8 @@ begin
       clk   => clk,
       valid => valid,
       ready => ready,
+      sop   => sop,
```
kraigher: There is no test added in tb_avalon_stream.vhd which means sop, eop is untested in the actual VC.

vunit/vhdl/verification_components/src/avalon_source.vhd
```diff
@@ -50,11 +52,25 @@ begin
       if msg_type = avalon_stream_transaction_msg then
         pop_avalon_stream_transaction(msg, avalon_stream_transaction);
         data <= avalon_stream_transaction.data;
+        if avalon_stream_transaction.sop then
```
kraigher: This code is never tested. See other comment.

PR #315, opened 2018-03-07
Adding axi stream signals

jklapel:
This adds in the rest of the AXI Stream signals (TKEEP, TSTRB, TID, TDEST, and TUSER).  It expands on the existing AXI Stream VC, keeping support for the generic stream VC but also expanding the AXI Stream Master VCI and adding the AXI Stream Slave VCI.

PR #385, opened 2018-08-27
Feature/axis signals

### LINKED ISSUE #380: axis tuser
Hi

I tried to add TUSER to the axi stream VC. I'm currently stuck with the optional character of the signal.
How would you implement optional parameters in both stream_master_pkg and axi_stream_pkg?

The stream_master_pkg does not know the length of tuser, so having a push_stream function without it fails because it cannot push the correct vector to the message. Or should I modify the protocol if there is no tuser data (i.e. somethimes push 3 values, sometimes 2)?
So I just left stream_master_pkg as is (tuser is axi specific anyway) and made the axi_stream_master drive tuser=zero if the message was not axi specific.

However, on the slave side, this distinction does not work as there is no pop_axi_stream_msg. There is no client function to receive the tuser signal (no pop_axi_stream).
What would be the best way to implement this?

Thanks and best regards,
emanuel

eschmidscs:
This should close #380. It reuses #315.
Additionally, I added AXI4STREAM_ERRM_TVALID_RESET

### DIFF
--- a/docs/about.rst
+++ b/docs/about.rst
@@ -144,7 +144,8 @@ Founders
 
 Notable contributors
 ********************
-- `Colin Marquardt <https://github.com/cmarqu>`_ (Cadence Incisive support)
+- `Colin Marquardt <https://github.com/cmarqu>`_: Cadence Incisive support
+- `Sławomir Siluk <https://github.com/slaweksiluk>`_: Verification Components such as Avalon and Wishbone
 
 
 License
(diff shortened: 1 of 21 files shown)


eschmidscs:
Besides the syntax issue with GHDL, there is a failing test as well (I tested with questa only).
I have to fix this tomorrow.

kraigher:
Great, It will probably take me some time to review this fully.

eschmidscs:
Two questions from a newbie:
- Is that travis build fail again something from the tooling? The error is inside osvvm.
- Why am I not getting notifications for updates of my own pull request? I'm watching the whole vunit project and getting tons of e-mails, but if somebody comments here, I'm not getting anything.

PR #392, opened 2018-08-31
Make axi_stream_master drive X if invalid

eschmidscs:
This is a somewhat radical change: It makes the axi_stream_master drive X (or as configured by generic) on most signals (tuser is different by ARM spec).

Reasoning: We had a discussion recently among the HDL people in the company about this. It was triggered by a bug, where a streaming block was tested correctly - but it was actually broken. The cause was that the stream source did not drive 100% duty cycle but applied the output value already after the last value was consumed.
The block did not correctly delay the valid signal (not long enough), but due to the early data value, the output was already correct.
The same happens the other way round, what would apply for the axi_stream_master: If the data value is kept after the transaction, a TVALID that is delayed too much would not be detected.

The people here agreed that a test driver must be conservative and actually cancel the data signals on a deactivated valid.
I just implemented this now. Maybe you have a different opinion? Or is there a better solution to this topic?

### DIFF
--- a/vunit/vhdl/verification_components/src/axi_stream_master.vhd
+++ b/vunit/vhdl/verification_components/src/axi_stream_master.vhd
@@ -16,7 +16,11 @@ use work.sync_pkg.all;
 
 entity axi_stream_master is
   generic (
-    master : axi_stream_master_t);
+    master : axi_stream_master_t;
+    drive_invalid          : boolean   := true;
+    drive_invalid_val      : std_logic := 'X';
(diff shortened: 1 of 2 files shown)


slaweksiluk:
I think think this is good approach. Once I spent some additional time on debugging because of data sampled when valid low.

kraigher:
Yes I agree that all values that are undefined when tvalid is low should be set to 'x' or a random value.

PR #395, opened 2018-09-23
Always use the most up-to-date version of modelsim.ini

### LINKED ISSUE #388: Updated value of VUNIT_MODELSIM_INI has no effect unless you delete the existing modelsim.ini file
Creating or updating the `VUNIT_MODELSIM_INI` environment variable has no effect unless you first delete the existing `modelsim.ini` file in `./vunit_out/modelsim/`.

bradleyharden:
The local copy of modelsim.ini is now re-created on each invocation
of VUnit, to capture any potential updates to VUNIT_MODELSIM_INI.
The unit tests for ModelSimInterface have been updated to eliminate
the expectation that a local modelsim.ini file exits. Instead,
setUp includes the creation of a simulated modelsim environment.
Upon initialization, ModelSimInterface will pull the modelsim.ini
file from this environment, as it would in the real case.

Closes #388

### DIFF
--- a/vunit/modelsim_interface.py
+++ b/vunit/modelsim_interface.py
@@ -100,11 +100,8 @@ class ModelSimInterface(VsimSimulatorMixin, SimulatorInterface):  # pylint: disa
 
     def _create_modelsim_ini(self):
         """
-        Create the modelsim.ini file if it does not exist
+        Create the modelsim.ini file
         """
-        if file_exists(self._sim_cfg_file_name):
-            return
-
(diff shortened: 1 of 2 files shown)


bradleyharden:
Is the only issue here the line break after the binary operator? Everything else appears to the be the same, seemingly unrelated issues with Travis.

kraigher:
Yes there is a problem with GHDL now but the pylint problems are from your commit.

PR #396, opened 2018-09-23
Allow recovery on preprocessing failure

bradleyharden:
When attempting to add a file with syntax errors, VUnit will
report that an error has occurred and skip preprocessing for
the file. A unit test for this case has also been added.

### DIFF
--- a/vunit/test/unit/test_ui.py
+++ b/vunit/test/unit/test_ui.py
@@ -135,6 +135,36 @@ end architecture;
                 report='log("Here I am!"); -- VUnitfier preprocessor: Report turned off, keeping original code.')
             self.assertEqual(fread.read(), expectd)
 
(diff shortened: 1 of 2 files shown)


bradleyharden:
What's going on with this? It doesn't look like any of these errors have to do with me. Is there an issue with the Travis environment?

kraigher:
@bradleyharden a recent update to GHDL has started to make VUnit regression tests fail: https://travis-ci.org/VUnit/vunit/jobs/433187950. There is no problem with your commit. We have to get GHDL fixed or work-arround the bug then this PR has to be rebased.

bradleyharden:
@kraigher Any idea when that will be?

kraigher:
The ghdl project is usually quite fast so any day now. On the other hand I could just use an older ghdl for Travis temporarily.

PR #407, opened 2018-11-09
Fix issue with missing coverage information in Riviera

soilheart:
When having multiple top-levels specified to RivieraPRO vsim (e.g. if one need to include glbl when simulating Xilinx IP) no coverage is collected for the testbench and it sub-hiearchies unless the testbench is the last top-level specified in the command.

Discussion with Aldec through FirstEDA is ongoing, but this is their currently suggested solution, so I thought I would open a pull request to see if this is an acceptable solution for the project, and if there are any further suggestions.

### DIFF
--- a/vunit/rivierapro_interface.py
+++ b/vunit/rivierapro_interface.py
@@ -231,13 +231,7 @@ class RivieraProInterface(VsimSimulatorMixin, SimulatorInterface):
 
         vsim_flags = ["-dataset {%s}" % fix_path(join(output_path, "dataset.asdb")),
                       pli_str,
-                      set_generic_str,
-                      "-lib",
-                      config.library_name,
-                      config.entity_name]
-
-        if config.architecture_name is not None:
-            vsim_flags.append(config.architecture_name)
+                      set_generic_str]
 
         if config.sim_options.get("enable_coverage", False):
             coverage_file_path = join(output_path, "coverage.acdb")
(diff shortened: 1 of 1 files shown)


vunit/rivierapro_interface.py
```diff
@@ -231,13 +231,7 @@ def _create_load_function(self,
...
         vsim_flags = ["-dataset {%s}" % fix_path(join(output_path, "dataset.asdb")),
                       pli_str,
-                      set_generic_str,
-                      "-lib",
-                      config.library_name,
-                      config.entity_name]
-
-        if config.architecture_name is not None:
-            vsim_flags.append(config.architecture_name)
+                      set_generic_str]
```
kraigher: What happens when the set generic string is separated from the top level? Have you tried this on a test bench which have generics set from Python?

kraigher:
Looks ok I just have one worry/comment which you might be able to answer.

vunit/rivierapro_interface.py
soilheart: I tried shuffling the arguments around with rand.shuffle() (as a part of diagnosing the coverage issues) and never noticed any issues with the re-ordering (except getting coverage data or not depending on the order of the testbench and glbl).

If you think it helps readability I can move the set_generic_str command down too.

soilheart:
I didn't get the acceptance tests running on my computer, and I noted that travis doesn't run the riviera acceptance tests. Do you think those could be affected by this change?
If so any idea what I may be missing to run the tests? (It doesn't seem to find my Riviera installation).

kraigher:
Set the VUNIT_RIVIERAPRO_PATH environment variable. But you can also just try to run any test bench that sets generics.

soilheart:
As I wrote above ordering doesn't seem to matter for the generics (ran 50 testbenches with random ordering of arguments, and everything seems to work).

PR #413, opened 2018-11-28
Fix bug with coverage flag not working with unique-sim in rivierapro

soilheart:


kraigher:
The pylint is broken on master also due to new pylint version. This bug would never have been there if VUnit was written in Rust.

PR #414, opened 2018-11-30
Fix issue with rivierapro not waiting for license before running sim

soilheart:
This enables riviera to wait for simulation licenses (such as RIVIERA_VHDL_SIMULATION_LV and RIVIERA_CODECOVERAGE_SUPPORT) between simulations by making the LICENSE_QUEUE variable available to the vunit_load proc.

I've tested and there's no change in behavior if the (tcl variable) LICENSE_QUEUE is not set.

Worth noting is that the LICENSE_QUEUE variable specifies the seconds to wait before asking for a new license, and nothing else, so LICENCE_QUEUE should preferably be set to something like 30 seconds.


joshrsmith:
As another Riviera user, thank you for this. I look forward to trying it!

PR #418, opened 2018-12-08
ghdl_interface: get ghdl executable name from environment variable GHDL.

tgingold:
This is for testing with various ghdl back-ends when the executable is not
the canonic one.

### DIFF
--- a/vunit/ghdl_interface.py
+++ b/vunit/ghdl_interface.py
@@ -28,6 +28,7 @@ class GHDLInterface(SimulatorInterface):
     """
 
     name = "ghdl"
+    executable = os.environ.get("GHDL", "ghdl")
     supports_gui_flag = True
     supports_colors_in_gui = True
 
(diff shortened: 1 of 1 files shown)


kraigher:
Isn't this redundant with the `VUNIT_GHDL_PATH` environment variable?
It can be used like this:
```bash
export VUNIT_GHDL_PATH=/home/kraigher/repo/install-ghdl-mcode/bin
```

tgingold:
VUNIT_GHDL_PATH assumes that the executable is named ghdl.
So it cannot be used when you want to test /usr/local/bin/ghdl_mcode and /usr/local/bin/ghdl_llvm
(which is my particular case).

kraigher:
We always prefix environment variables that affect VUnit with `VUNIT_` to avoid unintentional interaction with whatever environment variables the user might have set. Soo this variable should have VUNIT_ prefix.

kraigher:
Unless the GHDL environment variable is a standard withing GHDL always having this meaning. Then it does not need a prefix.

tgingold:
I see your point, but I think that the use of the GHDL environment variable is quite standard (at least in Makefiles).

PR #420, opened 2018-12-10
Fixed clock alignment in axi_stream_master vc

olafvandenberg:
Split axi stream master process into two separate processes, as was done in axi_lite_master before, to keep alignment with the clock edge and make it possible to implement a wait_until_idle. After handling a wait_for_time message, there is a realignment with the clock as well. To still be able to run at full speed, the queue is processed until it is empty before waiting for another rising edge.

Added an extra test to the testbench to verify clock alignment.
Slightly modified other tests according to changed timing behavior.
Added 'wait until idle' to pop before push test to verify it's behavior.

### DIFF
--- a/vunit/vhdl/verification_components/src/axi_stream_master.vhd
+++ b/vunit/vhdl/verification_components/src/axi_stream_master.vhd
@@ -74,7 +74,7 @@ begin
     wait until rising_edge(aclk) and not is_empty(message_queue);
 
     while not is_empty(message_queue) loop
-      
+
       msg := pop(message_queue);
       msg_type := message_type(msg);
 
--- a/vunit/vhdl/verification_components/src/axi_stream_master.vhd
+++ b/vunit/vhdl/verification_components/src/axi_stream_master.vhd
@@ -111,6 +111,7 @@ begin
         unexpected_msg_type(msg_type);
       end if;
 
+      delete(msg);
     end loop;
 
   end process;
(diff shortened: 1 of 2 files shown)


vunit/vhdl/verification_components/src/axi_stream_master.vhd
```diff
@@ -50,39 +70,49 @@ begin
...
-        tlast <= pop_std_ulogic(msg);
-        tkeep <= pop_std_ulogic_vector(msg);
-        tstrb <= pop_std_ulogic_vector(msg);
-        tid <= pop_std_ulogic_vector(msg);
-        tdest <= pop_std_ulogic_vector(msg);
-        tuser <= pop_std_ulogic_vector(msg);
-      else
-        if pop_boolean(msg) then
-          tlast <= '1';
+      msg := pop(message_queue);
```
kraigher: The msg needs to be deallocated unless memory will leak. Previously the msg was deallocated by being passed into the receive procedure in the main process. Now the msg is nulled in the main process when being pushed into the queue since ownership is transfered and thus it must be explicitly deallocated in the bus_process.

olafvandenberg: agree, will change that.. Technically it also needs to be done after the unexpected_msg_type call in the main process I guess?

kraigher: Yes it could have been done explicitly in the main process as well but the receive procedure will deallocate its input argument if it is non-null so there was no memory leak in the original code. Maybe it is confusing and could be more readable with an explicit deallocate.

PR #422, opened 2018-12-12
reset for axi_stream_master

olafvandenberg:
added reset output to use when the axi_stream_master is generating the reset.
changed wait until idle construction to use a notify from the bus process so that we are sure all transactions (including the reset) have been handled.
added checking of reset input on places where it is required so that a reset is handled properly in the master.
added axi_stream_reset to vci.
added reset input to axi_stream_slave.vhd.
added basic test for reset functionality.

### DIFF
--- a/vunit/vhdl/verification_components/src/axi_stream_master.vhd
+++ b/vunit/vhdl/verification_components/src/axi_stream_master.vhd
@@ -23,8 +23,7 @@ entity axi_stream_master is
   );
   port (
     aclk         : in  std_logic;
-    areset_in_n  : in  std_logic                                          := '1';
-    areset_out_n : out std_logic                                          := '1';
+    areset_n     : in  std_logic                                          := '1';
     tvalid       : out std_logic                                          := '0';
     tready       : in  std_logic                                          := '1';
     tdata        : out std_logic_vector(data_length(master)-1 downto 0)   := (others => '0');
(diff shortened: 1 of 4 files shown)


vunit/vhdl/verification_components/src/axi_stream_master.vhd
```diff
@@ -71,49 +80,64 @@ begin
...
-      end if;
 
-      delete(msg);
-    end loop;
+        delete(msg);
+      end loop;
 
+      notify_bus_process_done <= '1';
+      wait until notify_bus_process_done = '1';
+      notify_bus_process_done <= '0';
```
go2sh: Also, I'am not sure, why this is needed. waiting for tvalid = '0' and aclk should be sufficient.

kraigher:
@go2sh Has a point. It would be more modular if there was a separate reset generator VC and the axi stream just had reset as an input? What do you say @olafvandenberg?

vunit/vhdl/verification_components/src/axi_stream_master.vhd
olafvandenberg: this was necessary to be able to issue a reset 'request' and follow it up with a wait_until_idle to wait for it to complete.
When just adding a areset_out_n = '1' there was a timing hole and the wait_until_idle would complete to early.

olafvandenberg:
Ok, I agree, sound likes a plan, I'll need some time for it though

kraigher:
Yes a to drive a reset is nothing special so an general io vc is probably smart.

PR #429, opened 2018-12-18
Feature/clk align axi stream slave

olafvandenberg:


eschmidscs:
We've hit something similar some time ago.
You probably use the pop function or the check function with blocking=>True.
In this case, the slave takes one message, handles it and goes back to idle. Going back to idle causes ready to go low for one cycle.
You either have to use pop with references or check with blocking=>False.

That was different, probably before the merge you mentioned. But the process structure in the slave was changed.

umarcor:
@eschmidscs did you want to post that comment here or in #573?

eschmidscs:
good one... thanks and sorry!

PR #439, opened 2019-01-22
Add support for configuration attributes

barri:
Adds the ability to attach an attribute dictionary to a test or
testbench configuration.
These attributes are exported to json when --export-json is supplied to
the python runner.

### DIFF
--- a/vunit/test/unit/test_configuration.py
+++ b/vunit/test/unit/test_configuration.py
@@ -14,7 +14,8 @@ Tests the test test_bench module
 import unittest
 import contextlib
 from os.path import join
-from vunit.configuration import Configuration
+from vunit.configuration import (Configuration,
+                                 AttributeException)
 from vunit.test.mock_2or3 import mock
 from vunit.test.common import (with_tempdir,
                                create_tempdir)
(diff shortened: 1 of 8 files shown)


LarsAsplund:
To be consistent with how configurations works for other items, like generics, the attribute defined for a test case as a comment in the code should be considered an attribute for the default and unnamed configuration of that test case. Just like a default value of a generic is the value used for the default configuration.

Then there should be a `set_attributes` method allowing you to set the attributes such that the default attribute in the comment is replaced. Just like `set_generic` replaces the default value.

Whenever you do `add_config` to add a named configuration the default configuration is ignored. This means that the attribute in the comment is ignored and is not exported to JSON. This is not how it works now, the JSON export contains attributes in the code as well as those defined by the `add_config` method.

LarsAsplund:
You're right, I tested your pull-request again and it works as intended. My mistake.

vunit/ui.py
```diff
@@ -1503,6 +1547,7 @@ def add_config(self,  # pylint: disable=too-many-arguments
         :param pre_config: A :ref:`callback function <pre_and_post_hooks>` to be called before test execution.
         :param post_check: A :ref:`callback function <pre_and_post_hooks>` to be called after test execution.
         :param sim_options: A `dict` containing the sim_options to be set in addition to the default configuration
+        :param sim_options: A `dict` containing the attributes to be set in addition to the default configuration
```
LarsAsplund: sim_options -> attributes

vunit/test/unit/test_configuration.py
```diff
@@ -67,6 +67,14 @@ def test_adds_tb_path_generic(self, tempdir):
         config_tb_path = Configuration('name', design_unit_tb_path)
         self.assertEqual(config_tb_path.generics["tb_path"], (tb_path + "/").replace("\\", "/"))
 
+    def test_constructor_adds_no_attributes(self):
+        with _create_config() as config:
+            self.assertEqual({}, config.attributes)
+
+    def test_constructor_adds_supplied_attributes(self):
```
LarsAsplund: Adding an attribute named `foo` should be illegal and generate an error. `.foo` is ok. Add check and test for this.

PR #441, opened 2019-01-26
Add additional basic check macros for SystemVerilog VUnit

dbalthazor:
Added macros:

- `CHECK_NOT_EQUAL(A, B): true is A !== B`

- `CHECK_LESS(A, B): true if A is strictly less than B`

- `CHECK_GREATER(A, B): true if A is strictly greater than B`

- `CHECK_EQUAL_VARIANCE(A, B, VAR): true if A is between B+VAR and B-VAR`

I also added a barebones testbench and testbench runner for SV check macros, under `verilog/check/test`. At the moment, all it really checks for is if the macros are currently visible.

### DIFF
--- a/vunit/verilog/check/test/check_tb.sv
+++ b/vunit/verilog/check/test/check_tb.sv
@@ -10,32 +10,53 @@
 module check_tb;
 
 	class test_data;
-		randc	logic	[15:0] 	data_logic;
-		randc	integer			data_integer;
-		randc	byte			data_byte;
-		randc	shortint		data_shortint;
-		randc	longint			data_longint;
-		randc	time			data_time;
-		//randc	real			data_real;
-	endclass;
+		bit				[15:0] 	data_bit;
+		int				data_int;
+		byte			data_byte;
+		shortint		data_shortint;
+		time			data_time;
+		function void make_random;
+			data_bit = $random();
+			data_int = int'($random());
+			data_byte = byte'($random());
+			data_shortint = shortint'($random());
+			data_time = time'($random());
+		endfunction
+	endclass
 	test_data case_data1;
 	integer case_greater, case_less;
 
 	`TEST_SUITE begin
 		`TEST_SUITE_SETUP begin
 			case_data1 = new();
+			case_data1.make_random();
+			case_greater = case_data1.data_int+1;
+			case_less = case_data1.data_int-1;
 		end
 		`TEST_CASE("Check Macros Are Visible") begin
-				if(!case_data1.randomize())
-					$display("Randomization failed");
-				case_greater = case_data1.data_integer+1;
-				case_less = case_data1.data_integer-1;
-				`CHECK_EQUAL(case_data1.data_integer, case_data1.data_integer);
-				`CHECK_NOT_EQUAL(case_data1.data_integer, case_greater);
-				`CHECK_GREATER(case_greater, case_data1.data_integer);
-				`CHECK_LESS(case_less, case_data1.data_integer);
-				`CHECK_EQUAL_VARIANCE(case_less, case_data1.data_integer, 2);
-				`CHECK_EQUAL_VARIANCE(case_greater, case_data1.data_integer, 2);
+			// if(!case_data1.randomize())
+			//	$error("Randomization failed");
+			`CHECK_EQUAL(case_data1.data_int, case_data1.data_int);
+			`CHECK_NOT_EQUAL(case_data1.data_int, case_greater);
+			`CHECK_GREATER(case_greater, case_data1.data_int);
+			`CHECK_LESS(case_less, case_data1.data_int);
+			`CHECK_EQUAL_VARIANCE(case_less, case_data1.data_int, 2);
+			`CHECK_EQUAL_VARIANCE(case_greater, case_data1.data_int, 2);
 		end
+		/* `TEST_CASE("CHECK_EQUAL failure message") begin
+			`CHECK_EQUAL(case_data1.data_int, case_greater, "This test should fail.");
+		end
+		`TEST_CASE("CHECK_NOT_EQUAL failure message") begin
+			`CHECK_NOT_EQUAL(case_data1.data_int, case_data1.data_int, "This test should fail.");
+		end
+		`TEST_CASE("CHECK_GREATER failure message") begin
+			`CHECK_GREATER(case_data1.data_int, case_data1.data_int, "This test should fail.");
+		end
+		`TEST_CASE("CHECK_LESS failure message") begin
+			`CHECK_LESS(case_data1.data_int, case_data1.data_int, "This test should fail.");
+		end
+		`TEST_CASE("CHECK_EQUAL_VARIANCE failure message") begin
+			`CHECK_EQUAL_VARIANCE(case_data1.data_int, case_data1.data_int+5, 1, "This test should fail.");
+		end */
 	end
 endmodule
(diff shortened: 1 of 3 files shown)


vunit/verilog/check/test/check_tb.sv
```diff
@@ -0,0 +1,41 @@
...
+//
+// Copyright (c) 2014-2018, Lars Asplund lars.anders.asplund@gmail.com
+
+`timescale 10ns / 10ns
+`include "vunit_defines.svh"
+
+module check_tb;
+
+	class test_data;
+		randc	logic	[15:0] 	data_logic;
```
svenka3: logic --> bit

SV randomization is based on 2-state

vunit/verilog/check/test/check_tb.sv
```diff
@@ -0,0 +1,41 @@
...
+		randc	integer			data_integer;
```
svenka3: Consider int (vs integer - 4 state vs 2 state again)

vunit/verilog/check/test/check_tb.sv
```diff
@@ -0,0 +1,41 @@
...
+		randc	byte			data_byte;
+		randc	shortint		data_shortint;
+		randc	longint			data_longint;
+		randc	time			data_time;
+		//randc	real			data_real;
+	endclass;
```
svenka3: endclass : test_data 

(i.e. Use end-label, no ; after endclass please

vunit/verilog/include/vunit_defines.svh
```diff
@@ -53,3 +53,120 @@
              full_msg = {"CHECK_EQUAL failed! Got ",`"got`", "=",  got_str, " expected ", expected_str, ". ", msg}; \
              $error(full_msg); \
           end
+`define CHECK_NOT_EQUAL(got,expected,msg=__none__) \
```
LarsAsplund: Much of the code in the macros are shared. I suggest that you refactor to remove code duplication

vunit/verilog/check/test/check_tb.sv
LarsAsplund: Right now none of the main developers have licenses to to handle SV randomization. We use the free Intel version of ModelSim. Aldec provide us with a license for Riviera-PRO but it doesn't support "advanced" SV verification features. I can try to see if I can have that upgraded but it may take a while. Meanwhile we need a simpler testbench in order to run a full acceptance test suite.

dbalthazor: Any suggestions on how to handle randomization? My first thought is to bring over the random package from OSVVM, but I'm not sure if your license supports mixed HDL.

vunit/verilog/check/test/check_tb.sv
dbalthazor: thanks for the feedback, I think I addressed most of your suggestions in the latest commits.

vunit/verilog/check/test/check_tb.sv
```diff
@@ -0,0 +1,62 @@
...
+			// if(!case_data1.randomize())
+			//	$error("Randomization failed");
+			`CHECK_EQUAL(case_data1.data_int, case_data1.data_int);
+			`CHECK_NOT_EQUAL(case_data1.data_int, case_greater);
+			`CHECK_GREATER(case_greater, case_data1.data_int);
+			`CHECK_LESS(case_less, case_data1.data_int);
+			`CHECK_EQUAL_VARIANCE(case_less, case_data1.data_int, 2);
+			`CHECK_EQUAL_VARIANCE(case_greater, case_data1.data_int, 2);
+		end
+		/* `TEST_CASE("CHECK_EQUAL failure message") begin
```
LarsAsplund: SystemVerilog is not my language but I think you should be able to make a macro of the `$error` call in the check macros. Then you can redefine that macro in the testbench to write the error message to a string variable and then verify its contents (fail using `$error`). Empty on a passing check and a message on a failing check

vunit/verilog/check/test/check_tb.sv
LarsAsplund: Now it works with my ModelSim

vunit/verilog/check/test/check_tb.sv
dbalthazor: Great idea. I put the following definition in vunit_defines.svh
` define __ERROR_FUNC(msg) $error(msg)  `

Redefining it in check_tb.sv to:
` define __ERROR_FUNC(msg) $sformat(test_output, "%s", msg) `

prints the error output to test_output, which is a string I declared earlier.

It does give a warning about redefining the macro, and it will also redefine it for any file compiled after check_tb.sv. I think this will work well enough for at least an acceptance test of these macros. I'm in the process of creating a basic check_pkg and checker_pkg in SV that will implement these as functions rather than macros, which should make this a little easier to work with.

This would be so much easier if SV had something like try-catch in it.

LarsAsplund: Yes, it's strange that SV added so much SW to Verilog but omitted exception handling

vunit/verilog/include/vunit_defines.svh
```diff
@@ -26,30 +26,40 @@
 
 `define TEST_CASE_SETUP if (__runner__.is_test_case_setup())
 `define TEST_CASE_CLEANUP if (__runner__.is_test_case_cleanup())
+`define __ERROR_FUNC(msg) $error(msg)
+`define CREATE_MSG(full_msg,func_name,got,expected,msg=__none__) \
+	string __none__; \
+	string full_msg; \
+	$sformat(full_msg, "%s failed! Got %d expected %d. %s", func_name, got, expected, msg);
```
LarsAsplund: I think you should keep the original style when `CHECK_EQUAL(my_data, 1)` yields an error message showing that it was `my_data` that failed. It gives the user some extra context information to distinguish different checks from each other even if no message is supplied. In general you can have an output like this:

...Got \<first argument>=\<the erroneous value> expected \<prefix>\<expected value>....

where \<prefix> is "" for `CHECK_EQUAL` and `CHECK_EQUAL_VARIANCE`.  "!=", ">", "<" for `CHECK_NOT_EQUAL`, `CHECK_GREATER` and `CHECK_LESS` respectively.

vunit/verilog/include/vunit_defines.svh
```diff
@@ -26,30 +26,65 @@
...
+         string variance_str; \
+         string full_msg; \
+         int index; \
+         got_str = "";\
+         expected_str ="";\
+         variance_str="";\
+         `CREATE_ARG_STRING(got, got_str); \
+         `CREATE_ARG_STRING(expected, expected_str); \
+         `CREATE_ARG_STRING(variance, variance_str); \
+             full_msg = {"CHECK_EQUAL_VARIANCE failed! Got ",`"got`", "=",  got_str, " expected =", expected_str, ", +-", variance_str, ". ", msg}; \
```
LarsAsplund: " expected =" -> " expected " and ", +-" -> "+-" looks better

vunit/verilog/include/vunit_defines.svh
```diff
@@ -26,30 +26,65 @@
...
-               end \
-               for (int i=0; i<expected_str.len(); i++) begin \
-                  if (expected_str[i] != " ") begin \
-                     expected_str = expected_str.substr(i, expected_str.len()-1); \
-                     break; \
-                  end \
-               end \
-             full_msg = {"CHECK_EQUAL failed! Got ",`"got`", "=",  got_str, " expected ", expected_str, ". ", msg}; \
-             $error(full_msg); \
+          `CREATE_MSG(full_msg, "CHECK_EQUAL", got, expected, "=", msg); \
```
LarsAsplund: I think "=" should be "". Reads well without it.

LarsAsplund:
Thanks for your contribution @dbalthazor!

PR #453, opened 2019-03-06
Add option to not dependency scan xil_defaultlib

LukasVik:
In my current project we have some IP cores that add all their (encrypted) files to xil_defaultlib. Unlike cores from Xilinx that add their unencrypted top levels to xil_defaultlib and implementations to specific libraries e.g. `lib_fifo_v1_0_10`, which makes it possible to do some rudimentary dependency scan.

Without this PR, the files in xil_defaultlib are not compiled in the correct order in my case, which breaks the compilation. With this change I can disable the dependency scan and instead use a linear dependency chain for all of the IP files, which makes them compile in the correct order.

### DIFF
--- a/vunit/vivado/vivado.py
+++ b/vunit/vivado/vivado.py
@@ -86,7 +86,7 @@ def _read_compile_order(file_name):
     with open(file_name, "r") as ifile:
 
         for line in ifile.readlines():
-            library_name, file_type, file_name = line.strip().split(",", maxsplit=2)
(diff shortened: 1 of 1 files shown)


LukasVik:
Also a small fix to make `vivado.py` work in python 2.7

```
...
  File "<snip>\vunit\vivado\vivado.py", line 89, in _read_compile_order
    library_name, file_type, file_name = line.strip().split(",", maxsplit=2)
TypeError: split() takes no keyword arguments
```

Issue #458, opened 2019-03-15
Feature Request: Support set_sim_option name 'gtkwave.init_file.gui' in GTKWave similar to ModelSim, Active-HDL, and Riviera PRO


GlenNicholls:
ModelSim, Active-HDL, and Riviera PRO all have the `'<simulator_name>.init_file.gui'` name option for `set_sim_option()`. However, the only way to accomplish this with GTKWave is to do `python run.py -g --gtkwave-args=-S<script_name>.tcl` and this is somewhat annoying to do everytime I want to run the same script/scripts everytime I open the GUI. I am looking to configure the waveform viewer with some of my scripts and this gets very annoying after doing it a few times. Can support be added to define the scripts to be run from run.py similar to the other supported simulators?

kraigher:
Yes it makes sense to add it. If you want to make a PR just have a look in ghdl_interface.py and you can se how other sim options are handled there. It should be very straight forward to add this.

GlenNicholls:
Okay cool, I will dig into the code now and make a PR once I have some time to make and test the changes.

GlenNicholls:
@kraigher I have all of the code written and I'm working on testing it now. With this, it looks like it works when there are no spaces in the file path. However, I found that if there are any spaces in the path name, then GTK vomits with this error:

When the VUnit run.py has no spaces in the path name, i.e. `ui.set_sim_option('ghdl.init_file.gui', 'C:/tmp/gtkAddWave.tcl')`, everything looks fine. However, if I try to use the script in the run directory that has a space in the path name, it all blows up (`ui.set_sim_option('ghdl.init_file.gui', 'gtkAddWave.tcl')`):


```
$ python run.py "axi_dma_lib.tb_axi_dma_regs.Test destination address" -g -v
Re-compile not needed

Running test: axi_dma_lib.tb_axi_dma_regs.Test destination address
Running 1 tests

Starting axi_dma_lib.tb_axi_dma_regs.Test destination address
Output file: vunit_out\test_output\axi_dma_lib.tb_axi_dma_regs.Test_destinadb89a2c4a29e6dfdee0f89140e6749550048c785\output.txt
simulation stopped @105ns with status 0
init file: gtkAddWave.tcl
[...]

Issue #455, opened 2019-03-08
Feature Request: Add arg parse options for adding waveforms to viewer


GlenNicholls:
I think it would be a nice feature to allow the user to use a flag to display all signals at a user-specified hierarchy depth when they pass the `-g` flag to open the GUI. I get really annoyed having to manually add waveforms when I don't yet know what I'm looking for, so I end up adding all of them. GTKWave offers groups that can be named, ModelSim offers groups as well that can be named, and Active-HDL/Riviera-Pro offer virtual signal busses (I believe with `add -virtual "name_for_virtual" path/to/desired/entity/*`).

Here is rough-draft of what i am referring to for GTK:

```
# note hierarchy depth to display in waveform viewer
set depth 2

set top_level top; # change top only if applicable

# don't change this
set num_facs  [ gtkwave::getNumFacs ]
set dump_name [ gtkwave::getDumpFileName ]
set dmt       [ gtkwave::getDumpType ]

# create list with all elements
set elements [list]
for {set i 0} {$i < $num_facs} {incr i} {
    lappend elements [ gtkwave::getFacName $i ]
}

# Now display hierarchy depth in waveform viewer
set wave [list]
foreach element $elements {
  # initial length of element
  set i_len [string length $element]

  # final length of element after removing dots
  set f_len [string length [string map {. ""} $element]]

  # thus number of dots
  set n_dots [expr {$i_len-$f_len}]

  # if the number equals the required hierarchy, then we got one
  if {$n_dots == $depth} {
    lappend wave $element
  }
}

gtkwave::/Edit/Insert_Blank
# append all to waveform viewer
set num_added [ gtkwave::addSignalsFromList $wave ]
puts "$wave"
```

This will go through and add all waveforms to the viewer that are at `depth`, so I am working on figuring out how to name groups, `gtkwave:Edit/Create_Group`, from the command line for the following. It would be nice to tell VUnit that I want to see all signals at and before hierarchy depth of 3 and it would do the following in the waveform viewer:

```
tb_name_here { <- root depth or all tb signals
    clk
    data
    valid
}

tb_name_here.uut1 { <- UUT inside tb
    clk
    data
    valid
    nextState
}

tb_name_here.uut1.someEntity { <- some entity inside uut1
    clk
    data
    valid
    someSignalHere
    ...
}
```

[...]

PR #459, opened 2019-03-21
Updated ghdl_interface.py per #458

GlenNicholls:
This update allows a TCL script to be passed to GTKWave when the `-g` flag is passed if the environment has the `set_sim_option('ghdl.init_file.gui', '<fileName>.tcl')` set. The limitation is that only a single file can be passed, however `<fileName>.tcl` can source other TCL scripts if desired.

### DIFF
--- a/vunit/ui.py
+++ b/vunit/ui.py
@@ -191,6 +191,16 @@ The following simulation options are known.
    Extra simulation flags passed to ``ghdl --elab-run``.
    Must be a list of strings.
 
+``ghdl.gtkwave_script.gui``
+   A user defined TCL-file that is sourced after the design has been loaded in the GUI.
+   For example this can be used to configure the waveform viewer.
+   During script evaluation the ``vunit_tb_path`` variable is defined
+   as the path of the folder containing the test bench.
+   Must be a string. There are currently limitations in the HEAD revision of GTKWave
+   that prevent the user from sourcing a list of scripts directly. The following is the
+   current work around to sourcing multiple user TCL-files:
+   ``source <path/to/script.tcl>``
+
 .. |compile_option| replace::
    The name of the compile option (See :ref:`Compilation options <compile_options>`)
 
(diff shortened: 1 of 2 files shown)


GlenNicholls:
Another thing to note is that `ghdl.init_file.gui` might be better suited as `gtkwave.init_file.gui` to avoid confusion, but I left it using the interface name for simplicity.

kraigher:
As you mentioned it would be good if the name conveyed that gtkwave was involved. Maybe calling the sim_option ghdl.gtkwave_script would be better. 

Also the sim option should be mentioned in the documentation. The source for the documentation is within ui.py where there is a list of sim options in a docstring. Just adding the new option there with a small description is enough.

vunit/ui.py
```diff
@@ -191,6 +191,16 @@
    Extra simulation flags passed to ``ghdl --elab-run``.
    Must be a list of strings.
 
+``ghdl.gtkwave_script.gui``
+   A user defined TCL-file that is sourced after the design has been loaded in the GUI.
+   For example this can be used to configure the waveform viewer.
+   During script evaluation the ``vunit_tb_path`` variable is defined
```
kraigher: I do not see how vunit_tb_path tcl variable is available so this documentation is inaccurate. Maybe copy paste error?

GlenNicholls: Yep, sorry about that
```suggestion
```

umarcor:
I tried adding `vu.set_sim_option('ghdl.gtkwave_script.gui', 'addWave.tcl')` to my `run.py`, where `addWave.tcl` is https://github.com/VUnit/vunit/issues/455#issuecomment-470797811. It fails with:

```
GTKWave Analyzer v3.3.101 (w)1999-2019 BSI

[0] start time.
[1400000000] end time.
GTKWAVE | Executing Tcl script '"/src/addWave.tcl"'
GTKWAVE | couldn't read file ""/src/addWave.tcl"": no such file or directory
WM Destroy
```

However, if I replace https://github.com/VUnit/vunit/blob/master/vunit/ghdl_interface.py#L276 with `cmd += ["--script", "{}".format(abspath(init_file))]`, it works as expected.

@GlenNicholls, might those additional double quotes be platform-dependent?

Issue #460, opened 2019-03-21
Add code coverage support ActiveHDL


vmfox:
Following my recent chat on Gitter regarding cove coverage support for ActiveHDL I would like to contribute the changes that would allow this to happen. I have a working version of the relevant .py file which was tested on a real project, and I can prove that the modification generate coverage files for each unit and final merged coverage file. The run.py interface is identical to the already existing one and hence would not require any changes.

I am not able to branch off to add this feature so it would be great if someone could assist me with it.

kraigher:
The workflow of creating PRs is explained by github here: https://help.github.com/en/articles/creating-a-pull-request-from-a-fork

vmfox:
Tack så mycket.

kraigher:
I guess we can close this issue as the PR is merged.

PR #461, opened 2019-03-21
Added code coverage support to ActiveHDL

vmfox:
Follow up from #460.

### DIFF
--- a/vunit/activehdl_interface.py
+++ b/vunit/activehdl_interface.py
@@ -305,7 +305,6 @@ proc vunit_run {} {
         vcover_merge_process.consume_output()
         print("Done merging coverage files")
 
-
     def _create_common_script(self, config, output_path):
         """
         Create tcl script with functions common to interactive and batch modes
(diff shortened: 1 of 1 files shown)


vunit/activehdl_interface.py
```diff
@@ -26,7 +28,7 @@
 LOGGER = logging.getLogger(__name__)
 
 
-class ActiveHDLInterface(SimulatorInterface):
+class ActiveHDLInterface(VsimSimulatorMixin, SimulatorInterface):
```
kraigher: Why is it necessary to inherit from VsimSimulatorMixin? I think that could be a misunderstanding.

vmfox: It probably is a misunderstanding as it's a bit of a challenge to jump into such a big project. The assumption was that akin to Riviera Pro and Modelsim, ActiveHDL is also a vsim based simulator and hence would require this inheritance.  I have just tested the performance without this inheritance and everything seems to runs as it was before, so will update as a part of the next commit.

PR #469, opened 2019-04-08
fix: nonzero return values should produce a fail

umarcor:
Related to #465.

If a GHDL simulation is wrapped in a C application, when the simulation is successful but the application exists with a nonzero code, VUnit still reports the test as *passed*. This is because it ignores the return code and checks the output file instead.

This PR sets the status of a test to *fail* if the return code is nonzero.


kraigher:
VUnit ignores the return code to be able to end simulation with an assert of severity failure for non VHDL-2008 where std.end.stop exists. Thus failing all test when the exit code is non-zero will break VUnit-support for VHDL 93 and 2002.

umarcor:
What is your suggestion?

- Enable this fix only when VHDL 2008 is used.
- Find a different mechanism to tell VUnit when some wrapper of the sim does want to report a failure.

kraigher:
An acceptable way forward would be to check the simulator exit code in VHDL 2008. It would prevent people wrapping the simulator binary from using a standard less than 2008 but I guess that would be no problem.

umarcor:
> An acceptable way forward would be to check the simulator exit code in VHDL 2008.

I now added a check for the version.

> It would prevent people wrapping the simulator binary from using a standard less than 2008 but I guess that would be no problem.

If someone tries to wrap the binary with a standard less than 2008, the behaviour will be the same as if no fix existed. So, as far as I understand it, they can still do it, but they will have to handle false positives.

vunit/test_suites.py
```diff
@@ -171,6 +171,11 @@ def run(self, output_path, read_output):
 
         results = self._read_test_results(file_name=get_result_file_name(output_path))
 
+        if hasattr(self._simulator_if, 'get_vhdl_standard') and \
```
kraigher: I would prefer if hasattr was avoided. I also think the simulator_if should provide a method named `has_valid_exit_code` which is default False for all simulators but True for GHDL with vhdl_standard=2008 instead of checking the standard directly here. 

Also why "if name is PASSED" ?

umarcor: > I would prefer if hasattr was avoided. I also think the simulator_if should provide a method named `has_valid_exit_code` which is default False for all simulators but True for GHDL with vhdl_standard=2008 instead of checking the standard directly here.

Agree.

> Also why "if name is PASSED" ?

At first I just set all to FAILED directly. But that produced errors with the tests that expect to get SKIPPED. Now, only FAILED are rewritten and SKIPPED are kept. I believe that it is not a false positive to return SKIPPED, even if the binary failed.

umarcor:
@kraigher this is failing because `lib.tb_same_sim_some_fail.Test 1 got failed expected passed`. Should I change the expected value to `failed`?

kraigher:
The situation is that VUnit supports running all test cases within a test bench in the same simulation using the run_all_in_same_sim comment attribute. In these cases you might have three tests "Test 1", "Test 2" and "Test 3" that are run in sequence. If "Test 1" pass and "Test 2" fails the "Test 3" will be considered skipped as the simulation stops at the failure of "Test 2". The change will make "Test 1" fail also since the exit code will fail all tests that were run in the same simulation.

I think the acceptance criteria is correct and should not be changed. One way to solve it would be to only fail all tests if all test pass in the presence of non-zero exit code. Thus if any test fails we let the test results be the way they where even if the exit code was non-zero. This would behave the same way as the current solution for the default case where each test is run in its own simulation.

We have a lot of acceptance test in VUnit to ensure things we wanted to work a specific way does not unintentionally break in the future. In this case your

umarcor:
> I think the acceptance criteria is correct and should not be changed. One way to solve it would be to only fail all tests if all test pass in the presence of non-zero exit code. Thus if any test fails we let the test results be the way they where even if the exit code was non-zero. This would behave the same way as the current solution for the default case where each test is run in its own simulation.

I moved the fix after the check now.

I am not sure if the check should be kept as `status != PASSED` or changed to `status == FAILED`. ATM, if some pass and others are skipped, the check of nonzero exit code is not evaluated.

> We have a lot of acceptance test in VUnit to ensure things we wanted to work a specific way does not unintentionally break in the future. In this case your PR does not include any comment, unit test or acceptance test so there is a likelyhood that the behavior you want w.r.t the exit code will break in the future. Maybe a unit test with a comment is sufficient in this case.

I'm trying to add some unit test, but I am finding it difficult to implement without actually compiling a C file where GHDL is embedded. Any ideas?

kraigher:
The rule should be, if non zero exit code and not any status=FAILED and code we flip all PASSED to FAILED

Regarding unit test it should be a python test in test/unit/test_test_suites.py at the function level without involving any external tool. This is unlike an acceptable test where VUnit is tested end to end and the simulator is involved.

umarcor:
> The rule should be, if non zero exit code and not any status=FAILED and code we flip all PASSED to FAILED

Agree. Moreover, now all the tests must be PASSED to execute `call_post_check`. Would it be acceptable to allow PASSED and/or SKIPPED to be passed to `call_post_check` instead?

> Regarding unit test it should be a python test in test/unit/test_test_suites.py at the function level without involving any external tool. This is unlike an acceptable test where VUnit is tested end to end and the simulator is involved.

Isn't it possible to do something similar to the following?

``` vhdl
      end if;
    end loop;
    --p_disable_simulation_exit(runner);
    test_runner_cleanup(runner);
    assert false severity error;
    wait;
  end process;
end architecture;
```

To do so, I need some 'conversion' between `runner_sync_t` (the one used in the testbench) and `runner_t` (the one accepted by the function). Furthermore, the function or a wrapper would need to be made public.

Alternatively, I think I could use a `post_run`, which I believe is what you mean with `at the function level without involving any external tool`.

vunit/test/acceptance/artificial/vhdl/tb_same_sim_all_pass_nonzero.vhd
```diff
@@ -0,0 +1,39 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
```
kraigher: I would prefer if it was just a unit test at the python level and no new acceptance test or VHDL code was added.

vunit/test/unit/test_test_suites.py
```diff
@@ -107,3 +108,38 @@ def _read_test_results(contents, expected_test_cases):
                           test_suite_name=None,
                           test_cases=expected_test_cases)
             return run._read_test_results(file_name=file_name)  # pylint: disable=protected-access
+
+    def test_exit_code(self):
```
kraigher: I am missing some comment explaining what is the purpose.
I am also missing a test that checks that SKIPPED results are not modified.
I am also missing a test with two result where one is PASSED and one is FAILED and the results are unmodified.

vunit/test/unit/test_test_suites.py
```diff
@@ -107,3 +108,38 @@ def _read_test_results(contents, expected_test_cases):
...
             return run._read_test_results(file_name=file_name)  # pylint: disable=protected-access
+
+    def test_exit_code(self):
+        self.assertEqual(self._test_exit_code(True), False)
+        self.assertEqual(self._test_exit_code(False), False)
+        self.assertEqual(self._test_exit_code(True, True), False)
+        self.assertEqual(self._test_exit_code(False, True), True)
+
+    @staticmethod
+    def _test_exit_code(sim_ok=True, with_has_valid_func=False):
```
kraigher: Call it has_valid_exit_code instead of with_has_valid_func?

vunit/test/unit/test_test_suites.py
```diff
@@ -107,3 +108,38 @@ def _read_test_results(contents, expected_test_cases):
...
+                fptr.write("""\
+test_start:test1
+test_suite_done
+""")
+            sim_if = SimulatorInterface
+            if with_has_valid_func:
+                @staticmethod
+                def func():
+                    return True
+                sim_if.has_valid_exit_code = func
```
kraigher: always make a function that returns has_valid_exit_code

vunit/test/acceptance/artificial/vhdl/tb_same_sim_all_pass_nonzero.vhd
umarcor: If you confirm that it is irrelevant the difference between generating some results artificially in a unit test, or actually running several tests in the same simulator instance, I will just drop the additional acceptance test.

vunit/test/unit/test_test_suites.py
umarcor: > I am missing some comment explaining what is the purpose.

I will ad a comment.

> I am also missing a test that checks that SKIPPED results are not modified.

Do you mean that `passed, skipped` + `exit_code != 0` should produce `failed, skipped`?

Also, regarding the SKIPPED status:

> Would it be acceptable to allow PASSED and/or SKIPPED to be passed to call_post_check instead?

At the moment, `call_post_check` is only executed if all the tests PASSED.

> I am also missing a test with two result where one is PASSED and one is FAILED and the results are unmodified.

I will add them too.

vunit/test/unit/test_test_suites.py
umarcor: Related to the next comment, you mean to treat the argument as the variable that is actually returned by the function, instead of a parameter to decide whether to actually redefine the function?

kraigher: Yes that is what I mean. It seems more straightforward to do it that way.

vunit/test/acceptance/artificial/vhdl/tb_same_sim_all_pass_nonzero.vhd
kraigher: I think you can drop the acceptance test. A unit test is enough for this.

vunit/test/unit/test_test_suites.py
kraigher: > Do you mean that `passed, skipped` + `exit_code != 0` should produce `failed, skipped`?

Yes. Or passed, failed, skipped => failed, failed, skipped

> 
> Also, regarding the SKIPPED status:
> 
> > @umarcor:
> > Would it be acceptable to allow PASSED and/or SKIPPED to be passed to call_post_check instead?
> 
> At the moment, `call_post_check` is only executed if all the tests PASSED.
> 

I think post check should run if there are no FAILED tests. Thus it should run if there are SKIPPED tests as long as there are no FAILED tests. Currently there is no way to have SKIPPED tests without a FAILED test though.

umarcor:
I applied changes according to the comments above. Now, the following scenarios are tested:

- `{"test1": PASSED}`
- `{"test1": PASSED, "test2": SKIPPED}`
- `{"test1": FAILED, "test2": SKIPPED}`
- `{"test1": PASSED, "test2": FAILED}`
- `{"test1": PASSED, "test2": FAILED, "test3": SKIPPED}`

For each of them, four different tests are executed:

- `sim_ok=True && has_valid_exit_code=False`, *regular* zero exit.
- `sim_ok=False && has_valid_exit_code=False`, *regular* nonzero exit.
- `sim_ok=True && has_valid_exit_code=True`, *new* zero exit.
- `sim_ok=False && has_valid_exit_code=True`, *new* nonzero exit.

I also added a commit where the style of the other tests is modified, just to reduce verbosity. Functionaly is untouched.

vunit/ui.py
```diff
@@ -813,7 +813,7 @@ def enable_check_preprocessing(self):
         """
         self._check_preprocessor = CheckPreprocessor()
 
-    def main(self, post_run=None):
+    def main(self, post_run=None, exit_on_zero=True):
```
kraigher: This change is unrelated and should not be part of the PR.

vunit/vhdl/run/src/run.vhd
```diff
@@ -103,7 +103,8 @@ package body run_pkg is
     external_failure : boolean := false;
     allow_disabled_errors : boolean := false;
     allow_disabled_failures : boolean := false;
-    fail_on_warning : boolean := false) is
+    fail_on_warning : boolean := false;
```
kraigher: This change is not necessary any more.

vunit/vhdl/run/src/run.vhd
```diff
@@ -129,7 +130,9 @@ package body run_pkg is
       core_pkg.test_suite_done;
     end if;
 
-    if not p_simulation_exit_is_disabled(runner_state) then
+    if do_not_exit then
```
kraigher: This change is not necessary any more.

vunit/vhdl/run/src/run_api.vhd
```diff
@@ -50,7 +50,8 @@ package run_pkg is
     external_failure : boolean := false;
     allow_disabled_errors : boolean := false;
     allow_disabled_failures : boolean := false;
-    fail_on_warning : boolean := false);
+    fail_on_warning : boolean := false;
```
kraigher: This change is not necessary any more.

vunit/test/unit/test_test_suites.py
```diff
@@ -21,77 +22,53 @@ class TestTestSuites(TestCase):
...
-                expected_test_cases=["test1"])
+test_suite_done""", False)
         except RuntimeError as exc:
             self.assertIn("unknown test case test3", str(exc))
         else:
             assert False, "RuntimeError not raised"
 
-    @staticmethod
-    def _read_test_results(contents, expected_test_cases):
+    def _read_test_results(self, expected, contents, do_assert=True):
```
kraigher: Adding do_assert argument was not necessary.

vunit/test/unit/test_test_suites.py
```diff
@@ -21,77 +22,53 @@ class TestTestSuites(TestCase):
...
     def test_read_results_unknown_test(self):
         try:
-            self._read_test_results(
-                contents="""\
+            self._read_test_results(["test1"], """\
 test_start:test1
 test_start:test3
-test_suite_done""",
-                expected_test_cases=["test1"])
+test_suite_done""", False)
```
kraigher: do_assert=False is not necessary.

vunit/test/unit/test_test_suites.py
```diff
@@ -105,5 +82,74 @@ def _read_test_results(contents, expected_test_cases):
...
+            if do_assert:
+                self.assertEqual(results, expected)
+            return results
+
+    def test_exit_code(self):
+        """
+        Test that results are overwritten when none is FAILED but the exit code is nonzero
+        """
+
+        def test(contents, results, expected=None, werechecked=None):
```
kraigher: Terminology is not consistent. werechecked is sometimes called done.

vunit/test/unit/test_test_suites.py
```diff
@@ -105,5 +82,74 @@ def _read_test_results(contents, expected_test_cases):
...
+            {"test1": FAILED, "test2": SKIPPED},
+            [False, False, False, True]
+        )
+
+        test("""\ntest_start:test1\n""", {"test1": FAILED, "test2": SKIPPED})
+        contents = """\ntest_start:test1\ntest_start:test2\n"""
+        test(contents, {"test1": PASSED, "test2": FAILED})
+        test(contents, {"test1": PASSED, "test2": FAILED, "test3": SKIPPED})
+
+    def _test_exit_code(self, contents, expected, sim_ok=True, has_valid_exit_code=False, waschecked=False):
```
kraigher: Terminology is not consistent, waschecked vs. done

vunit/test/unit/test_test_suites.py
umarcor: One of the cases does not use `assertEqual`, so I wanted to preserve it. However, as you say, it will fail before the assertion anyway.

I removed the option now.

vunit/test/unit/test_test_suites.py
umarcor: See previous comment.

vunit/test/unit/test_test_suites.py
umarcor: `werechecked` is a list of `waschecked`, which is the expected Boolean value of `done`. `done` is the actual value returned by function `_check_results`.

I could rename `werechecked` to `were_expected_to_be_checked` and `waschecked` to `was_expected_to_be_checked`; so `done` would be `was_actually_checked`.

For now, I changed it to ommit naming `done` explicitly, by comparing dicts at once, instad of using two asserts.

vunit/test/unit/test_test_suites.py
umarcor: See comment above.

umarcor:
> VUnit ignores the return code to be able to end simulation with an assert of severity failure for non VHDL-2008 where std.end.stop exists.

I am hitting this issue while testing #476 with `vhdl_standard="93"`. A normal execution with VUnit is partially correct:

- The C code before the simulation is executed.
- The simulation is executed.
- When the simulation exists, the wrapper exists, so no C post-checks are executed.
- The python script does not exit.

``` bash
Running test: lib.tb_external_buffer.test
Running 1 tests

Starting lib.tb_external_buffer.test
Output file: vunit_out/test_output/lib.tb_external_buffer.test_2f235d631488b2560e734c8488d8d48e8d0cd49e/output.txt
0: 11
1: 22
2: 33
3: 44
4: 55
5: 0
6: 0
7: 0
8: 0
9: 0
10: 0
11: 0
12: 0
13: 0
14: 0
               0 fs - default              -    INFO - Init test
               0 fs - default              -    INFO - End test
/src/vunit/vhdl/core/src/stop_body_93-2002.vhd:10:5:@0ms:(report failure): Stopping simulation with status 0
/src/examples/vhdl/external_buffer/vunit_out/test_output/lib.tb_external_buffer.test_2f235d631488b2560e734c8488d8d48e8d0cd49e/ghdl/tb_
external_buffer-tb:error: report failed
/src/examples/vhdl/external_buffer/vunit_out/test_output/lib.tb_external_buffer.test_2f235d631488b2560e734c8488d8d48e8d0cd49e/ghdl/tb_
external_buffer-tb:error: simulation failed
pass (P=1 S=0 F=0 T=1) lib.tb_external_buffer.test (0.4 seconds)

==== Summary =======================================
pass lib.tb_external_buffer.test (0.4 seconds)
====================================================
pass 1 of 1
====================================================
Total time was 0.4 seconds
Elapsed time was 0.4 seconds
====================================================
All passed!
EXCEPT
HELLO
```

But, when the binary is loaded dynamically from Python through `ctypes`, the execution is aborted:

- The C code before the simulation is executed.
- The simulation is executed.
- When the simulation exists, the wrapper receives a SIGABRT, which can be captured. So C post-checks can be potentially executed.
- However, `Aborted` is printed (I don't know who prints it) and the python script is terminated. 

``` bash
0: 11
1: 22
2: 33
3: 44
4: 55
5: 127
6: 0
7: 0
8: 176
9: 255
10: 41
11: 1
12: 0
13: 0
14: 0
               0 fs - default              -    INFO - Init test
               0 fs - default              -    INFO - End test
/src/vunit/vhdl/core/src/stop_body_93-2002.vhd:10:5:@0ms:(report failure): Stopping simulation with status 0
/src/examples/vhdl/external_buffer/vunit_out/test_output/lib.tb_external_buffer.test_2f235d631488b2560e734c8488d8d48e8d0cd49e/ghdl/tb_
external_buffer-tb:error: report failed

SIGABRT caught!
Aborted
```

On the one hand, I wonder whether it is possible to capture the exit in the C code for a normal VUnit simulation. It is surprising to me that SIGABRT is not used. Nonetheless, if I could capture when `ghdl_main` exits, it would be possible to run post-checks in C even with VHDL <2008.

On the other hand, it should be possible to catch the abortion in the shared library and allow python to continue. When data is allocated from python, it should not be relevant how the simulation exited, because data will be written already. Alternatively, I should test if the abortion is produced because an executable binary is loaded dynamically; i.e., compiling the binary as a shared library explicitly might fix this (ghdl/ghdl#800).

PR #476, opened 2019-04-15
[2/3] Add external modes to string_ptr and integer_vector_ptr (through VHPIDIRECT)

umarcor:
Based on #482 and #507.
Refs #462, #465, #470.

> @kraigher [wrote](https://github.com/VUnit/vunit/pull/470#issuecomment-482924422):
> So in this case the first task becomes to add a dynamic byte array data type to VUnit that can have an external implementation in a C-object. It is easier to review if such a task is done in isolation as a separate PR.

In this PR, external modes are added to `integer_vector_ptr.vhd`, following the same approach as in #507.

---

In order to support using `*_ptr` with or without VHPI, two different implementations of the external resources are provided. One of them, `external_*_pkg-vhpi.vhd`, declares the functions/procedures as external; therefore, C implementations must be provided. The other one, `external_*_pkg-novhpi.vhd` does not declare the functions/procedures as external; C implementations are not required, but it is not possible to create vectors with `id/=0` (an assertion of level error is raised).

The list of objects is added through `set_sim_option("ghdl.elab_flags", ["-Wl," + " ".join(files)])`.

---

An example is added, `external_buffer`, to test accessing the same external buffer/array using two methods. The external C application allocates a buffer of length 15 and writes to the first 5 positions. In the VHDL testbench, two `integer_vector_ptr` are created. The first one is used to copy the first five elements to positions 5-9. And the second one is used to copy data from positions 5-9, to positions 10-14. The C application prints all the positions before and after the execution of the simulation.

---

`deallocate`, `reallocate` and `resize` are not implemented for external models. For the case `id>0`, it would be easy to implement it. Since the 'connection' is a pointer, it is possible to use `malloc` or `mmap` in the C implementation. Coherently, it would be possible to update the length in the VHDL model according to the actual size in the C app.

However, when `extfnc`, it can be complex or not possible. Since `read_byte`/`write_byte` can be used to interact with external services/processes through pipes, sockets, packets, messages, etc., the implementation is likely to be very specific. Nonetheless, in this example the implementation can be the same for both 'access' modes.

---

ghdl/ghdl#797

### DIFF
--- a/examples/vhdl/external_buffer/run.py
+++ b/examples/vhdl/external_buffer/run.py
@@ -25,47 +25,35 @@ copied/modified. The content of the buffer is printed both before and after the
 simulation.
 """
 
-from vunit import VUnit
+from vunit import VUnit, ROOT
 from os import popen
 from os.path import join, dirname
 
 src_path = join(dirname(__file__), "src")
+ext_srcs = join(ROOT, "vunit", "vhdl", "data_types", "src", "external", "ghdl")
 
 # Compile C applications to an objects
 c_iobj = join(src_path, "imain.o")
 c_bobj = join(src_path, "bmain.o")
 
-print(
-    popen(
-        " ".join(
-            [
-                "gcc",
-                "-fPIC",
-                "-DTYPE=int32_t",
-                "-c",
-                join(src_path, "main.c"),
-                "-o",
-                c_iobj,
-            ]
-        )
-    ).read()
-)
-
-print(
-    popen(
-        " ".join(
-            [
-                "gcc",
-                "-fPIC",
-                "-DTYPE=uint8_t",
-                "-c",
-                join(src_path, "main.c"),
-                "-o",
-                c_bobj,
-            ]
-        )
-    ).read()
-)
+for val in [["int32_t", c_iobj], ["uint8_t", c_bobj]]:
+    print(
+        popen(
+            " ".join(
+                [
+                    "gcc",
+                    "-fPIC",
+                    "-DTYPE=" + val[0],
+                    "-I",
+                    ext_srcs,
+                    "-c",
+                    join(src_path, "main.c"),
+                    "-o",
+                    val[1],
+                ]
+            )
+        ).read()
+    )
 
 # Enable the external feature for strings/byte_vectors and integer_vectors
 vu = VUnit.from_argv(vhdl_standard="2008", compile_builtins=False)
@@ -76,8 +64,16 @@ lib.add_source_files(join(src_path, "tb_ext_*.vhd"))
 
 # Add the C object to the elaboration of GHDL
 for tb in lib.get_test_benches(pattern="*tb_ext*", allow_empty=False):
-    tb.set_sim_option("ghdl.elab_flags", ["-Wl," + c_bobj], overwrite=True)
+    tb.set_sim_option(
+        "ghdl.elab_flags",
+        ["-Wl," + c_bobj, "-Wl,-Wl,--version-script=" + join(ext_srcs, "grt.ver")],
+        overwrite=True,
+    )
 for tb in lib.get_test_benches(pattern="*tb_ext*_integer*", allow_empty=False):
-    tb.set_sim_option("ghdl.elab_flags", ["-Wl," + c_iobj], overwrite=True)
+    tb.set_sim_option(
+        "ghdl.elab_flags",
+        ["-Wl," + c_iobj, "-Wl,-Wl,--version-script=" + join(ext_srcs, "grt.ver")],
+        overwrite=True,
+    )
 
 vu.main()
--- a/examples/vhdl/external_buffer/run.py
+++ b/examples/vhdl/external_buffer/run.py
@@ -31,22 +31,53 @@ from os.path import join, dirname
 
 src_path = join(dirname(__file__), "src")
 
-c_obj = join(src_path, "main.o")
-# Compile C application to an object
+# Compile C applications to an objects
+c_iobj = join(src_path, "imain.o")
+c_bobj = join(src_path, "bmain.o")
+
 print(
     popen(
-        " ".join(["gcc", "-fPIC", "-c", join(src_path, "main.c"), "-o", c_obj])
+        " ".join(
+            [
+                "gcc",
+                "-fPIC",
+                "-DTYPE=int32_t",
+                "-c",
+                join(src_path, "main.c"),
+                "-o",
+                c_iobj,
+            ]
+        )
     ).read()
 )
 
-# Enable the external feature for strings
+print(
+    popen(
+        " ".join(
+            [
+                "gcc",
+                "-fPIC",
+                "-DTYPE=uint8_t",
+                "-c",
+                join(src_path, "main.c"),
+                "-o",
+                c_bobj,
+            ]
+        )
+    ).read()
+)
+
+# Enable the external feature for strings/byte_vectors and integer_vectors
 vu = VUnit.from_argv(vhdl_standard="2008", compile_builtins=False)
-vu.add_builtins({"string": True})
+vu.add_builtins({"string": True, "integer": True})
 
 lib = vu.add_library("lib")
 lib.add_source_files(join(src_path, "tb_ext_*.vhd"))
 
 # Add the C object to the elaboration of GHDL
-vu.set_sim_option("ghdl.elab_flags", ["-Wl," + c_obj])
+for tb in lib.get_test_benches(pattern="*tb_ext*", allow_empty=False):
+    tb.set_sim_option("ghdl.elab_flags", ["-Wl," + c_bobj], overwrite=True)
+for tb in lib.get_test_benches(pattern="*tb_ext*_integer*", allow_empty=False):
+    tb.set_sim_option("ghdl.elab_flags", ["-Wl," + c_iobj], overwrite=True)
 
 vu.main()
--- a/examples/vhdl/external_buffer/src/main.c
+++ b/examples/vhdl/external_buffer/src/main.c
@@ -9,7 +9,13 @@ positions. Then, the VHDL simulation is executed, where the (external) array/buf
 is used. When the simulation is finished, the results are checked. The content of
 the buffer is printed both before and after the simulation.
 
-NOTE: This file is expected to be used along with tb_ext_byte_vector.vhd or tb_ext_string.vhd
+This source file is used for both string/byte_vector and integer_vector.
+Accordingly, TYPE must be defined as uint8_t or int32_t during compilation.
+Keep in mind that the buffer (D) is of type uint8_t*, independently of TYPE, i.e.
+accesses are casted.
+
+NOTE: This file is expected to be used along with tb_ext_string.vhd, tb_ext_byte_vector.vhd
+or tb_ext_integer_vector.vhd
 */
 
 #include <stdio.h>
@@ -21,13 +27,15 @@ extern int ghdl_main (int argc, char **argv);
 uint8_t *D[1];
 const uint32_t length = 5;
 
-// Check procedure, to be executed when GHDL exits.
-// The simulation is expected to copy the first 1/3 elements to positions [1/3, 2/3),
-// while incrementing each value by one, and then copy elements from [1/3, 2/3) to
-// [2/3, 3/3), while incrementing each value by two.
+/*
+ Check procedure, to be executed when GHDL exits.
+ The simulation is expected to copy the first 1/3 elements to positions [1/3, 2/3),
+ while incrementing each value by one, and then copy elements from [1/3, 2/3) to
+ [2/3, 3/3), while incrementing each value by two.
+*/
 static void exit_handler(void) {
-  int i, j, z, k;
-  uint8_t expected, got;
+  uint i, j, z, k;
+  TYPE expected, got;
   k = 0;
   for (j=0; j<3; j++) {
     k += j;
@@ -35,7 +43,7 @@ static void exit_handler(void) {
       z = (length*j)+i;
 
       expected = (i+1)*11 + k;
-      got = D[0][z];
+      got = ((TYPE*)D[0])[z];
       if (expected != got) {
         printf("check error %d: %d %d\n", z, expected, got);
         exit(1);
@@ -50,7 +58,7 @@ static void exit_handler(void) {
 int main(int argc, char **argv) {
   // Allocate a buffer which is three times the number of values
   // that we want to copy/modify
-  D[0] = (uint8_t *) malloc(3*length*sizeof(uint8_t));
+  D[0] = (uint8_t *) malloc(3*length*sizeof(TYPE));
   if ( D[0] == NULL ) {
     perror("execution of malloc() failed!\n");
     return -1;
@@ -58,11 +66,12 @@ int main(int argc, char **argv) {
   // Initialize the first 1/3 of the buffer
   int i;
   for(i=0; i<length; i++) {
-    D[0][i] = (i+1)*11;
+    ((TYPE*)D[0])[i] = (i+1)*11;
   }
   // Print all the buffer
+  printf("sizeof: %lu\n", sizeof(TYPE));
   for(i=0; i<3*length; i++) {
-    printf("%d: %d\n", i, D[0][i]);
+    printf("%d: %d\n", i, ((TYPE*)D[0])[i]);
   }
 
   // Register a function to be called when GHDL exits
@@ -72,26 +81,42 @@ int main(int argc, char **argv) {
   return ghdl_main(argc, argv);
 }
 
-// External through access (mode = extacc)
+// External string/byte_vector through access (mode = extacc)
 
 void set_string_ptr(uint8_t id, uint8_t *p) {
-  //printf("C set_string_ptr(%d, %p)\n", id, p);
   D[id] = p;
 }
 
 uintptr_t get_string_ptr(uint8_t id) {
-  //printf("C get_string_ptr(%d): %p\n", id, D[id]);
   return (uintptr_t)D[id];
 }
 
-// External through functions (mode = extfnc)
+// External string/byte_vector through functions (mode = extfnc)
 
 void write_char(uint8_t id, uint32_t i, uint8_t v ) {
-  //printf("C write_char(%d, %d): %d\n", id, i, v);
   D[id][i] = v;
 }
 
 uint8_t read_char(uint8_t id, uint32_t i) {
-  //printf("C read_char(%d, %d): %d\n", id, i, D[id][i]);
   return D[id][i];
+}
+
+// External integer_vector through access (mode = extacc)
+
+void set_intvec_ptr(uint8_t id, uintptr_t *p) {
+  D[id] = (uint8_t*)p;
+}
+
+uintptr_t get_intvec_ptr(uint8_t id) {
+  return (uintptr_t)D[id];
+}
+
+// External integer_vector through functions (mode = extfnc)
+
+void write_integer(uint8_t id, uint32_t i, int32_t v) {
+  ((int32_t*)D[id])[i] = v;
+}
+
+int32_t read_integer(uint8_t id, uint32_t i) {
+  return ((int32_t*)D[id])[i];
 }
--- a/examples/vhdl/external_buffer/src/main.c
+++ b/examples/vhdl/external_buffer/src/main.c
@@ -21,10 +21,8 @@ or tb_ext_integer_vector.vhd
 #include <stdio.h>
 #include <stdlib.h>
 #include <stdint.h>
+#include "vhpidirect_user.h"
 
-extern int ghdl_main (int argc, char **argv);
-
-uint8_t *D[1];
 const uint32_t length = 5;
 
 /*
@@ -34,7 +32,7 @@ const uint32_t length = 5;
  [2/3, 3/3), while incrementing each value by two.
 */
 static void exit_handler(void) {
-  uint i, j, z, k;
+  unsigned i, j, z, k;
   TYPE expected, got;
   k = 0;
   for (j=0; j<3; j++) {
@@ -79,44 +77,4 @@ int main(int argc, char **argv) {
 
   // Start the simulation
   return ghdl_main(argc, argv);
-}
-
-// External string/byte_vector through access (mode = extacc)
-
-void set_string_ptr(uint8_t id, uint8_t *p) {
-  D[id] = p;
-}
-
-uintptr_t get_string_ptr(uint8_t id) {
-  return (uintptr_t)D[id];
-}
-
-// External string/byte_vector through functions (mode = extfnc)
-
-void write_char(uint8_t id, uint32_t i, uint8_t v ) {
-  D[id][i] = v;
-}
-
-uint8_t read_char(uint8_t id, uint32_t i) {
-  return D[id][i];
-}
-
-// External integer_vector through access (mode = extacc)
-
-void set_intvec_ptr(uint8_t id, uintptr_t *p) {
-  D[id] = (uint8_t*)p;
-}
-
-uintptr_t get_intvec_ptr(uint8_t id) {
-  return (uintptr_t)D[id];
-}
-
-// External integer_vector through functions (mode = extfnc)
-
-void write_integer(uint8_t id, uint32_t i, int32_t v) {
-  ((int32_t*)D[id])[i] = v;
-}
-
-int32_t read_integer(uint8_t id, uint32_t i) {
-  return ((int32_t*)D[id])[i];
 }
--- a/examples/vhdl/external_buffer/src/tb_ext_integer_vector.vhd
+++ b/examples/vhdl/external_buffer/src/tb_ext_integer_vector.vhd
@@ -1,0 +1,55 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this file,
+-- You can obtain one at http://mozilla.org/MPL/2.0/.
+--
+-- Copyright (c) 2014-2019, Lars Asplund lars.anders.asplund@gmail.com
+
+-- NOTE: This file is expected to be used along with foreign languages (C)
+-- through VHPIDIRECT: https://ghdl.readthedocs.io/en/latest/using/Foreign.html
+-- See main.c for an example of a wrapper application.
+
+--library vunit_lib;
+--context vunit_lib.vunit_context;
+
+library vunit_lib;
+use vunit_lib.run_pkg.all;
+use vunit_lib.logger_pkg.all;
+use vunit_lib.types_pkg.all;
+use vunit_lib.integer_vector_ptr_pkg.all;
+
+entity tb_external_integer_vector is
+  generic ( runner_cfg : string );
+end entity;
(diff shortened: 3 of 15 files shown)


vunit/test/acceptance/test_external_run_scripts.py
```diff
@@ -115,6 +116,13 @@ def test_vhdl_array_axis_vcs_example_project(self):
     def test_vhdl_axi_dma_example_project(self):
         self.check(join(ROOT, "examples", "vhdl", "axi_dma", "run.py"))
 
+    @unittest.skipIf(
+        GHDLInterface.supports_vhpi(),
```
kraigher: I do not understand why a GHDLInterface method is called here.

vunit/ui.py
```diff
@@ -870,6 +877,20 @@ def _main(self, post_run):
         all_ok = self._main_run(post_run)
         return all_ok
 
+    def _add_ext_pkg(self):
```
kraigher: I think this needs a better name than "ext_pkg"

vunit/ui.py
```diff
@@ -461,6 +461,13 @@ def add_source_files_from_csv(self, project_csv_path, vhdl_standard=None):
                     LOGGER.error("More than one library and one file in csv description")
         return files
 
+    def set_objects(self, files):
```
kraigher: The name set_objects is to general. You assume the object file will contain a VHPI implementation of write_byte and read_byte.

vunit/ui.py
```diff
@@ -870,6 +877,20 @@ def _main(self, post_run):
         all_ok = self._main_run(post_run)
         return all_ok
 
+    def _add_ext_pkg(self):
+        """
+        Add sources corresponding to VHPIDIRECT arrays (or their placeholders)
+        """
+        extmem = "ext_pkg-novhpi.vhd"
+        if self._project.vhpi:
```
kraigher: The name of the vhpi field is to general. You are assuming it means a VHPI implementation of read_byte/write_byte was added.

vunit/ui.py
```diff
@@ -894,6 +915,7 @@ def _main_run(self, post_run):
         """
         simulator_if = self._create_simulator_if()
         test_list = self._create_tests(simulator_if)
+        self._add_ext_pkg()
```
kraigher: I do not think this belongs here.

vunit/ui.py
kraigher: Also I think the point where the external byte array support should be selected by the user is in the add_data_types method. That method should take a file name for an alternate implementation.

vunit/test/acceptance/test_external_run_scripts.py
umarcor: It is used to check if the simulator supports VHPI before trying to run the tests. I think that it might be possible to replace it with `SimulatorInterface`, even if `supports_vhpi` is not implemented in other simulator interfaces. But I am not sure about it.

vunit/ui.py
umarcor: What is your suggestion? I have considered `external_memory_pkg`, `extmem_pkg`, `external_pkg`, `ext_pkg`, `vhpi_pkg`, etc.

vunit/ui.py
umarcor: This is not required. It is a general purpose method to add pre-built objects to the linker args in the elaboration command. I can be used to wrap GHDL in a C program, to provide the implementation of VHPI functions/procedures or for both features at the same time. In this example, both features are used. But that's not a requirement.

Moreover, the provided VHPI functions can be any other defined by the user in the testbench or in custom packages. This is because VHPI functions declared in multiple different VHDL sources/modules can then be provided all together in a single fat object or in multiple slimmer objects.

Nonetheless, I agree that the name could be better. But I think that the name is tightly related to how does the user tell that VHPI is to be used. E.g., the user might want to use VHPI for custom functions, but not have the 'external' vectors/memories. In this context, the objects added through `set_objects` are not required to define neither `write_byte` nor `read_byte`. Precisely, `extmem_pkg-novhpi` should be added to the project.

vunit/ui.py
umarcor: > The name of the vhpi field is to general. You are assuming it means a VHPI implementation of read_byte/write_byte was added.

Yes, it is too general. However, I am not worried about it for now. It is just a placeholder. I think that the relevant issue here is the granularity. Where should this variable be defined? Per project? Per testbench? Per configuration?

Once we decide that, I think that the name of the option should be directly related to the new name of `ext_pkg`, since this variable is precisely used when the package is imported/added.

> Also I think the point where the external byte array support should be selected by the user is in the add_data_types method. That method should take a file name for an alternate implementation.

I could not find any method named `add_data_types`. Do you mean that it should be added? If so, wouldn't it be the same as `set_objects`?

I am not sure about what you mean with *should take a file name for an alternate implementation*. Is 'file name' a C source file? A built C object? A VHDL source file?

Regarding C sources, you commented in #462 that VUnit should not built them. Also, I cannot think of a use case where `ext_pkg-*vhpi.vhd` need to be modified by the user. That's why I think that `add_data_types` and `set_objects` might be the same indeed.

For completeness, there is a use case where the use might want to provide an alternative `ext_pkg-novhpi.vhd` file: to provide a custom model which is external to VUnit but still implemented in VHDL only. However, I think that this is out of the scope of my proposal. Certainly, I would name it 'custom model', instead of 'external model'.

vunit/ui.py
umarcor: I know. But I could not find how and at which point do you decide to compile `*200x` or `*-93` variants of certain files. I think that the same approach might be valid here.

vunit/test/acceptance/test_external_run_scripts.py
kraigher: The acceptance tests are run with different environment variables to select different simulators. A call like `GHDLInterface.supports_vhpi()` would always return the same thing regardless of which simulator that was selected and is supposed to run the test. The easiest way is just to currently only run the test if GHDL is selected by having a `simulator_is("ghdl")`. 

A more future proof way would be to add a new function to vunit/test/common.py called `simulator_check` that takes a lamba function which will take the simulator class of the selected simulator and return True or False based on if the test should run or not. For example:
```python
@unittest.skipIf(simulator_check(lambda simclass: not simclass.supports_vhpi())
```

vunit/ui.py
kraigher: The purpose of the package is to select the backend/implementation of the byte array.
Thus the method could be called: `set_external_byte_array_implementation`

vunit/ui.py
kraigher: My comment was based on the observation that it set `self._project.vhpi = True` which in your code implied that there was an external byte array implementation. It was thus not a general solution. The user could not use a set_object call to add another VHPI-object unrelated to the byte array.

If we should have a general `set_object` method on the VUnit API I think it should be a separate PR since it is not strictly necessary to achieve the goal of this PR. Also we need to think about how it is presented to the user. It just boils down to a sim_option which the user could unintentionally overwrite with another set_sim_option call. It would not be clear to the user how set_object would conflict with set_sim_option.

vunit/ui.py
kraigher: > Yes, it is too general. However, I am not worried about it for now. It is just a placeholder. I think that the relevant issue here is the granularity. Where should this variable be defined? Per project? Per testbench? Per configuration?

Since we need to compile a different package to change the behaviour it needs to be per run.py file. You cannot have different versions of packages for different test benches in the VUnit project model.

> I could not find any method named add_data_types. Do you mean that it should be added? If so, wouldn't it be the same as set_objects?

I remember now that we automatically add the data_types since they are used by the built in runner. In we should make the byte array an optional component that is added explicitly by the user in the run.py file like `add_byte_array(external_implementation="pkg.vhd")`. The verification components would also add the byte array without an external implementation automatically if not already added by the user previously.

vunit/test/acceptance/test_external_run_scripts.py
umarcor: Thanks for the insight. I will try to go with the lambda function.

umarcor: > The easiest way is just to currently only run the test if GHDL is selected by having a `simulator_is("ghdl")`.

I think this would be dirtier than the lambda function, because not all GHDL backends support VHPI. So we would need to evaluate two conditions.

vunit/ui.py
umarcor: I am considering to propose the contribution of 'external' modes for three 'types': `byte_vector` (`string`), `integer_vector` and `memory`. Should we name the method `set_external_array_implementations`? Or do we want to have up to three different methods?

vunit/ui.py
umarcor: > My comment was based on the observation that it set `self._project.vhpi = True` which in your code implied that there was an external byte array implementation. It was thus not a general solution. The user could not use a set_object call to add another VHPI-object unrelated to the byte array.

You are correct, my wording was not good. I meant that it is the placeholder for a general solution. Not that the actual implementation is general as I want it to be.
 
> If we should have a general `set_object` method on the VUnit API I think it should be a separate PR since it is not strictly necessary to achieve the goal of this PR.

I do think it is necessary to achieve this goal. The mechanism to provide an object is the same, independently of which are the functions implemented in it. In this PR we need a mechanism to let the user provide an object, at least.

Nevertheless, it is ok to handle it in a separate PR and then rebase this on top of it.

> Also we need to think about how it is presented to the user. It just boils down to a sim_option which the user could unintentionally overwrite with another set_sim_option call. It would not be clear to the user how set_object would conflict with set_sim_option.

As said, the current implementation is far from being good. I want to trigger this discussion we are having.

For GHDL, the requirement is to provide a sim_option. As you point out, the current implmentation allows the user to unwantedly overwrite the sim_option. As an alternative, I was considering to implement `set_object` and `add_object` as methods that manipulate an attribute of type list (probably named `_objects`) to Configuration. Then, the list would be added in the `_get_command` method of GHDLInterface. So, even if it is a sim_option, it would be treated as a different set.

This is because I think that the integration with other simulators might require handling the objects differently than adding them as a sim_option. For example, with ModelSim and FLI it seems that objects need to be in the path of the binary, but no CLI arg is required: https://opencores.org/usercontent/doc/1380917197, https://gitlab.com/reds-public/fss/tree/master/fss_demo_uart.

vunit/ui.py
umarcor: > Since we need to compile a different package to change the behaviour it needs to be per run.py file. You cannot have different versions of packages for different test benches in the VUnit project model.

Oh, that's unfortunate. But thanks for the clarification. I did need it. Is this a constraint that applies to any package (even those written by the user) or is it only about VUnit's internal packages?

Nonetheless, I think that the solution here is to i) allow the user to provide 'objects' per configuration and ii) provide a 'dummy' C file where stubs of all the required C functions are declared. This will allow the user to set `dummy` for all the configs, and then overwrite the implementation for specific testbenches.

> I remember now that we automatically add the data_types since they are used by the built in runner. In we should make the byte array an optional component that is added explicitly by the user in the run.py file like `add_byte_array(external_implementation="pkg.vhd")`. The verification components would also add the byte array without an external implementation automatically if not already added by the user previously.

Related to a comment above, should we let the option be `add_external_arrays`, so it is possible to decide about `byte_vector`, `integer_vector` and/or `memory`?

vunit/test/acceptance/test_external_run_scripts.py
umarcor: > Thanks for the insight. I will try to go with the lambda function.

This is implemented now.

vunit/ui.py
umarcor: This is avoided now, since external packages are added explicitly through a builtin method.

vunit/ui.py
umarcor: The function is named `add_external_arrays` now. It accepts two optional arguments:

- `packages = [False, False, False]`: whether external packages should be loaded. The first item corresponds to `byte_vector`, the second one to `integer_vector` and the third one to `memory`. For now, the first one is used only.
- `implementations = [[], [], []]`: if the previous is True, optionally provide the list of VHDL files that implement the packages. For now, the first list is used only.

umarcor:
@kraigher, I am having problems to automatically load the data types when the user does not do it neither explicitly (`add_external_arrays()`) nor implicitly (through `add_verification_components()`).

Currently, only the first call to `add_external_arrays()` is effective; all other calls are ignored. When the user calls it explicitly, the user settings are used. When it is called implicitly through `add_verification_components()`, `*-novhpi.vhd` versions are used.

I tried to add `self._add_external_arrays()` in `add_vhdl_builtins()`. The issue is that this is executed as soon as the `vu` instance is created in the `run.py`, so later explicit or implicit calls by the user are ignored. As a results `*-novhpi.vhd` are always used.

Where should I add `builtins._add_external_arrays()` so it is always executed before `vu.main()` but after the user has optionally configured it?

vunit/ui.py
umarcor: See #481.

umarcor:
> Where should I add `builtins._add_external_arrays()` so it is always executed before `vu.main()` but after the user has optionally configured it?

I solved it by saving and checking `self._compile_builtins = compile_builtins`.

This PR is now based on #481.

umarcor:
After talking to @LarsAsplund, I implemented an example of how to allocate memory in Python, instead of doing it in C. A C file is still required, but it is just the 'middleware' between Python and VHDL. All the functionality is implemented either in the testbench or in the Python script.

The main difference compared to other examples, is that the binary is built with `ghdl.elab_e`. Then, the location is retrieved and it is loaded as a shared library in the python script. This allows to use `ctypes` to bind Python and C resources to each other, just as we do between VHDL and C with VHPI.

The example implements two executions:

- REGULAR EXECUTION: the binary is loaded as a shared library, but it is executed with default CLI arguments, similat to a regular VUnit execution.
- PYTHON ALLOCATION: the buffer is allocated in Python (instead of C) after loading the binary as a shared library but before executing main. When the simulation is executed, the Python variable is directly modified by the testbench.

It is further possible to set callbacks between C and Python, i.e., GHDL triggers a C function through VHPI, which itself triggers a Python method through ctypes. This can be useful to, e.g., allow a VHDL procedure push a value/message to a Python queue. Precisely, I think this is the main objective of @LarsAsplund.

umarcor:
Python-GHDL co-execution on any GNU/Linux platform is now supported, as long as GHDL is built with the option in ghdl/ghdl#805.

To Do:

- [ ] Properly document `pkgs` and `impls` options.

kraigher:
I would prefer if focus was on the external memory model to not get too many unrelated threads of development. For the external memory model only byte arrays are needed currently. External integer arrays and string arrays are not needed for an external memory model and has an unclear use-case.

umarcor:
> I would prefer if focus was on the external memory model to not get too many unrelated threads of development.

The PR about the external memory model is #470. This is about external arrays/vectors/buffers, which is completely unrelated to verification components.

Nonetheless, development is not split in multiple threads, all the PR are linear:

#482 -> #476 -> #70

> For the external memory model only byte arrays are needed currently.

Byte arrays are implemented as an alias of `string` and `string_ptr` (see https://github.com/dbhi/vunit/blob/feat-external-arrays/vunit/vhdl/data_types/src/byte_vector_ptr_pkg.vhd and https://github.com/dbhi/vunit/blob/feat-external-arrays/vunit/vhdl/data_types/src/types/byte_vector_pkg.vhd). This is because we need to use `character` as the type between VHDL and C, for it to be an 8 bit type. Therefore, implementing an independent `byte_vector_ptr` would require to unnecessarily duplicate `string_ptr`.

> External integer arrays and string arrays are not needed for an external memory model and has an unclear use-case.

As said, byte arrays are implemented as string arrays. The use case to have external array models which are independent from the memory package in the verification components, is to allow seamless co-simulation. See the example added in this PR: https://github.com/dbhi/vunit/blob/feat-external-arrays/examples/vhdl/external_buffer/run.py#L104-L118. Or the modified `array_axis_vcs` example in #465.

Moreover, the reason to implement external models for both integer vectors and strings, is because AXI interfaces are normally multiples of 32-bits. So, in many use cases this allows for 4x less functions calls for each read/write.

umarcor:
Now that #482 is merged, I believe that this PR is easier to review. However, I will keep it as a Draft yet, because I'd like to discuss some of the latest enhancements/decisions:

---

Currently, `run.py` in example 'external_buffer' accepts argument `--build`, to select between a regular `elab_run` or the 'ghdl.elab_e' mode introduced in #467. When `--build` is used, two artifacts are required: the executable binary and the list of args that VUnit would use to execute it. The solution I use to achieve is quite hackish:
  - When 'ghdl.elab_e' is used, simulation arguments are saved to a `args.txt` in the same same as the executable binary. I think this feature is ok. But, @kraigher, you might want to discuss this in a separate issue.
  - In order to get the location of `args.txt` (the `output_path`) programmatically, a global variable and a `set_pre_config` function are used. This might/should be improved, as discussed in #478.

---

Three C-related files are added to `vunit/vhdl/data_types/src/external`:

- `stubs.c`: stubs of the 8 required functions when `use_external=[True, True]` is used. This is required when multiple testbenches are managed in the same vunit project, but some of them do not use external models.
- `vhpidirect_user.h`: reference implementation that defines a global array of pointers of type byte/char (`uint8_t *D[256]`), and the functions to set/get/read/write them and their content either as byte vectors or as integer vectors. set/get functions are used with vectors of type `>0` and read/write are used with vectors of type `<0`. Including this header file is enough to co-simulate GHDL using the external models introduced in this PR.
- `grt.ver`: if the executable binary produced by VUnit + GHDL is to be loaded dynamically, functions defined in C sources must be visible. This file allows to do so.

Note that these C sources are not used automatically by VUnit. The user must use them in the `run.py` files, as shown in example `external_buffer`. However, the same sources are suitable for multiple examples/designs, and most users won't need to modify them. That's why I added them to the same location as the related VHDL sources.

---

As commented in https://github.com/VUnit/vunit/pull/469#issuecomment-485723516, these enhancements work only partially with VHDL 1993. It is possible to load some data to memory from Python and execute GHDL; but GHDL exits with `abort`, which prevents any Python post-check from being executed. This might be fixed in the future, since it is not an expected behaviour (see https://github.com/ghdl/ghdl/issues/803). For now, I think it is enough to suggest using this feature with VHDL 2008 only.

Some notes about this are added to `examples/vhdl/external_buffer/sigabrt.md`.

---

A Python source is added, `vunit/cosim.py` which implements some utility functions to co-execute a GHDL binary that uses the external models. This includes functions to:

- Open/load and close/unload an dynamically loadable object (executable binary or shared library).
- Convert a list of strings to a format which is suitable for calling a foreign function (e.g. `main` or `ghdl_main`).
- Convert a list of numbers to an array of bytes/integers, suitable to be shared with GHDL, and vice versa.
- Convert a buffer containing an image in row-major order to a base64 encoded PNG image.

Accordingly, a `cosim.py` file is added to example `external_buffer`. This uses some of the functions above and the output of `ghdl.elab_e`. It can be tested with:

``` bash
# cd examples/vhdl/external_buffer
# python3 run.py -v --build
# python3 cosim.py
```

Note that the simulation is executed twice. First, `main` is executed, which relies on the code in `main.c` to allocate, initialize and check the content of the buffers. Second, those tasks are done in Python and `ghdl_main` is executed directly, by just ignoring that `main` exists. This is a very interesting feature, because it is possible to use multiple entrypoints to the simulation (either in C or in Python), by using the same binary.

umarcor:
@kraigher,  @LarsAsplund, in order to show why I find important to implement external strings and integer vectors as an independent feature from verification components, I recorded the following screencast:

vunitcosim

The UUT is the `external_buffer` example explained above. It reads block_len bytes from a buffer and writes them back twice. Once in addresses `block_len to 2*^block_len-1` and another one in addresses `2*block_len to 3*^block_len-1`. The first time the numbers are incremented by one, and the second time they are incremented by two.

In the screencast, first, the design/test is compiled with `python3 run.py -v --build`. This is a normal VUnit execution with option `ghdl.elab_e` (#467). The output is an executable binary generated by GHDL and the CLI args that VUnit would use, which are saved in `args.txt`.

Then, with `cosim.py`, the binary is loaded in Python dynamically and the args are read from the txt file. First, `main` is executed, which allocates data and prints it from C sources. Then, buffers are allocated in Python and `ghdl_main` is executed directly, effectively ignoring any C application code. You can see in the screencast that the C code print lines of type `%i: %i`, while the python code print `py %i: %i`. Note that initial values are different: C starts with `11, 22, 33, 44, 55` and Python starts with `111, 122, 133, 144, 155`.

In this design, one external string buffer is used for the data, and one external integer vector buffer is used to pass parameters such as the block length.

Everything above is already implemented in this PR.

Last, in the screencast a Flask server is started with `serve.py`. This provides a Vue.js frontend that allows to co-simulate interactively. The executable binary and args are loaded as in `cosim.py`, but the simulation is executed step-by-step, instead of running continuously until the end. It is possible to run the simulation for a given number of clock cycles only, or to have it re-triggered. Each time it is triggered, the data in the frontend is updated. Both modes are shown in the screencast: first 'for' is used a couple of times, and then 'every' is used until the simulation ends.

Some notes:

- Since GHDL does not support step-by-step execution, yet, I achieved it through clock gating in the testbench. I implemented addition and comparison functions/procedures for a custom type composed of two 32 bit integers which are defined as part of an external array/buffer of integers. As a result, there is a 64 bit counter and a limit. The limit can be modified from VHDL/GHDL, from Python and/or from Vue.js, so any of the actors in the co-simulation can make it advance. These optional utilities are added to VUnit as a package and an entity in `vunit/vhdl/run/src`.
  - At some point, GHDL might support step-by-step execution: [ghdl/ghdl#800](https://github.com/ghdl/ghdl/issues/800#issuecomment-487084044).
- The Vue.js project is the same for any example. I added it temporally to `examples/vhdl/vue`. For each of the examples/designs, only a subcomponent needs to be added, so it is easier for non-javascript developers to add views for their designs.
- In the screencast, the content of the buffer that is shared between Python and GHDL is shown as a table. This is exactly the buffer where GHDL is writing during simulation, not a (delayed) copy of it. There is a delay between Python and Vue.js, but not between GHDL and Python.
- The view as a table is just a proof of concept. It is possible to represent the arrays of bytes/integers with any custom fancy animation/model that the user might want to implement.

---

vunitcosim_hls

This other example is based on an AXI Master core written in Vivado HLS and exported to VHDL. VUnit's verification components and the external memory from #470 are used. Furthermore, the `std_logic` signals in the top entity of the core are shown in the Vue frontend.

Note that, the buttons/leds representing the `std_logic` signals show the status at the clock cycle when the content is updated. The status between updates is not recorded. However, it is possible to inspect it with GtkWave: since this is a VUnit simulation, after all, it is possible to generate a waveform by providing the corresponding CLI arg.

Instead of a single table, the memory is split in two views. This is just an example to show the flexibility: it is possible to allocate a single large/huge buffer to be shared between Python and GHDL, but only show some blocks of it at certain pointed addresses.

---

vunitcosim_hsconv2

This last example shares a large amount of data through an AXI Stream Slave and reads it from an AXI Stream Master, using VUnit's verification components. Precisely, a sintetic hyperspectral cube in BIP (band interleaved by pixel) format of size 256 x 128 x 20 x 16 bit is shared. With 'sintetic' I mean that the example is not a real hyperspectral image; instead I took 20 images, stacked them together and converted the cube to BIP format.

Since it is not possible to share 16 bit types betwen GHDL and a foreign language, actually an external integer vector is used (32 bits per value). Nonetheless, the point is that Python is used to read the shared buffer and encode each spatial image as a base64 encoded PNG. These PNG images are passed to the Vue frontend, which displays them directly. The user can interactively (during simulation) select which of the bands/layers to display.

Four images are shown, from left to right: the input buffer, the expected reference output, the buffer where GHDL is writing the results, and the difference between the reference and the results.

Once again, this is an example of the flexibility of the approach, to show that we can reuse the Vue project and common `cosim.py` functions for quite different designs. It is possible to show tables along with the images, for better inspection of the data.


---

All of the examples above do read data from the simulation, but they also write some data back. Precisely, each time 'Run' is clicked, four integer values are modified in a shared integer vector. This means that it is possible to allow the frontend to modify the content of the tables and write it back to GHDL while the simulation is waiting to advance. This is not implemented, tho.

I am currently focused on providing structured access to internal buffers of the design. I want to use a single parameter to add some 'virtual copy' to internal BRAMs/FIFOs, so it is possible to inspect them during simulation. I find that all the simulators have problems recording large memories in waveforms, because they need to store lots of metadata. This approach would work around it by allowing to inspect them at different times of the simulation, but without recording everything.

For future work, I think it would be very interesting to further develop the Vue frontend, in order to provide multiple views for application specific cases. Since it is possible to match a VHDL peripheral with a Vue model, all kinds of possibilities arise: some VGA/HDMI receiver/monitor, a line follower robot with a circuit, the 3D model of a BLDC motor (with e.g. Three.js)... However, I won't be able to work on this. Hope some other finds it interesting enough.

Regarding VUnit, I think that this Vue frontend is just a fancy example of what might be possible with external strings and integer vectors. But I believe that the interesting part upstream are the features that allow `cosim.py`, not `serve.py`. This is because `cosim.py` does not require any additional dependency, but `serve.py` needs Flask, and optionally Pillow and numpy.

Moreover, `cosim.py` itself is helpful, but it is not required. This PR alone would allow two of the examples above to be executed with upstream VUnit.

kraigher:
If possible I would like you to reduce the scope of this PR to maybe only external byte arrays and make other enhancements afterwards. It becomes to much in one PR and it makes it harder for me to find the time to review it. To merge a small change at a time is easier to accept than several things at once when my time to get up to speed with what you are doing is limited.

1138-4EB:
> If possible I would like you to reduce the scope of this PR to maybe only external byte arrays and make other enhancements afterwards.

That's the content of just the first commit in this PR: https://github.com/VUnit/vunit/pull/476/commits/7715f0d1278847b99a3a9d5af253d261fa47230d. Should I open a new PR with this?

umarcor:
> If possible I would like you to reduce the scope of this PR to maybe only external byte arrays and make other enhancements afterwards.

That's the content of just the first commit in this PR: https://github.com/VUnit/vunit/pull/476/commits/7715f0d1278847b99a3a9d5af253d261fa47230d. Should I open a new PR with this?

kraigher:
Yes. Smaller PR is always better. It lowers the threshold for reviews.

umarcor:
Now that #507 is merged, this PR is just a follow-up:

- In the first two commits, external modes are added to `integer_vector_ptr`, following the same procedure as with `string_ptr`.
- In the third commit, C functions that are reusable are added to VUnit's codebase:
  - `vhpidirect_user.h` includes functions that correspond to the VHPIDIRECT callbacks used by `vunit/vhdl/data_types/src/external/*.vhd`.
  - `stubs.c` is to be used by testbenches that do not use external modes, in case some other testbench in the project requires external modes.
  - `grt.ver` allows functions to be visible when the binary generated by GHDL is dynamically loaded.
- In the last commit, example `external_buffer` is simplified by using the now built-in C sources.

Note that `vhpidirect_user.h` is the API for foreign processes (Python, Octave, etc.) loading the binary generated by GHDL. Future contributions might add alternative `external*.vhd` and matching reference C sources for other simulators, such as ModeSim/QuestaSim's FLI, which match the API.

PR #483, opened 2019-04-23
check_axi_stream non-blocking

### LINKED ISSUE #477: axi_stream_slave with check function requires TREADY
Hi

Change 3e098f567af0a442ae9f3a41b0fcb90f2092daed modifies the behaviour of the check_stream/check_axi_stream function.
The check functions call the pop function one by one. The command queue to the VC thus always runs empty, causing TREADY going low with the current implementation. This results in max 50% duty cycle of the checked stream.
However, TREADY is optional by the AXIS spec.

I don't grasp the full background of the change above. Is the only reason to have it implemented the same way?

Best regards,
emanuel

eschmidscs:
This resolves #477 

I reused the same function with an additional argument with default value for backward compatibility (compared to the current master).
I prefer that to a separate function, but let me know if you think otherwise.

Only the specific axi_stream stuff is updated. One could think of doing the same to the generic stream_slave, but I can't handle the consequences (would need to update the uart...).

Running all checks in verification_components seems to fail with questa due to some delta-cycle differences with GHDL... :-/ GHDL looks fine here.

### DIFF
--- a/vunit/vhdl/verification_components/src/axi_stream_pkg.vhd
+++ b/vunit/vhdl/verification_components/src/axi_stream_pkg.vhd
@@ -183,6 +183,7 @@ package axi_stream_pkg is
 
   constant push_axi_stream_msg : msg_type_t := new_msg_type("push axi stream");
   constant pop_axi_stream_msg : msg_type_t := new_msg_type("pop axi stream");
+  constant check_axi_stream_msg : msg_type_t := new_msg_type("check axi stream");
   constant axi_stream_transaction_msg : msg_type_t := new_msg_type("axi stream transaction");
 
   alias axi_stream_reference_t is msg_t;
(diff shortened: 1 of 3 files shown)


kraigher:
Did the tests fail in questa on master as well? Unfortunately we cannot run CI on the commercial simulators like we do with GHDL.

eschmidscs:
The test already there, yes. It is the "tb_avalon_master.wait until idle" test in probably all configurations. I did not look into it - it fails with questa but is ok with ghdl.

Then the new testcase "tb_axi_stream.test back-to-back failing check" fails with questa, but not with ghdl. This is related to the setting-up of the logging mock. I played around a bit with the order of the delays and commands. It seems I can make it work with questa easily, but then it fails with ghdl.
I got the feeling that it depends on the scheduling of ready processes in one delta cycle...

PR #493, opened 2019-05-17
Update setup.py to show correct information display options

cjchin:
showing --requires and --license correctly


kraigher:
This PR looks empty when I look in Files Changed.

cjchin:
Ah my apologies. I made a mistake setting my upstream. I will fix it.

PR #507, opened 2019-06-03
[1/3] Add external models to string_ptr; add byte_vector* as an alias

umarcor:
Coming from #476. Related to #470.

In this PR, external modes are added to `string_ptr.vhd` through VHPIDIRECT (see https://ghdl.readthedocs.io/en/latest/using/Foreign.html). With 'external mode' we mean:

- VHDL can use `get`/`set` *callback* functions defined in C (or any C-alike compatible language) to interact with strings (byte vectors).
- An array of pointers to arrays of chars (`char* []`) can be shared between VHDL an any C-alike compatible language.

The usage is as follows:

- `new_string_ptr(len, mode, eid, value)` creates a new array of length `len` with mode `mode` and identifier `id`.
  - If `mode = internal` (default), the array is allocated in the testbench and it is initialized with `value`.
  - If `mode = extfnc`, the array is not allocated. When values are `get`/`set`, VHPIDIRECT functions/procedures `read_char`/`write_char` are called.
  - If `mode = extacc`, the array is not allocated. A pointer is created and the value is retrieved through VHPIDIRECT function `get_string_ptr`. The returned value must be a pointer to an array of type `uint8_t` (size of 1 byte). When data is `get`/`set`, content is directly modified in the shared memory.

Notes:

- When `mode /= internal`, `value` is not used and `len` is used to tell VHDL the maximum size of the external buffer. Currently, this is a fixed value, i.e., there is no 'synchronization' of this value.
- When `mode /= internal`, `eid` is used to decide which of the shared buffers to use.
- These features have been tested with C, Python and C++ (octave) tools.

---

In order to support using `string_ptr` with or without VHPIDIRECT, two different implementations of the external resources are provided:

- `external_string_pkg-vhpi.vhd`, declares the functions/procedures as external; therefore, C implementations must be provided.
- `external_string_pkg-novhpi.vhd` does not declare the functions/procedures as external; C implementations are not required, but it is not possible to create vectors with `mode /= internal` (an assertion of level error is raised).

The user needs to select whether to use external C objects. Furthermore, alternative implementions of `*vhpi.pkg` sources might be provided. This is not expected to be required for most use cases, but it is supported. This is achieved through a new `external` parameter to `from_args` and `from_argv`:

- `None` (defaults to False)
- `{'string': False}`
- `{'string': True}`
- `{'string': ['path/to/custom/file']}`

---

`deallocate`, `reallocate` and `resize` are not implemented for external modes (`mode /= internal`). For the case `extacc`, it would be easy to implement it. Since the 'connection' is a pointer, it is possible to use `malloc` or `mmap` in the C implementation. Coherently, it would be possible to update the length in the VHDL model according to the actual size in the C app.

However, when `extfnc`, it can be complex or not possible. Since `read_char`/`write_char` can be used to interact with external services/processes through pipes, sockets, packets, messages, etc., the implementation is likely to be very specific. Nonetheless, in this example the implementation can be the same for both modes.

---

`byte_vector` is added as an alias of `string`.

### DIFF
--- a/examples/vhdl/external_buffer/run.py
+++ b/examples/vhdl/external_buffer/run.py
@@ -1,0 +1,52 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# Copyright (c) 2014-2019, Lars Asplund lars.anders.asplund@gmail.com
+
+"""
+External Buffer
+---------------
+
+`Interfacing with foreign languages (C) through VHPIDIRECT <https://ghdl.readthedocs.io/en/latest/using/Foreign.html>`_
+
+An array of type ``uint8_t`` is allocated in a C application and some values
+are written to the first ``1/3`` positions. Then, the VHDL simulation is
+executed, where the (external) array/buffer is used.
+
+In the VHDL testbenches, two vector pointers are created, each of them using
+a different access mechanism (``extfunc`` or ``extacc``). One of them is used to copy
+the first ``1/3`` elements to positions ``[1/3, 2/3)``, while incrementing each value
+by one. The second one is used to copy elements from ``[1/3, 2/3)`` to ``[2/3, 3/3)``,
+while incrementing each value by two.
+
+When the simulation is finished, the C application checks whether data was successfully
+copied/modified. The content of the buffer is printed both before and after the
+simulation.
+"""
+
+from vunit import VUnit
+from os import popen
+from os.path import join, dirname
+
+src_path = join(dirname(__file__), 'src')
+
+c_obj = join(src_path, 'main.o')
+# Compile C application to an object
+print(popen(' '.join([
+    'gcc', '-fPIC',
+    '-c', join(src_path, 'main.c'),
+    '-o', c_obj
+])).read())
+
+# Enable the external feature for strings
+vu = VUnit.from_argv(vhdl_standard='2008', compile_builtins=False)
+vu.add_builtins({'string': True})
+
+lib = vu.add_library('lib')
+lib.add_source_files(join(src_path, '*.vhd'))
+
+# Add the C object to the elaboration of GHDL
+vu.set_sim_option('ghdl.elab_flags', ['-Wl,' + c_obj])
+
+vu.main()
--- a/examples/vhdl/external_buffer/run.py
+++ b/examples/vhdl/external_buffer/run.py
@@ -44,7 +44,7 @@ vu = VUnit.from_argv(vhdl_standard='2008', compile_builtins=False)
 vu.add_builtins({'string': True})
 
 lib = vu.add_library('lib')
-lib.add_source_files(join(src_path, '*.vhd'))
+lib.add_source_files(join(src_path, 'tb_ext_*.vhd'))
 
 # Add the C object to the elaboration of GHDL
 vu.set_sim_option('ghdl.elab_flags', ['-Wl,' + c_obj])
--- a/examples/vhdl/external_buffer/src/main.c
+++ b/examples/vhdl/external_buffer/src/main.c
@@ -1,0 +1,97 @@
+/*
+External Buffer
+
+Interfacing with foreign languages (C) through VHPIDIRECT:
+https://ghdl.readthedocs.io/en/latest/using/Foreign.html
+
+An array of type uint8_t is allocated and some values are written to the first 1/3
+positions. Then, the VHDL simulation is executed, where the (external) array/buffer
+is used. When the simulation is finished, the results are checked. The content of
+the buffer is printed both before and after the simulation.
+
+NOTE: This file is expected to be used along with tb_ext_byte_vector.vhd or tb_ext_string.vhd
+*/
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdint.h>
+
+extern int ghdl_main (int argc, char **argv);
+
+uint8_t *D[1];
+const uint32_t length = 5;
+
+// Check procedure, to be executed when GHDL exits.
+// The simulation is expected to copy the first 1/3 elements to positions [1/3, 2/3),
+// while incrementing each value by one, and then copy elements from [1/3, 2/3) to
+// [2/3, 3/3), while incrementing each value by two.
+static void exit_handler(void) {
+  int i, j, z, k;
+  uint8_t expected, got;
+  k = 0;
+  for (j=0; j<3; j++) {
+    k += j;
+    for(i=0; i<length; i++) {
+      z = (length*j)+i;
+
+      expected = (i+1)*11 + k;
+      got = D[0][z];
+      if (expected != got) {
+        printf("check error %d: %d %d\n", z, expected, got);
+        exit(1);
+      }
+      printf("%d: %d\n", z, got);
+    }
+  }
+  free(D[0]);
+}
+
+// Main entrypoint of the application
+int main(int argc, char **argv) {
+  // Allocate a buffer which is three times the number of values
+  // that we want to copy/modify
+  D[0] = (uint8_t *) malloc(3*length*sizeof(uint8_t));
+  if ( D[0] == NULL ) {
+    perror("execution of malloc() failed!\n");
+    return -1;
+  }
+  // Initialize the first 1/3 of the buffer
+  int i;
+  for(i=0; i<length; i++) {
+    D[0][i] = (i+1)*11;
+  }
+  // Print all the buffer
+  for(i=0; i<3*length; i++) {
+    printf("%d: %d\n", i, D[0][i]);
+  }
+
+  // Register a function to be called when GHDL exits
+  atexit(exit_handler);
+
+  // Start the simulation
+  return ghdl_main(argc, argv);
+}
+
+// External through access (mode = extacc)
+
+void set_string_ptr(uint8_t id, uint8_t *p) {
+  //printf("C set_string_ptr(%d, %p)\n", id, p);
+  D[id] = p;
+}
+
+uintptr_t get_string_ptr(uint8_t id) {
+  //printf("C get_string_ptr(%d): %p\n", id, D[id]);
+  return (uintptr_t)D[id];
+}
+
+// External through functions (mode = extfnc)
+
+void write_char(uint8_t id, uint32_t i, uint8_t v ) {
+  //printf("C write_char(%d, %d): %d\n", id, i, v);
+  D[id][i] = v;
+}
+
+uint8_t read_char(uint8_t id, uint32_t i) {
+  //printf("C read_char(%d, %d): %d\n", id, i, D[id][i]);
+  return D[id][i];
+}
--- a/examples/vhdl/external_buffer/src/tb_ext_string.vhd
+++ b/examples/vhdl/external_buffer/src/tb_ext_string.vhd
@@ -1,0 +1,55 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this file,
+-- You can obtain one at http://mozilla.org/MPL/2.0/.
+--
+-- Copyright (c) 2014-2019, Lars Asplund lars.anders.asplund@gmail.com
+
+-- NOTE: This file is expected to be used along with foreign languages (C)
+-- through VHPIDIRECT: https://ghdl.readthedocs.io/en/latest/using/Foreign.html
+-- See main.c for an example of a wrapper application.
+
+--library vunit_lib;
+--context vunit_lib.vunit_context;
+
+library vunit_lib;
+use vunit_lib.run_pkg.all;
+use vunit_lib.logger_pkg.all;
+use vunit_lib.types_pkg.all;
+use vunit_lib.string_ptr_pkg.all;
+
+entity tb_external_string is
+  generic ( runner_cfg : string );
+end entity;
+
+architecture tb of tb_external_string is
+
+  constant block_len : natural := 5;
+
+  constant ebuf: string_ptr_t := new_string_ptr( 3*block_len, extfnc, 0);  -- external through VHPIDIRECT functions 'read_char' and 'write_char'
+  constant abuf: string_ptr_t := new_string_ptr( 3*block_len, extacc, 0);  -- external through access (requires VHPIDIRECT function 'get_string_ptr')
+
+begin
+
+  main: process
+    variable val, ind: integer;
+  begin
+    test_runner_setup(runner, runner_cfg);
+    info("Init test");
+    for x in 1 to block_len loop
+      val := character'pos(get(ebuf, x)) + 1;
+      ind := block_len+x;
+      set(ebuf, ind, character'val(val));
+      info("SET " & to_string(ind) & ": " & to_string(val));
+    end loop;
+    for x in block_len+1 to 2*block_len loop
+      val := character'pos(get(abuf, x)) + 2;
+      ind := block_len+x;
+      set(abuf, ind, character'val(val));
+      info("SET " & to_string(ind) & ": " & to_string(val));
+    end loop;
+    info("End test");
+    test_runner_cleanup(runner);
+    wait;
+  end process;
+
+end architecture;
--- a/vunit/builtins.py
+++ b/vunit/builtins.py
@@ -12,6 +12,7 @@ Functions to add builtin VHDL code to a project for compilation
 from os.path import join, abspath, dirname, basename
 from glob import glob
 from vunit.vhdl_standard import VHDL
+from vunit.test.common import simulator_check
 
 VHDL_PATH = abspath(join(dirname(__file__), "vhdl"))
 VERILOG_PATH = abspath(join(dirname(__file__), "verilog"))
@@ -67,11 +68,47 @@ class Builtins(object):
 
             self._vunit_lib.add_source_file(file_name)
 
-    def _add_data_types(self):
+    def _add_data_types(self, external=None):
         """
-        Add data types packages
+        Add data types packages (sources corresponding to VHPIDIRECT arrays, or their placeholders)
+
+        :param external: struct to select whether to enable external models for 'string'. Allowed values are:
+                         None, {'string': False}, {'string': True} or {'string': ['path/to/custom/file']}.
         """
         self._add_files(join(VHDL_PATH, "data_types", "src", "*.vhd"))
+
+        use_ext = {'string': False}
+        files = {'string': None}
+
+        if external:
+            for ind, val in external.items():
+                if isinstance(val, bool):
+                    use_ext[ind] = val
+                else:
+                    use_ext[ind] = True
+                    files[ind] = val
+
+        for ind in use_ext:
+            if use_ext[ind] and simulator_check(lambda simclass: not simclass.supports_vhpi()):
+                raise RuntimeError("the selected simulator does not support VHPI; must use non-VHPI packages...")
+
+        ext_path = join(VHDL_PATH, "data_types", "src", "external")
+
+        def default_pkg(cond, type_str):
+            """
+            Return name of VHDL file with default VHPIDIRECT foreign declarations.
+            """
+            return join(ext_path, 'external_' + type_str + '-' + ('' if cond else 'no') + 'vhpi.vhd')
+
+        if not files['string']:
+            files['string'] = [
+                default_pkg(use_ext['string'], 'string'),
+                join(ext_path, "external_string-body.vhd")
+            ]
+
+        for ind in files:
+            for name in files[ind]:
+                self._add_files(name)
 
     def _add_array_util(self):
         """
@@ -166,11 +203,14 @@ in your VUnit Git repository? You have to do this first if installing using setu
         """
         self._vunit_lib.add_source_files(join(VERILOG_PATH, "vunit_pkg.sv"))
 
-    def add_vhdl_builtins(self):
+    def add_vhdl_builtins(self, external=None):
         """
         Add vunit VHDL builtin libraries
+
+        :param external: struct to select whether to enable external models for 'string'. Allowed values are:
+                         None, {'string': False}, {'string': True} or {'string': ['path/to/custom/file']}.
         """
-        self._add_data_types()
+        self._add_data_types(external=external)
         self._add_files(join(VHDL_PATH, "*.vhd"))
         for path in ("core", "logging", "string_ops", "check", "dictionary", "run", "path"):
             self._add_files(join(VHDL_PATH, path, "src", "*.vhd"))
--- a/vunit/ui.py
+++ b/vunit/ui.py
@@ -118,7 +118,7 @@ The following simulation options are known.
   A list of PLI file names.
 
 ``ghdl.flags``
-   Extra arguments passed to ``ghdl --elab-run`` command *before* executable specific flags. Must be a list of strings.
+   Extra arguments passed to ``ghdl --elab-run`` command *before* executable specific flags.
    Must be a list of strings.
 
 ``incisive.irun_sim_flags``
@@ -308,7 +308,11 @@ class VUnit(object):  # pylint: disable=too-many-instance-attributes, too-many-p
 
         """
         args = VUnitCLI().parse_args(argv=argv)
-        return cls.from_args(args, compile_builtins=compile_builtins, vhdl_standard=vhdl_standard)
+        return cls.from_args(
+            args,
+            compile_builtins=compile_builtins,
+            vhdl_standard=vhdl_standard
+        )
 
     @classmethod
     def from_args(cls, args, compile_builtins=True, vhdl_standard=None):
@@ -323,9 +327,13 @@ class VUnit(object):  # pylint: disable=too-many-instance-attributes, too-many-p
         :param vhdl_standard: The VHDL standard used to compile files,
                               if None the VUNIT_VHDL_STANDARD environment variable is used
         :returns: A :class:`.VUnit` object instance
-        """
 
-        return cls(args, compile_builtins=compile_builtins, vhdl_standard=vhdl_standard)
+        """
+        return cls(
+            args,
+            compile_builtins=compile_builtins,
+            vhdl_standard=vhdl_standard
+        )
 
     def __init__(self, args, compile_builtins=True, vhdl_standard=None):
         self._args = args
@@ -1079,11 +1087,14 @@ avoid location preprocessing of other functions sharing name with a VUnit log or
                             no_color=self._args.no_color)
         runner.run(test_cases)
 
-    def add_builtins(self):
+    def add_builtins(self, external=None):
         """
         Add vunit VHDL builtin libraries
+
+        :param external: struct to select whether to enable external models for 'string'. Allowed values are:
+                         None, {'string': False}, {'string': True} or {'string': ['path/to/custom/file']}.
         """
-        self._builtins.add_vhdl_builtins()
+        self._builtins.add_vhdl_builtins(external=external)
 
     def add_com(self):
         """
--- a/vunit/vhdl/data_types/src/byte_vector_ptr_pkg.vhd
+++ b/vunit/vhdl/data_types/src/byte_vector_ptr_pkg.vhd
@@ -1,0 +1,103 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this file,
+-- You can obtain one at http://mozilla.org/MPL/2.0/.
+--
+-- Copyright (c) 2014-2019, Lars Asplund lars.anders.asplund@gmail.com
+--
+-- The purpose of this package is to provide a byte vector access type (pointer)
+-- that can itself be used in arrays and returned from functions unlike a
+-- real access type. This is achieved by letting the actual value be a handle
+-- into a singleton datastructure of string access types.
+--
+
+use work.types_pkg.all;
+use work.string_ptr_pkg.all;
+
+package byte_vector_ptr_pkg is
+
+  alias val_t is byte_t;
+
+  alias storage_mode_t is work.string_ptr_pkg.storage_mode_t;
+  alias byte_vector_ptr_t is string_ptr_t;
+  alias null_byte_vector_ptr is null_string_ptr;
+
+  alias new_byte_vector_ptr is new_string_ptr[string, storage_mode_t, integer return ptr_t];
+
+  alias is_external is is_external[ptr_t return boolean];
+  alias deallocate is deallocate[ptr_t];
+  alias length is length[ptr_t return integer];
+
+  impure function new_byte_vector_ptr (
+    length : natural := 0;
+    mode   : storage_mode_t := internal;
+    id     : integer := 0;
+    value  : val_t   := 0
+  ) return ptr_t;
+
+  procedure set (
+    ptr   : ptr_t;
+    index : natural;
+    value : val_t
+  );
+
+  impure function get (
+    ptr   : ptr_t;
+    index : natural
+  ) return val_t;
+
+  procedure reallocate (
+    ptr    : ptr_t;
+    length : natural;
+    value  : val_t := 0
+  );
+
+  procedure resize (
+    ptr    : ptr_t;
+    length : natural;
+    drop   : natural := 0;
+    value  : val_t := 0
+  );
+end package;
+
+package body byte_vector_ptr_pkg is
+  impure function new_byte_vector_ptr (
+    length : natural := 0;
+    mode   : storage_mode_t := internal;
+    id     : integer := 0;
+    value  : val_t   := 0
+  ) return ptr_t is begin
+    return work.string_ptr_pkg.new_string_ptr(length, mode, id, character'val(value));
+  end;
+
+  procedure set (
+    ptr   : ptr_t;
+    index : natural;
+    value : val_t
+  ) is begin
+    work.string_ptr_pkg.set(ptr, index+1, character'val(value));
+  end;
+
+  impure function get (
+    ptr   : ptr_t;
+    index : natural
+  ) return val_t is begin
+    return character'pos(work.string_ptr_pkg.get(ptr, index+1));
+  end;
+
+  procedure reallocate (
+    ptr    : ptr_t;
+    length : natural;
+    value  : val_t := 0
+  ) is begin
+    work.string_ptr_pkg.reallocate(ptr, length, character'val(value));
+  end;
+
+  procedure resize (
+    ptr    : ptr_t;
+    length : natural;
+    drop   : natural := 0;
+    value  : val_t   := 0
+  ) is begin
+    work.string_ptr_pkg.resize(ptr, length, drop, character'val(value));
+  end;
+end package body;
--- a/vunit/vhdl/data_types/src/byte_vector_ptr_pkg.vhd
+++ b/vunit/vhdl/data_types/src/byte_vector_ptr_pkg.vhd
@@ -17,7 +17,6 @@ package byte_vector_ptr_pkg is
 
   alias val_t is byte_t;
 
-  alias storage_mode_t is work.string_ptr_pkg.storage_mode_t;
   alias byte_vector_ptr_t is string_ptr_t;
   alias null_byte_vector_ptr is null_string_ptr;
 
@@ -30,7 +29,7 @@ package byte_vector_ptr_pkg is
   impure function new_byte_vector_ptr (
     length : natural := 0;
     mode   : storage_mode_t := internal;
-    id     : integer := 0;
+    eid    : index_t := -1;
     value  : val_t   := 0
   ) return ptr_t;
 
@@ -63,10 +62,10 @@ package body byte_vector_ptr_pkg is
   impure function new_byte_vector_ptr (
     length : natural := 0;
     mode   : storage_mode_t := internal;
-    id     : integer := 0;
+    eid    : index_t := -1;
     value  : val_t   := 0
   ) return ptr_t is begin
-    return work.string_ptr_pkg.new_string_ptr(length, mode, id, character'val(value));
+    return work.string_ptr_pkg.new_string_ptr(length, mode, eid, character'val(value));
   end;
 
   procedure set (
--- a/vunit/vhdl/data_types/src/external/external_string-body.vhd
+++ b/vunit/vhdl/data_types/src/external/external_string-body.vhd
@@ -1,0 +1,28 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this file,
+-- You can obtain one at http://mozilla.org/MPL/2.0/.
+--
+-- Copyright (c) 2014-2019, Lars Asplund lars.anders.asplund@gmail.com
+
+package body external_string_pkg is
+  procedure write_char (
+    id : integer;
+    i  : integer;
+    v  : character
+  )is begin
+    assert false report "VHPI write_char" severity failure;
+  end;
+
+  impure function read_char (
+    id : integer;
+    i  : integer
+  ) return character is begin
+    assert false report "VHPI read_char" severity failure;
+  end;
+
+  impure function get_ptr (
+    id : integer
+  ) return extstring_access_t is begin
+    assert false report "VHPI get_string_ptr" severity failure;
+  end;
+end package body;
--- a/vunit/vhdl/data_types/src/string_ptr_pkg-body-93.vhd
+++ b/vunit/vhdl/data_types/src/string_ptr_pkg-body-93.vhd
@@ -5,59 +5,185 @@
 -- Copyright (c) 2014-2019, Lars Asplund lars.anders.asplund@gmail.com
 
 package body string_ptr_pkg is
-  shared variable current_index : integer := 0;
-  shared variable ptrs : vava_t := null;
+  type storage_t is record
+    id     : integer;
+    mode   : storage_mode_t;
+    length : integer;
+  end record;
+  constant null_storage : storage_t := (integer'low, internal, integer'low);
+
+  type storage_vector_t is array (natural range <>) of storage_t;
+  type storage_vector_access_t is access storage_vector_t;
+
+  type ptr_storage is record
+    idx   : natural;
+    ptr   : natural;
+    eptr  : natural;
+    idxs  : storage_vector_access_t;
+    ptrs  : vava_t;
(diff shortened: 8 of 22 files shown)


umarcor:
@kraigher, I think that the most important part to understand this PR is https://github.com/dbhi/vunit/blob/ext-strings/vunit/vhdl/data_types/src/string_ptr_pkg-body-200x.vhd#L56-L154. Precisely, https://github.com/dbhi/vunit/blob/ext-strings/vunit/vhdl/data_types/src/string_ptr_pkg-body-200x.vhd#L65-L72:

``` bash
type ptr_storage is record
  id    : integer;
  ptr   : integer;
  eptr  : integer;
  ids   : storage_vector_access_t;
  ptrs  : vava_t;
  eptrs : evava_t;
end record;
```

`id`, `ptr` and `eptr` are just pointers to the latest element in `ids`, `ptrs` and `eptrs`. Precisely, `id` is exactly the same as `current_index` in the master branch.

- `ids` contains an element/reference for each allocated string/vector, no matter if it is external or internal.
- `ptrs` contains an element/reference for each internally allocated string/vector, i.e. when the user uses `new_string_ptr(id=>0)`.
- `eptrs` contains an element/reference for each externally allocated string/vector, i.e. when the user uses `new_string_ptr(id=>n)` where `n>0`.

When the user uses `new_string_ptr(id=>n)` where `n<0`, the reference is only located in `ids`, no additional metainfo is required.

As a result, each time `get`, `set`, `length`, etc. are called, first we retrieve the corresponding storage_t from `ids`, which will tell us what kind of vector we are dealing with. Then, we execute up to three different snippets. E.g.: https://github.com/dbhi/vunit/blob/ext-strings/vunit/vhdl/data_types/src/string_ptr_pkg-body-200x.vhd#L213-L220

vunit/builtins.py
```diff
@@ -68,13 +68,55 @@ def _add_files(self, pattern):
...
+
+        :param use_external: list of a single Boolean, to select whether to enable external models for string.
+        :param impls: optional list of lists containing alternative implementations for external models.
         """
         self._add_files(join(VHDL_PATH, "data_types", "src", "types", "*.vhd"))
         self._add_files(join(VHDL_PATH, "data_types", "src", "*.vhd"))
 
+#        Add sources corresponding to VHPIDIRECT arrays (or their placeholders)
+
+        from vunit.test.common import simulator_check
```
kraigher: We always have imports on the top of the file/module

vunit/builtins.py
```diff
@@ -68,13 +68,55 @@ def _add_files(self, pattern):
 
             self._vunit_lib.add_source_file(file_name)
 
-    def _add_data_types(self):
+    def _add_data_types(self, use_external=None, impls=None):
```
kraigher: The interface seems wierd, why not just have a list of implementation where all of them are enabled by being mentioned in the list itself. Disabling is the absence from the list. Why have a separate boolean array?

vunit/vhdl/data_types/src/byte_vector_ptr_pkg.vhd
```diff
@@ -0,0 +1,37 @@
...
+-- The purpose of this package is to provide a byte vector access type (pointer)
+-- that can itself be used in arrays and returned from functions unlike a
+-- real access type. This is achieved by letting the actual value be a handle
+-- into a singleton datastructure of string access types.
+--
+
+use work.byte_vector_pkg.all;
+use work.string_ptr_pkg.all;
+
+package byte_vector_ptr_pkg is
```
kraigher: I remember the string having indexes starting at 1 to match the VHDL string. A byte array would have indexes starting at 0. I do not see how this is handled?

umarcor:
> The pull request does not contain any testing that the external mechanism works creating a technical debt.

The fact that all the existing tests pass is a proof that `id=0` works as it did before, so backwards compatibility is ensured. Unfortunately, testing that data can/is actually read/written for externally allocated buffers is a little more complex.

Nonetheless, the `external_buffer` example in #476 (which is just two commits ahead of this PR), is precisely meant to test this feature. Please, see: https://github.com/dbhi/vunit/blob/feat-external-arrays/examples/vhdl/external_buffer/run.py. I didn't add it here in order to keep this PR as simple as possible, but I can do it if you want.

vunit/builtins.py
umarcor: We have three possible use cases:

- `*vhpi.vhd` and `*body.vhd`.
- `*novhpi.vhd` and `*body.vhd`.
- `custom.vhd` and `custom-body.vhd`.

`*novhpi.vhd`, `*vhpi.vhd` and `*body.vhd` are part of the VUnit codebase; `custom.vhd` and `custom-body.vhd` are not.

On most cases the user will want to select whether to enable external models or not, and he/she will not want to know where are the sources located in the Vunit codebase. Only in marginal cases will he/she want to provide an alternative implementation.

However, your answer made me realize that python is a dynamic language. Hence, I can use a single list to hold either `None`, `Boolean` or a list of strings.

I will implement this.

vunit/vhdl/data_types/src/byte_vector_ptr_pkg.vhd
umarcor: I didn't find that issue. It is true that string indexes start at 1 while integer_vectors start at 0 ([integer_vector_ptr_pkg-body-200x.vhd#L70](https://github.com/VUnit/vunit/blob/master/vunit/vhdl/data_types/src/integer_vector_ptr_pkg-body-200x.vhd#L70) and [string_ptr_pkg-body-200x.vhd#L74](https://github.com/VUnit/vunit/blob/master/vunit/vhdl/data_types/src/string_ptr_pkg-body-200x.vhd#L74)). But, AFAIK, that is transparent to the user (see [integer_vector_ptr_pkg-body-200x.vhd#L100](https://github.com/VUnit/vunit/blob/master/vunit/vhdl/data_types/src/integer_vector_ptr_pkg-body-200x.vhd#L100) and [string_ptr_pkg-body-200x.vhd#L104](https://github.com/VUnit/vunit/blob/master/vunit/vhdl/data_types/src/string_ptr_pkg-body-200x.vhd#L104)).

Nevertheless, I did implement an offset when an string_ptr is declared as an access and the pointer value is retrieved through VHPIDIRECT:

```
  if s.len /= 0 then --is_external
    if s.len < 0 then
      write_char(s.id, index, value);
    else
      st.eptrs(s.id)(index+1) := value;
    end if;
  else
    st.ptrs(s.id)(index) := value;
  end if;
```

As you see, when the array is of type `eptr` (external through access), the index is incremented. But with `id=0` (current main implementation) or through external functions, the offset is not required.

Apart from that, `byte_vector` is just an alias of `string`. I.e., it is possible to write/read values of type `natural` in a `string_ptr` (and are automatically converted to/from `character`). See the overloaded get, set, resize... in `string_ptr_pkg.vhd`.

Moreover, in example `external_buffer`, `byte_vector` is used instead of `string` and data is read/written to index `0` without error. See https://github.com/dbhi/vunit/blob/feat-external-arrays/examples/vhdl/external_buffer/src/test/tb_external_buffer.vhd and https://user-images.githubusercontent.com/38422348/58498696-dde2c700-817e-11e9-9bd0-d70aa2134fd9.gif

kraigher: I would claim the 1 vs 0 index is not transparent to the user at the string is allocated with range 1 to length and thus you cannot access element 0.

kraigher:
If the example serves as the test of id != 0 I would like to include it in this PR.

vunit/builtins.py
```diff
@@ -68,13 +69,55 @@ def _add_files(self, pattern):
 
             self._vunit_lib.add_source_file(file_name)
 
-    def _add_data_types(self):
+    def _add_data_types(self, external=None):
```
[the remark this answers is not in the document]
umarcor: Done.

umarcor:
> If the example serves as the test of id != 0 I would like to include it in this PR.

It is added now.

BTW, I expect py27-acceptance tests to pass when GHDL is 'fixed' or when #509 is merged.

vunit/vhdl/data_types/src/byte_vector_ptr_pkg.vhd
umarcor: You are correct. It was not properly implemented. I now fixed it and I added three additional tests to `external_buffer` in order to ensure that the six possible cases are tested: string/byte_vector with id 0 / <0 / >0.

examples/vhdl/external_buffer/run.py
```diff
@@ -0,0 +1,39 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
kraigher: The example contains no comment or documentation that explains what is the goal and what is being demonstrated by the example.

examples/vhdl/external_buffer/src/main.c
```diff
@@ -0,0 +1,66 @@
+#include <stdio.h>
```
kraigher: This file contains no comment of any purpose or what is achieved.

examples/vhdl/external_buffer/src/main.c
```diff
@@ -0,0 +1,66 @@
...
+uint8_t read_char(uint8_t id, uint32_t i) {
+  //printf("C read_char(%d, %d): %d\n", id, i, D[id][i]);
+  return D[id][i];
+}
+
+static void exit_handler(void) {
+  int i, j, z, k;
+  uint8_t e, d;
+  k = 0;
+  for (j=0; j<3; j++) {
```
kraigher: Why 3, what does it mean and where is it defined?

examples/vhdl/external_buffer/src/main.c
```diff
@@ -0,0 +1,66 @@
...
+    k += j;
+    for(i=0; i<length; i++) {
+      e = (i+1)*11 + k;
+      z = (length*j)+i;
+      d = D[0][z];
+      if (e != d) {
+        printf("check error %d: %d %d\n", z, e, d);
```
kraigher: It is helpful to distinguish between "got" and "expected" when printing these kinds of errors.

examples/vhdl/external_buffer/src/tb_byte_vector.vhd
```diff
@@ -0,0 +1,39 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
```
kraigher: This file should probably not be an example but a test within the data_types library itself just like the string_ptr has tests. Since the byte_vector is just an alias of string_ptr it does not need to duplicate all tests but could just test that the 0-indexing works.

examples/vhdl/external_buffer/src/tb_external_byte_vector.vhd
```diff
@@ -0,0 +1,50 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
```
kraigher: Lacks documentation how it relates to main.c

vunit/vhdl/data_types/src/string_ptr_pkg-body-93.vhd
```diff
@@ -5,105 +5,300 @@
 -- Copyright (c) 2014-2019, Lars Asplund lars.anders.asplund@gmail.com
 
 package body string_ptr_pkg is
-  shared variable current_index : integer := 0;
-  shared variable ptrs : vava_t := null;
+  type storage_t is record
+    id   : integer;
+    len  : integer; -- 0: default/internal; >0: external through access; <0: external through funcs
```
kraigher: I do not think it is clear to overload the len field in this way. It is better to let the lenght be just that and have an enumeration datatype and field to represent the three different modes.

vunit/vhdl/data_types/src/string_ptr_pkg-body-93.vhd
```diff
@@ -5,105 +5,300 @@
...
+    elsif old'length <= len then
+      acc := new storage_vector_t(0 to acc'length + 2**16);
+      for i in old'range loop acc(i) := old(i); end loop;
+      deallocate(old);
+    end if;
+  end;
 
   impure function new_string_ptr (
-    length : natural := 0
+    len : natural := 0;
```
kraigher: I do not think the length should be called len. In all other VUnit API:s we do not use abbreviated len. Also it is not backwards compatible since it breaks the API as VHDL allows arguments to be specified by name.

examples/vhdl/external_buffer/run.py
umarcor: Indeed, should this be considered an example or an internal test?

examples/vhdl/external_buffer/src/main.c
umarcor: I will add comments.

examples/vhdl/external_buffer/src/tb_byte_vector.vhd
umarcor: Related to the comment above, should I (re)move `tb_byte_vector.vhd` and `tb_string.vhd`, or all the tests?

vunit/vhdl/data_types/src/string_ptr_pkg-body-93.vhd
umarcor: I will change it.

examples/vhdl/external_buffer/run.py
kraigher: An automated test of functionality provided by VUnit is always necessary regardless. The question is if the test can also serve as an example. To serve as an example it must also be pedagogical. However a test should also be documented to some extent. I am fine with just making this an internal test for now.

examples/vhdl/external_buffer/src/tb_byte_vector.vhd
kraigher: The internal string is already covered by tests so it can be removed if that is the only purpose of the file. The internal byte array needs additional tests for the zero indexing and should be moved to the data types test folder.

vunit/vhdl/data_types/src/string_ptr_pkg-body-93.vhd
umarcor: This is related to the following comment in #482:

> GHDL drops a lot of warnings because declaring parameter length hides the function name. Options are renamed to len to avoid it.

So, it is already implemented for integer_vector. But I can revert it if you want.

kraigher: Ok I missed that. The GHDL warning was recently added and we do not want to change our entire api because of it. The warning can be turned off per file anyway. If this was done foe integer_vector it would be good if it was reverted.

examples/vhdl/external_buffer/run.py
umarcor: In the end, I left the two external testbenches here, and added comments so it is also pedagogical.

examples/vhdl/external_buffer/src/main.c
umarcor: Done.

examples/vhdl/external_buffer/src/main.c
umarcor: I added comments now.

examples/vhdl/external_buffer/src/main.c
umarcor: Done.

examples/vhdl/external_buffer/src/tb_byte_vector.vhd
umarcor: Done.

examples/vhdl/external_buffer/src/tb_external_byte_vector.vhd
umarcor: Added.

vunit/vhdl/data_types/src/string_ptr_pkg-body-93.vhd
umarcor: Done. BTW, the code looks much better now.

vunit/vhdl/data_types/src/string_ptr_pkg-body-93.vhd
umarcor: I reverted 'length' being renamed to 'len' and 'value' being renamed to 'val'.

umarcor:
@kraigher, I think I addressed all of the suggested changes. Let me know when you have another look at it.

kraigher:
Ok I will look again. By the way I though the integer vector length -> len revert would become a separate PR. It would have been cleaner as that could be merged on its own.

examples/vhdl/external_buffer/run.py
```diff
@@ -0,0 +1,52 @@
...
```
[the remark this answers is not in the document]
umarcor: I'd say that it is already added. Both the `external_buffer` example and the `tb_byte_vector_ptr` are being actually executed in Travis: https://travis-ci.org/VUnit/vunit/jobs/544507378. `external_buffer` is skipped in `py27-acceptance-ghdl` because mcode backend is used, and foreign language is not supported: https://travis-ci.org/VUnit/vunit/jobs/544507377. `supports_vhpi` is used in the `simulator_check`, which was added in 3488.

I did not add it to the 'examples' list of the docs, tho. This is on purpose. I do not want to do it until you tell me that the codebase is ready to be merged, because some descriptions/variables are subject to be modified according to your feedback.

examples/vhdl/external_buffer/src/tb_ext_string.vhd
```diff
@@ -0,0 +1,54 @@
...
+
+entity tb_external_string is
+  generic ( runner_cfg : string );
+end entity;
+
+architecture tb of tb_external_string is
+
+  constant block_len : natural := 5;
+
+  constant ebuf: string_ptr_t := new_string_ptr( 3*block_len, 1);  -- external through VHPIDIRECT functions 'read_char' and 'write_char'
```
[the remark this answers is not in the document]
umarcor: You are correct. It was very short-sighted by me to 'fix' it internally, but do not make the external API easier to understand.

I will modify `new_string` so it accepts both an id and a mode type. I will also add two new* functions, to avoid the user requiring the mode type to be visible.

umarcor:
> By the way I though the integer vector length -> len revert would become a separate PR. It would have been cleaner as that could be merged on its own.

I will make a separate PR from it.

BTW, regarding this PR, do you think it is worth having separated `vunit/vhdl/data_types/src/types/string_pkg.vhd` and `vunit/vhdl/data_types/src/types/byte_vector_pkg.vhd` files? Can we merge them in a single file? Each of them is just 10 lines, so the file will not be longer than 40 lines (including the license) even after external integer_vector is added later.

examples/vhdl/external_buffer/run.py
```diff
@@ -0,0 +1,52 @@
...
+
+Interfacing with foreign languages (C) through VHPIDIRECT:
+https://ghdl.readthedocs.io/en/latest/using/Foreign.html
+
+An array of type ``uint8_t`` is allocated in a C application and some values
+are written to the first ``1/3``positions. Then, the VHDL simulation is
+executed, where the (external) array/buffer is used.
+
+In the VHDL testbenches, two vector pointers are created, each of them using
+a different access mechanism (``id>0`` or ``id>0``). One of them is used to copy
```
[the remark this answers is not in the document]
umarcor: Fixed.

examples/vhdl/external_buffer/src/tb_ext_string.vhd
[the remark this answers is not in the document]
umarcor: > I will also add two new* functions, to avoid the user requiring the mode type to be visible.

These are not required after all, because the enumeration type is visible through an alias.

bradleyharden:
@umarcor, why did you choose to implement this as an extension of string_ptr_t? To me, it seems like it might work better as three separate types. All three versions have differences that are extremely important yet extremely subtle when implemented as a single type.

umarcor:
> @umarcor, why did you choose to implement this as an extension of string_ptr_t? To me, it seems like it might work better as three separate types.

The functionality, from VHDL's point of view, is unmodified. Extending the same type allows to preserve backwards compatibility and make the 'external' feature plug and play. I.e., any string_ptr that is already being used in any codebase can be exposed simply by adding an option to the declaration. For example, a queue or a stream VC. Moreover, the same codebase can be used either as an internal or external feature by changing a constant/variable/generic.

Implementing it as three separate types would imply to have duplicated codebases for internal/external tests. In order to avoid it, string_ptr should be kept as a 'wrapper', so the three separate types would be 'children' types. Unfortunately, this does not exist in VHDL, does it?

Overall, I think that having 2x3=6 sources which are 95% similar is more complex to maintain that having two sources.

>  All three versions have differences that are extremely important yet extremely subtle when implemented as a single type.

I need to discuss this with @kraigher when he comes back, but by looking at the current proposal (with `case` statements) I think that the codebase of `ptrs` and `eptrs` can be effectively merged (i.e. `internal` and `extacc`). In the end, the difference ATM is who is expected to do the allocation (VHDL for ptrs and C/Python fro eptrs). But it is also possible for VHDL to allocate eptrs and pass the value through `set_string_ptr`. Hence, we should probably let GHDL initialize strings/vectors of type `extacc` if they contain a 'null' pointer when `new_string_ptr` is called.

The only reason I didn't implement an example where VHDL handles a string of type extacc completely (allocate, deallocate, resize, reallocate, etc.) is because I did not find a use case for it. In all the projects where I'm using this, buffers/data is pre-allocated by the foreign program. However, as commented in #521, it makes perfect sense for some user to 'inspect' the content of some VC, queue, or buffer from some fancy external interface (see [#476](https://github.com/VUnit/vunit/pull/476#issuecomment-496632427)).

Nonetheless, type `extfnc` might be different enough to be implemented as a separate type. Related to #521, we could use `get_value` instead of `read_char`/`read_integer`, and `set_value` instead of `write_char`/`write_integer`, by adding a `type` parameter to `get_value` and `set_value`. This would allow to reuse the same C functions to access buffers as arbitrary C types. I don't know how would this play with VHDL, tho. I need to understand better if the dynamic type systems can handle arbitrary records/structs.

vunit/vhdl/data_types/src/string_ptr_pkg.vhd
```diff
@@ -53,29 +77,55 @@ package string_ptr_pkg is
 
   procedure set (
     ptr   : ptr_t;
-    index : natural;
-    value : val_t
+    index : positive := 1;
+    value : val_t    := val_t'low
+  );
+
+  procedure set (
```
LarsAsplund: I think the motivation for having both byte vector ptrs and string ptrs is that sometimes we deal with integers and sometimes we deal with characters. That should be reflected in the set and get subprograms so that you can't use both characters and integers on both pointer types

vunit/vhdl/data_types/src/string_ptr_pkg.vhd
```diff
@@ -36,13 +44,29 @@ package string_ptr_pkg is
   ) return ptr_t;
 
   impure function new_string_ptr (
-    length : natural := 0
+    length : natural := 0;
```
LarsAsplund: If not all parameters are relevant to all storage modes wouldn't it be better to have split the APIs? At least separate `new_str_ptr` form a new API `new_ext_string_ptr`.

examples/vhdl/external_buffer/src/tb_ext_string.vhd
```diff
@@ -0,0 +1,54 @@
...
+
+  main: process
+    variable val, ind: integer;
+  begin
+    test_runner_setup(runner, runner_cfg);
+    info("Init test");
+    for x in 1 to block_len loop
+      val := get(ebuf, x) + 1;
+      ind := block_len+x;
+      set(ebuf, ind, val);
```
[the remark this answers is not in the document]
umarcor: Yes. At one point I thought that string could be used for both, and this is a reminiscence. I removed overloaded `set`/`get` functions from `string_ptr` now. However, for testing purposes, the same bytes that are used to test `byte_vector` are casted to characters in this test bench.

vunit/vhdl/data_types/src/byte_vector_ptr_pkg.vhd
```diff
@@ -0,0 +1,62 @@
...
+
+  alias is_external is is_external[ptr_t return boolean];
+  alias deallocate is deallocate[ptr_t];
+  alias length is length[ptr_t return integer];
+  alias reallocate is reallocate[ptr_t, natural, natural];
+  alias resize is resize[ptr_t, natural, natural, natural];
+
+  procedure set (
+    ptr   : byte_vector_ptr_t;
+    index : natural := 0;
```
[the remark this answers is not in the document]
umarcor: I removed default from here and from string_ptr.

vunit/vhdl/data_types/src/byte_vector_ptr_pkg.vhd
```diff
@@ -0,0 +1,62 @@
...
+    value : natural := 0
```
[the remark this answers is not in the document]
umarcor: There is already a `byte_t` in `memory_pkg` (see https://github.com/VUnit/vunit/search?q=byte_t&unscoped_q=byte_t). Should we move it to the core (`data_types`) and use it from `verification_components`? In which package should we define it?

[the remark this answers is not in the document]
umarcor: > I don't think there should be a default value. Makes the code unreadable if used.

Default values for `set`/`get` are removed now.

vunit/builtins.py
```diff
@@ -68,13 +69,52 @@ def _add_files(self, pattern):
...
         """
-        self._add_files(join(VHDL_PATH, "data_types", "src", "types", "*.vhd"))
         self._add_files(join(VHDL_PATH, "data_types", "src", "*.vhd"))
 
+        use_ext = {'string': False}
+        files = {'string': None}
+
+        if external is not None:
+            for ind in external:
+                val = external[ind]
```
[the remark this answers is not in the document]
umarcor: Done.

vunit/builtins.py
```diff
@@ -68,13 +69,52 @@ def _add_files(self, pattern):
...
+            if use_ext[ind] and simulator_check(lambda simclass: not simclass.supports_vhpi()):
+                raise RuntimeError("the selected simulator does not support VHPI; must use non-VHPI packages...")
+
+        ext_path = join(VHDL_PATH, "data_types", "src", "external")
+
+        def default_pkg(cond, type_str):
+            """
+            Return name of VHDL file with default VHPIDIRECT foreign declarations.
+            """
+            nostr = 'no'
```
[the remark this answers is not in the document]
umarcor: Done.

vunit/ui.py
```diff
@@ -290,12 +290,16 @@ class VUnit(object):  # pylint: disable=too-many-instance-attributes, too-many-p
...
 
     @classmethod
-    def from_argv(cls, argv=None, compile_builtins=True, vhdl_standard=None):
+    def from_argv(cls, argv=None, compile_builtins=True, vhdl_standard=None, external=None):
         """
         Create VUnit instance from command line arguments.
 
         :param argv: Use explicit argv instead of actual command line argument
         :param compile_builtins: Do not compile builtins. Used for VUnit internal testing.
+        :param vhdl_standard: The VHDL standard used to compile files into this library,
```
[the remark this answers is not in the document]
umarcor: See #544.

vunit/vhdl/data_types/src/byte_vector_ptr_pkg.vhd
```diff
@@ -0,0 +1,62 @@
...
+-- that can itself be used in arrays and returned from functions unlike a
+-- real access type. This is achieved by letting the actual value be a handle
+-- into a singleton datastructure of string access types.
+--
+
+use work.byte_vector_pkg.all;
+use work.string_ptr_pkg.all;
+
+package byte_vector_ptr_pkg is
+
```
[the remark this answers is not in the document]
umarcor: I changed it so that string_ptr accepts characters only and byte_vector accepts naturals only. As a result, `byte_vector_ptr_pkg` now contains wrappers around string_ptr_pkg, instead of just plain aliases.

> Wouldn't it be better with shared functionality in a common package.

I don't see how to achieve this. byte_vector does not have any functionality. It is just a wrapper/alias around string_ptr with two differences:

- The data type is `natural`, instead of `character`.
- The first index is `0`, instead of `1`.

vunit/vhdl/data_types/src/external/external_string-body.vhd
```diff
@@ -0,0 +1,28 @@
...
+  impure function read_char (
+    id : integer;
+    i  : integer
+  ) return character is begin
+    assert false report "VHPI read_char" severity failure;
+  end;
+
+  impure function get_ptr (
+    id : integer
+  ) return extstring_access_t is begin
```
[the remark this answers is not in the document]
umarcor: I don't understand this. `extstring_access_t` is mapped to a C pointer to an array of bytes (string). It is not possible to map `ptr_t` (record type) to a C pointer to an array, is it?

Maybe you mean just to rename `extstring_access_t` to `extstring_ptr_t`? I that is the case, I am not sure about it. Please, see `string_pkg`.

vunit/vhdl/data_types/src/string_ptr_pkg.vhd
```diff
@@ -53,29 +77,55 @@ package string_ptr_pkg is
 
   procedure set (
     ptr   : ptr_t;
-    index : natural;
-    value : val_t
+    index : positive := 1;
```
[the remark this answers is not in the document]
umarcor: They are removed now.

vunit/vhdl/data_types/src/string_ptr_pkg.vhd
umarcor: As commented above, this is changed now. byte_vector accepts naturals only, and string accepts characters only.

vunit/vhdl/data_types/src/external/external_string-body.vhd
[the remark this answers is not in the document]
umarcor: I don't understand this. `extstring_access_t` is mapped to a C pointer to an array of bytes (string). It is not possible to map `ptr_t` (record type) to a C pointer to an array, is it?

Maybe you mean just to rename `extstring_access_t` to `extstring_ptr_t`? I that is the case, I am not sure about it. Please, see `string_pkg`.

vunit/vhdl/data_types/src/string_ptr_pkg.vhd
umarcor: On the one hand, I think we should avoid to have separate APIs, at least for `internal` and `extacc` modes. This is because these modes are expected to be swappable:

- Define `mymode` in a *global* package.
- Set `mymode := internal` for a regular VUnit simulation (`id` is ignored) with no VHPIDIRECT support at all. Simulate with e.g. ModelSim.
- Without modifying a single additional VHDL line, set `mymode := extacc`. Now, simulating with GHDL, all the strings are accesible externally and inspectable at runtime.

Note that this is true for any new_string_ptr or new_byte_vector_ptr defined in any component/module of the design. The hierarchy is ignored and there is no need to traverse all the entities, because `id` values are absolute. Of course, care needs to be taken, since multiple ptrs with the same id will access the same shared buffer.

On the other hand, I agree that the current implementation of `length` is not clean enough. Precisely, `storage_t` contains a `length` field because:

- `type extstring_access_t is access string(1 to integer'high);`. It is not possible to define `type extstring_access_t is access string;`, since unconstrained arrays are handled as *fast pointers* by GHDL and *that's looking for problems* (see ghdl/ghdl#797).
- As a result, attribute `length` on strings of type `extacc` will always return `integer'high`, independently of the size that was allocated in C. In most use cases, the required size will be far smaller, so we use `length` when the string_ptr is created to let VUnit know the actual limit of that array.
  - As commented above, it is possible to have these arrays resized/reallocated by either GHDL or C. But a mechanism needs to be implemented in order to update the corresponding `length` field in the other part.
  - Furthermore, the length of each of the buffers in the C part is currently defined through custom variables. I.e., there is no structured approach, such as an array of integers that contains the actual length of each of the buffers in the array of pointers. This can also be implemented.
- With strings of type `extfnc` the point is that no shared pointer exists. All the interactions are done through foreign functioncs.
- With strings of type `internal`, field `length` of type `storage_t` is useless (`0`).

umarcor:
> `reallocate` and `resize` also need some updates WRT available types for the `value`parameter

I split some of the suggested/commented changes which can be merged independently: #542, #543, #544. Then, I rebased this PR on top of them and I cleaned the commits so that types are explicitly different.

vunit/vhdl/data_types/src/types.vhd
```diff
@@ -0,0 +1,37 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
```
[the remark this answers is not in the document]
LarsAsplund: It's probably not a big deal. Most people will use the context anyway. I vote for merging.

vunit/vhdl/data_types/src/byte_vector_ptr_pkg.vhd
```diff
@@ -0,0 +1,108 @@
...
+  alias  val_t  is natural;
```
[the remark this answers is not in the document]
umarcor: There are two mixed issues here:

- `val_t` is already used in string_ptr and integer_vector_ptr in order to make the similarities between both packages explicit. See, side by side:
  - Current differences: https://raw.githubusercontent.com/VUnit/vunit/master/vunit/vhdl/data_types/src/integer_vector_ptr_pkg-body-200x.vhd and https://raw.githubusercontent.com/VUnit/vunit/master/vunit/vhdl/data_types/src/string_ptr_pkg-body-200x.vhd.
  - Expected differences when these PRs are merged: https://raw.githubusercontent.com/dbhi/vunit/feat-external-arrays/vunit/vhdl/data_types/src/integer_vector_ptr_pkg-body-200x.vhd and https://raw.githubusercontent.com/dbhi/vunit/ext-strings/vunit/vhdl/data_types/src/string_ptr_pkg-body-200x.vhd. Note that this is being change as I adapt to your reviews.

- For byte_vector_ptr, we should used `byte_t` instead of natural.

I think that we agree on the latter. Hence, I'm going to complete the proposals in #543 (move `byte_t` and merge *type* packages.

Regarding the first point. What do you think?

vunit/vhdl/data_types/src/string_ptr_pkg-body-93.vhd
```diff
@@ -5,81 +5,215 @@
...
+      acc := new storage_vector_t(0 to acc'length + 2**16);
+      for i in old'range loop acc(i) := old(i); end loop;
+      deallocate(old);
+    end if;
+  end;
 
   impure function new_string_ptr (
     length : natural := 0;
+    mode   : storage_mode_t := internal;
+    id     : integer := 0;
```
kraigher: I assume id=0 means automatic id assignment. This is assumed with internal string. Maybe an check is needed that is it 0 when it cannot be a custom id.

umarcor: Not really. Note that this `id` is NOT the same `id` that was in the codebase until 6 months ago. That was renamed to `ref`: https://github.com/VUnit/vunit/blob/master/vunit/vhdl/data_types/src/string_ptr_pkg.vhd#L20 and https://github.com/VUnit/vunit/blob/master/vunit/vhdl/data_types/src/integer_vector_ptr_pkg.vhd#L21.

The mechanism to share buffers between C and VHDL is a `char* []`, i.e. an array of pointers to an arrays of chars. This `id` is used to decide which of those arrays of chars corresponds to a `string` en VHDL. See `examples/vhdl/external_buffer/src/main.c`:

- If mode is extacc, VHDL will execute `get_string_ptr(id)` to retrieve that pointer address.
- If mode is extfnc, VHDL will pass the id in each call to `write_char` or `read_char`.

In this example, using `write_char` (mode extfnc) is equivalent to VHDL writing directly with mode extacc. However, this is only because of the specific C implementation of `write_char`. In a practical example `write_char` is expected to do some more complex task, such as sending data through HTTP. That's where the id is handy.

When mode is internal, field `id` is not used at all.

Nonetheless, `ref` still exists. Each string, no matter the mode, has a ref; it is assigned automaticallly, as it has always been. See:

https://github.com/VUnit/vunit/blob/ed75c878c5767d82b302d2be16cb25e5d2d9e6fb/vunit/vhdl/data_types/src/string_ptr_pkg-body-200x.vhd#L131-L158

Let's suppose we want to use four strings:

- internal
- extacc with id 0
- internal
- extfnc with id 1

```
        st.ids st.ptr st.eptr
string0    0        0
string1    1               0
string2    2        1
string3    3
```

So,

- `st.ids` is an array of `ref` that contains four entries, one for each string. 
- `st.ptr` is the array of refs/pointers to internal strings (this is the element that has always existed, everything else is wrapped around it), and contains two entries. 
- `st.eptr` is the array of refs/pointers to extacc strings (`id` is used to get_string_ptr once only).

As a result:

`st.ids(0).id` is `0` (index to read/write from/to st.ptr)
`st.ids(1).id` is `0` (index to read/write from/to st.eptr)
`st.ids(2).id` is `1` (index to read/write from/to st.ptr)
`st.ids(3).id` is `1` (id to read/write through read_char and write_char)

Maybe I am overusing the name `id`? Currently, it is:

- A parameter for extacc to get the pointer from C.
- A parameter for extfnc to pass to read_char/write_char.
- The name of the field `st.ids`, which contains a number (id) for each string.
- The name of field `st.ids().id` which contains a number representing either:
  - The corresponding index in `st.ptr`.
  - The corresponding index in `st.eptr`.
  - The `id` parameter used when creating an extfnc string.

kraigher: Ok I understand a bit better, but it still means it is user error to call the function with internal and an explicit id either by name of position?

```vhdl
-- Case 1
new_string_ptr(length, mode => internal, id => 1) -- Not valid
-- Case 2
new_string_ptr(length, mode => internal, id => 0) -- Also not valid
-- Case 3
new_string_ptr(length, mode => internal) -- Valid
```

Maybe it is not so important to check "Case 2" but "Case 1" could easily give an error which might help debugging.

umarcor: Sure. I'd say that such a test existed: *id /= 0 is not supported with mode internal*. Maybe I unwantedly dropped it during some rebase/modification. Let me check.

BTW, I'm also understanding and having a look at the add_builtins comment above.

vunit/ui.py
```diff
@@ -290,14 +290,16 @@ class VUnit(object):  # pylint: disable=too-many-instance-attributes, too-many-p
     """
 
     @classmethod
-    def from_argv(cls, argv=None, compile_builtins=True, vhdl_standard=None):
+    def from_argv(cls, argv=None, compile_builtins=True, vhdl_standard=None, external=None):
```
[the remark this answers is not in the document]
umarcor: > not add external argument here but instead let the user do compile_builtins=False and then add the builtins explicitly later using add_builtins with the external argument

Done.

> It would have been better not to add the builtins by default 

Do you mean to remove `compile_builtins` from `from_argv` and `from_args`? That'd be a breaking change, wouldn't it?

vunit/vhdl/data_types/src/string_ptr_pkg-body-93.vhd
umarcor: I added an assertion for case 1. Note that case 3 defaults to case 2. Therefore, we have no mechanism to tell them apart. Should the default of id be some non-supported value, we could identify case 2.

umarcor: Now:

- I moved `index_t` to `types_pkg`.
- I renamed `id` fields to `eid` and `idx`, to make it easier to tell them apart.
- `eid` is now of type `index_t` and it defaults to `-1`.
- Assertions are added to detect cases 1 and 2, and also when extacc or extfnc are used without providing `eid`.

vunit/vhdl/data_types/src/external/external_string-body.vhd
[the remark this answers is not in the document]
LarsAsplund: You're right. There's a mix but actually more access_t nowadays

vunit/vhdl/data_types/src/byte_vector_ptr_pkg.vhd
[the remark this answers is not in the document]
LarsAsplund: I would probably done it differently and favor an unambiguous type over the emphasis on similarities. However, it's not a showstopper. I can live with this.

umarcor:
> I'm ok with these changes now

🚀🚀🚀

However, let's not merge for now. After lunch, I will re-read all the comments in this issue again, to ensure that me/we are not missing anything. I will also add some further tests (e.g., to use different `eid` values, i.e., multiple vectors).

Thanks for your support @go2sh. Much appreciated.

go2sh:
Awesome work in total (your PR series), this will rock. A did a quick test on my setup and seams to work fine, but I haven't checked the code in detail. Right now I use your code to get comfortable with VPI etc.

vunit/vhdl/data_types/src/string_ptr_pkg-body-93.vhd
umarcor: @kraigher, shall I add tests for these assertions? If so, do they fit in  vunit/vhdl/data_types/test/tb_byte_vector_ptr.vhd?

umarcor:
I added four new files to the `external_buffer` example (`cp.py`, `cp.c`, `tb_extcp_string.vhd` and `tb_extcp_byte_vector.vhd`). Tests are similar to the other set of tbs, but data is copied from the array with `eid=0` to the array with `eid=1`. There was no explicit test for this feature before, only `eid=0` was tested.

@kraigher, if you are ok with the status of this PR, we can merge it and move forward.

kraigher:
I am ok with this.

bradleyharden:
Now that this is merged, will `string_ptr_pkg` be stable? I would like to update #522, but not if I will have to change it again soon.

In that PR, I modify `queue_t` to store string pointers, rather than characters, and I modify `string_ptr_pkg` to keep track of old, unused pointer indices in stack.

umarcor:
The second part of this PR is still pending: #476. Here, `string_ptr` was modified. In #476, the same modifications are applied to `integer_vector_ptr`. Since #522 modifies both types, I'd suggest to either:

- split #522 in two PRs, to have the `string_ptr` part merged now;
- or, wait until #476 is merged and do everything at the same time.

Note that there are other PRs in the series (such as #470 and #568). However, any of those are to be evaluated after your contributions. This is because those higher level features might be adapted/improved based on #522.

bradleyharden:
The modifications to integer_vector_ptr were really only to provide consistency between the types. It's not required. I'll probably get it up and running on top of this PR and then update integer_vector_ptr after #476

PR #512, opened 2019-06-05
Modified merge_coverage to handle long vcover merge commands.

pStefanNilsson:
The vcover command (launched with python subprocess) may exceed the
limit of 32768 characters in the Win32 API when merging too many
coverage files.

This is avoided by creating a temporary text file with the path to
all coverage files to be merged and applying the new text file using
the vcover merge -inputs flag.

### DIFF
--- a/vunit/modelsim_interface.py
+++ b/vunit/modelsim_interface.py
@@ -339,14 +339,16 @@ proc _vunit_sim_restart {} {
         if args is None:
             args = []
 
-        vcover_cmd = [join(self._prefix, 'vcover'), 'merge'] + args + [file_name]
-
-        for coverage_file in self._coverage_files:
-            if file_exists(coverage_file):
-                vcover_cmd.append(coverage_file)
-            else:
-                LOGGER.warning("Missing coverage file: %s", coverage_file)
-
+        vcover_cmd = [join(self._prefix, 'vcover'), 'merge', '-inputs',  join(self._output_path, 'coverage_files.txt')] + args + [file_name]
+        coverage_files = join(self._output_path, 'coverage_files.txt')
+        with open(coverage_files, "w") as fptr:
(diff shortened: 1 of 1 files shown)


vunit/modelsim_interface.py
```diff
@@ -339,14 +339,16 @@ def merge_coverage(self, file_name, args=None):
...
 
-        vcover_cmd = [join(self._prefix, 'vcover'), 'merge'] + args + [file_name]
-
-        for coverage_file in self._coverage_files:
-            if file_exists(coverage_file):
-                vcover_cmd.append(coverage_file)
-            else:
-                LOGGER.warning("Missing coverage file: %s", coverage_file)
-
+        vcover_cmd = [join(self._prefix, 'vcover'), 'merge', '-inputs',  join(self._output_path, 'coverage_files.txt')] + args + [file_name]
```
kraigher: use coverage_files constant from the line below instead of duplicating file name creation

vunit/modelsim_interface.py
```diff
@@ -339,13 +339,15 @@ def merge_coverage(self, file_name, args=None):
...
-                LOGGER.warning("Missing coverage file: %s", coverage_file)
+        coverage_files = join(self._output_path, 'coverage_files.txt')
+        vcover_cmd = [join(self._prefix, 'vcover'), 'merge', '-inputs'] + [coverage_files] + args + [file_name]
+        with open(coverage_files, "w") as fptr:
+            for coverage_file in self._coverage_files:
+                if file_exists(coverage_file):
+                    fptr.write(str(coverage_file) + "\n")
+                else:
+                    LOGGER.warning("Missing coverage file: %s", coverage_file)
+        fptr.close()
```
kraigher: The with-statement above is guaranteed to close the file so this is not necessary.

Issue #517, opened 2019-06-13
Mocking: check_only_log with any message


ttobsen:
In some cases, I just want to check if a log message occures with a given `log_level_t`, but I don't want to verify the content of the message itself. For exammple I just want to check that a error has raised, but I don't care what the content of the error message was.

I'm not sure if this is currently possible, because the `check_only_log` requires the `msg` parameter. If not, maybe it makes sense to `overload check_log` and `check_only_log` with dropping the `msg` parameter?

The main use case for me is, that I want to check that an error has raised, which has a pretty complex message. I'm not interested at all, if the single chars are as expected, I just want to ensure that the error raised.

kraigher:
It is not possible at the moment but it would be a good feature to have. Maybe by adding a default msg to the check procedures with the semantic that the message is not checked. This is the way it works with the log time field.

ttobsen:
Alright, but there is ine thing why I would suggest to overload the function:

The `msg` parameter is not at the end of the list where the default parameters start. The `msg` and the `log_level` would have to be twisted, which would break compatibility for existing test environments. If not twisting, the function call would look like 
`check_log(my_logger, msg => open, error)`

If you don't mind, I would open a PR and maybe you can help me a little bit?

kraigher:
ok go ahead

LarsAsplund:
A common mistake (at least for me) when verifying a verification component is that I write a test case which I think will trigger a specific error check but instead it fails too early on another error check. Without checking the message there is a risk that you would fail to discover that mistake. Providing a procedure allowing you to ignore the message may cause people to take a bad decision an go for the "lazy" solution. Another approach that would mitigate that risk is to supply a function allowing you to get the log object and then you can do whatever you like with its contents. For example you can assure that your complex error message at least contains a known static substring using `find` from the `string_ops` package. Makes your test a bit more solid. It's not a one-liner but it's a solution that provides some "protection" for the most common use case.

ttobsen:
I understand your doubts. The problem is, that VHDL doesn't provide complex string operations. The best and cleanest option would be, to have something like regular expressions (or at least wildcard support for * and ?). But this is a lot of work to implement.

But having a "lazy" option at least with support of checking a substring using the `find` function, could be a good compromise as a first iteration.

LarsAsplund:
> But having a "lazy" option at least with support of checking a substring using the find function, could be a good compromise as a first iteration.

[...]

PR #518, opened 2019-06-15
WIP: Add option to check log message by finding a substring

ttobsen:
This is just a first iteration for the implementation which is discussed in #517.

The implementation is working, I simply verified it using:

```vhdl
	logging_proc : process
		variable my_logger : logger_t := get_logger("My Logger");
	begin

		mock(my_logger, error);
		
		error(my_logger, "Fifo is full now, let's read data back.");		
		check_log(my_logger, "", error, msg_is_substr => true);
		
		error(my_logger, "Fifo is full now.");		
		check_log(my_logger, "Fifo is full now.", error, msg_is_substr => false);
		
		unmock(my_logger);

		test_runner_cleanup(runner);

	end process;
```

I'm not very deep in the Vunit VHDL sources, maybe there are string helper functions which could help to improve the code (which is a bit ugly in my opinion). But this first iteration should be good enough for having further discussions.

### DIFF
--- a/vunit/vhdl/logging/src/logger_pkg-body.vhd
+++ b/vunit/vhdl/logging/src/logger_pkg-body.vhd
@@ -798,17 +798,43 @@ package body logger_pkg is
     set(p_mock_queue_length, 0, get(p_mock_queue_length, 0) - 1);
     return make_string(got_logger_name, got_msg, got_level, got_log_time, got_line_num, got_file_name, check_time);
   end;
+  
+  impure function pop_log_item_with_substring(check_time : boolean; substr : string) return string is
+    constant SUBSTR_PREFIX : string := "Substring: ";
+    constant got_logger_name : string := pop_string(mock_queue);
+    constant got_level : log_level_t := log_level_t'val(pop_byte(mock_queue));
+    constant got_msg : string := pop_string(mock_queue);
+    constant got_log_time : time := pop_time(mock_queue);
+    constant got_line_num : natural := pop_integer(mock_queue);
+    constant got_file_name : string := pop_string(mock_queue);
+	variable found_str : string(1 to SUBSTR_PREFIX'length + substr'length) := SUBSTR_PREFIX & substr;
+  begin
+    set(p_mock_queue_length, 0, get(p_mock_queue_length, 0) - 1);
+    
+    if find(got_msg, substr) = 0 then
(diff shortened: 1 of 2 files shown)


vunit/vhdl/logging/src/logger_pkg-body.vhd
```diff
@@ -799,40 +799,91 @@ package body logger_pkg is
...
+    return make_string(got_logger_name, found_str, got_level, got_log_time, got_line_num, got_file_name, check_time);
+  end;
+
   procedure check_log(logger : logger_t;
                       msg : string;
                       log_level : log_level_t;
                       log_time : time := no_time_check;
                       line_num : natural := 0;
-                      file_name : string := "") is
+                      file_name : string := "";
```
LarsAsplund: A better approach is to completely disconnect from the existing `check_log` API and create a new procedure `pop_log_item(logger : logger_t) return log_item_t` and then provide helper functions to get the properties out of the `log_item_t` so the user doesn't become dependent on the implementation of that type. Now we don't have to change anything if `string_ops` gets better reg exp support and someone just caring about another subset of the properties, for example the message but not the level, could also use this approach.

ttobsen: The question is, do want to use the current implementation with find() method as discussed in #517 as a first iteration? I'm 100% for refactoring using your mentioned approach, but I'm not sure if I find some time in the next weeks or months. :disappointed:

LarsAsplund: I think it would be better to wait. I don't mind refactoring as long as it is behind the API. If we agree on my suggested solution we would have to revert the current API in the future and cause potential backward compatibility issues with people that already started to use it,

LarsAsplund: @ttobsen Any progress on this?

ttobsen: Btw: I think i will start a new MR with the second approach using helper functions. Are there any API changes planned which could have an effect of this implementation?

LarsAsplund: No, this is not urgent. I was just checking status on som of the MRs. There are no planned API changes in this area.

PR #520, opened 2019-06-17
WIP: Added VC template

LarsAsplund:
I've started to work on a VC template. The purpose is to:

* Set a number of rules that a VC/VCI should follow to become a good citizen in the ecosystem
* Promote naming guidelines for a common look and feel
* Create test to verify compliance

So far I've just started but please contribute with you own ideas.


1138-4EB:
I think that this conflicts with #511, in the sense that [vc_uart](https://github.com/VUnit/vc_uart) or [vc_axi](https://github.com/VUnit/vc_axi) are expected to be references/templates for anyone willing to contribute with a third-party VC.

Hence, I think that this PR could be a new repo (say `VUnit/vc_foo_bar`), which would be the reference and only the reference, instead of vc_uart and vc_axi which will be actual usable VCs. The only requirement so far is to use subdir `examples` for examples, `src` for sources, and to add a minimal `vunit_cfg.json`. But this is being discussed yet.

What do you think?

LarsAsplund:
I think in the long run it will be good to separate what defines a good VC from the actual VCs since the template VC project will have another focus. Good examples are important but there is also a question about providing a minimum template, rationale for design patterns, tutorials, compliance tests etc. I also think the template is owned by the core project since what's good design partly relates to what the core provides in terms of VC functionality. Ideally, I would like the VUnit core to provide a compliance test that could evaluate a VC given the repo root. Initially the compliance test can be extremely simple but it will give us a good starting platform:
- Anyone can evaluate a VC and suggest or contribute to a project to make it improve
- Our web site can not only provide links to projects but also maintain an up-to-date "score"
- We can provide badges related to that score
- By making the compliance test based on the repo root it may even be possible to test if the repo is using the badge and exposing their score.
- ...

If we're are in control of the rules for the "game" we can design it such that people strive for high quality without us having to actively enforcing it.

So a new repo is probably good but maybe the capability to run the compliance test should be part of the core and present in any installation.

1138-4EB:
> Good examples are important but there is also a question about providing a minimum template, rationale for design patterns, tutorials, compliance tests etc.

I thought that all these would fit into the 'example' VC:

```
- docs
- examples
  - test_examples.py
  - example_a
  - example_b
  - ...
- src
  - vc_*_context.vhd
  - *_pkg.vhd
  - ...
- test
  - tb_*.vhd
- .gitattributes
- .gitignore
- .travis.yml
- run.py
- tox.ini
- vunit_cfg.json
```

The minimum template would fit in `src`, tutorials in `examples` and compliance tests in `test`. The rationale for design patterns can be contained in `docs` (`*.rst` or `*.md`) and/or in `src` (comments in sources). Note that:

- The 'examples' subdir can be automatically integrated into the website by extending #516. From the perspective of the user, all the examples (the ones in the main repo, and those in all the 'supported/known' VC repos are listed together -separated in subsections-).
- The docs can contain a 'dummy' configuration file and index, so that all the content can be integrated in VUnit's site. But, it is possible to publish exactly the same content to it's own site at the same time. Maybe, we should define a subdir inside `docs` which is to be integrated in the main web site, while other documentation sources are expected to be independent.
- `.travis.yml`, `tox.ini`, `run.py` and `test_examples.py` provide the base for a consistent testing setup. I.e., if desired, VUnit that run the test suited.

> I also think the template is owned by the core project since what's good design partly relates to what the core provides in terms of VC functionality.

I think that 'owning' this part of the project does not imply it being part of the main repo. 'Own' relates to licensing and permissions, not not location.

Overall, I am ok with keeping some parts or everything in the main repo, that's your decission. But thinking on the UX of a potential contributor willing to create a new VC in a separate repo, I think it is easier to clone a dummy/reference repo and focus on the documentation/examples/descriptions in it; instead of having to cherry-pick the relevant sources from the main repo.

> Ideally, I would like the VUnit core to provide a compliance test that could evaluate a VC given the repo root.

> So a new repo is probably good but maybe the capability to run the compliance test should be part of the core and present in any installation.

Yes, I think that these compliance tests should be part of the main repo. It would allow third-party VC repos to use them on their own, in order to evaluate their status. It would also allow VUnit to test the compliance with no regard to the tests/examples available in the third-party repo.

> Our web site can not only provide links to projects but also maintain an up-to-date "score"

The 'logic' to extract examples (list and docstrings) and the list of VCs provided by each VC repo is ready: https://github.com/1138-4EB/vunit/blob/site-rework/tools/docs_utils.py#L100-L210 It would be straighforward to extract other metadata, or to 'generate' it.

There is no PR related to this yet, becase #516 and #511 should be merged first. BTW, could you please have a look at #516? I think it is interesting by itself, independently of all these VC-related changes.

> We can provide badges related to that score

I know how to dynamically generate badges for the docs (Sphinx). However, I don't know how third-party repos can integrate them in their README's and have them updated automatically. We could probably serve a JSON files and use the 'Dynamic' option in https://shields.io/; but I have never tried it.

> By making the compliance test based on the repo root it may even be possible to test if the repo is using the badge and exposing their score.

I don't understand why we would want to test this. As long as we can test the repo and provide it a score/badge in VUnit's web site, we should not care about it being shown in the README also, should we? Is this some kind of 'social rating'?

LarsAsplund:
> I think that 'owning' this part of the project does not imply it being part of the main repo. 'Own' relates to licensing and permissions, not not location.

I'm all for a separate repo. By ownership I was more referring to responsibility. The responsibility for maintaining best practices should not be with a specific VC. UART or any other VC will never care about all VC use cases nor use all of the features that VUnit provides for VC building.

> Yes, I think that these compliance tests should be part of the main repo. It would allow third-party VC repos to use them on their own, in order to evaluate their status. It would also allow VUnit to test the compliance with no regard to the tests/examples available in the third-party repo.

Exactly.

> I don't understand why we would want to test this. As long as we can test the repo and provide it a score/badge in VUnit's web site, we should not care about it being shown in the README also, should we? Is this some kind of 'social rating'?

I'm basically just pointing out that with a VUnit core owned compliance test based on the VC repo root we have a platform for creating whatever means we think necessary to create a high-quality and interoperable ecosystem. Not really saying that all of them are great. The first step is to have a "rating" system. To start with that would be a pass/fail compliance test. The second step is to make that rating as visible as possible. The more visible a score is the more motivation to keep it high. Personally I care a lot more about the quality badges on vunit.github.io than what other ranking sites may say about the project. I'm thinking that a VC maintainer would care more about the quality metrics presented in her repo than the ones presented on vunit.github.io. Enforcing a badge in the README might be a bit too much though...

1138-4EB:
> UART or any other VC will never care about all VC use cases nor use all of the features that VUnit provides for VC building.

As a matter of fact, there is one example for UART, two for AXI and none for Avalon or Wishbone. The number of tests is also quite different. None of them has any specific documentation.

Regarding the rating/badges, gotcha, and I agree. I can try a first iteration by creating a table with three fields per VC: 'tests pass', 'has examples' and 'has docs'. The rating might be:

- tests + egs + docs success
- tests [+ egs] [+ docs]: important
- !tests: critical

Examples, if exist, must also pass/run.

I'd like to know what are your ideas for integrating these new contents (table, per VC docs, per VC examples, etc.) in the web site. Do you have any plan to (re)organize the documentation about verification components?

As you can see in https://1138-4eb.github.io/vunit/, I moved 'Examples' out of the 'User Guide', but VC are still in Documentation > VHDL Libraries > Verification Components. Furthermore, since the sidebar is different, the number of clicks to reach from the home to the Verification Components is reduced from 3 to 2. This might sound stupid, by I find myself having to browse that section quite frequently, and it annoys me.

LarsAsplund:
My first priority right now would be to have some basic use cases in the template. Blocking/non-blocking transactions and transactions with or without a reply. Based on our discussions I think the next step would be to create a template repository and then the mechanism allowing such a repo to be evaluated from a VUnit core installation

1138-4EB:
> Based on our discussions I think the next step would be to create a template repository and then the mechanism allowing such a repo to be evaluated from a VUnit core installation

Since vc_axi and vc_uart are created already, I was thinking about doing the mechanism to evaluate them, while the template is being completed. I believe that the 'evaluation' will be based on running some python tests. Therefore, from the point of view of the mechanism involving Git + Travis + Sphinx + Pages, the specific names of the tests do not matter.

LarsAsplund:
> I believe that the 'evaluation' will be based on running some python tests.

Yes, all I need now is functionality for generating a compliance test suite for a specific VC and then create a `run.py` for that

bradleyharden:
Minor addition, but I think there should also be functions to retrieve the actor, logger, etc. from the VC. Should they be named `actor`, `logger`, etc. or `as_actor`, `get_logger`, etc.?

1138-4EB:
Ref: https://help.github.com/en/articles/creating-a-template-repository

vunit/ui/library.py
```diff
@@ -290,11 +291,25 @@ def package(self, name):
 
         return PackageFacade(self._parent, self._library_name, name, design_unit)
 
-    def entity(self, name):
+    def get_entity(self, name):
         """
         Get an entity within the library
+        """
+        library = self._project.get_library(self._library_name)
```
1138-4EB: Maybe lines 298-302 can be moved to a _helper function to avoid duplication (284-288)?

vunit/ui/package_facade.py
```diff
@@ -22,6 +22,7 @@ def __init__(self, parent, library_name, package_name, design_unit):
         self._library_name = library_name
         self._package_name = package_name
         self._design_unit = design_unit
+        self._source_file = self._design_unit.source_file
```
1138-4EB: Since `_source_file` is accessible through `source_file()`, is it really required to define it apart from `_design_unit`?

vunit/vc/compliance_test.py
```diff
@@ -0,0 +1,339 @@
...
+            vc_instantiation = (
+                vc_instantiation[:-2]
+                + """
+    port map(
+"""
+            )
+
+            vc_instantiation += port_mappings[:-2] + "\n    );\n"
+
+        tb_template = Template(
```
1138-4EB: Instead of 'hardcoding' the template in the Python source, would it be feasible to define it in a `*.vhd.tpl` template file?

vunit/__init__.py
```diff
@@ -14,6 +14,8 @@
 from vunit.vunit_cli import VUnitCLI
 from vunit.about import version, doc
 from vunit.json4vhdl import read_json, encode_json
+from vunit.vc.verification_component import VerificationComponent
```
eine: Maybe `from vunit.vc import VerificationComponent, VerificationComponentInterface`?

vunit/vc/__init__.py
```diff
@@ -0,0 +1,9 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# Copyright (c) 2014-2020, Lars Asplund lars.anders.asplund@gmail.com
+
+"""
+Code related to VUnit verification components.
+"""
```
eine: Maybe import the "public classes" here?

```py
from vunit.vc.verification_component import VerificationComponent
from vunit.vc.verification_component_interface import VerificationComponentInterface
```

eine:
Apart from the in-code comments above, some other preliminary points:

- Can we merge the rules about having a single package/entity in a single file?
- Can we have the rules numbered or assigned any other identifier?
- I think it would be cleaner to define the templates in `test_compliance_test.py` as separate `*.vhd.tpl` files.
- We should avoid to use `dirname`, `join`, etc. in new code, and use `from pathlib import Path` instead.

LarsAsplund:
> Apart from the in-code comments above, some other preliminary points:
> 
> * Can we merge the rules about having a single package/entity in a single file?

Yes-

> * Can we have the rules numbered or assigned any other identifier?

Yes. I think I'll move all that to the user guide as a start for the documentation

> * I think it would be cleaner to define the templates in `test_compliance_test.py` as separate `*.vhd.tpl` files.

I'll try to move it out of the way somewhere

> * We should avoid to use `dirname`, `join`, etc. in new code, and use `from pathlib import Path` instead.

I guess it's time to abandon old habits

LarsAsplund:
> Minor addition, but I think there should also be functions to retrieve the actor, logger, etc. from the VC. Should they be named `actor`, `logger`, etc. or `as_actor`, `get_logger`, etc.?

@bradleyharden Right now I've bundled all standard configuration items in a record and I require that every VC can provide that when called with `get_std_cfg`. If that is in place you can reuce the getters. I.e. `get_logger(get_std_cfg(handle))`. A VC designer could of course provide wrappers as  a convenience but I'm thinking that a standard configuration would allow us to make (minor) updates that would be available to all VCs without any updates on the VC side.

vunit/vc/compliance_test.py
LarsAsplund: I agree that these code snippets are big and few enough to move elsewhere. However, I think your suggestion to do the same thing for `test_compliance__test.py` would result in too fragmented code. Those snippets are very small and moving them elsewhere would make the tests hard to read.

LarsAsplund:
I've pushed an updated version to address the comments I received.

LarsAsplund:
Since we're doing some backward compatibility breaking changes it would be good the hear what some of the other contributors have to say. @slaweksiluk @eschmidscs @dstadelm

eschmidscs:
@LarsAsplund Sorry, didn't see your comment until now (actually @dstadelm had to point it out).

I currently don't grasp the full change. What you discuss here sounds feasible to me - and is yours to decide in the end.

Concerning backwards compatibility: It seems to me that this is not a major blocker, certainly some fixes. Am I wrong?
Can I just merge this PR locally into master and check what happens?

I'm generally happy if I know ahead that something will break instead of figuring out that something broke (as has happened with the non-blocking axi_stream check). It is normally not too hard to fix once you have a time slot available.

eine:
@eschmidscs, this PR was just rebased and brought up to date. Please, check it out, and run the compliance tests with your simulators, and/or use your usual designs/tests. Report any issue you find. Moreover, if you want to actually review the content, that will be welcome too. If any of your PRs is affected by this one, you might want to rebase them on top. Yet, beware that this might be further rebased, so I suggest you keep a backup of yours.

vunit/vc/compliance_test.py
```diff
@@ -0,0 +1,139 @@
...
+        template_code,
+        vci_name,
+    ) = VerificationComponentInterface.create_vhdl_testbench_template(
+        args.vc_lib_name, args.vci_path, args.vc_handle_t
+    )
+    if not template_code or not vci_name:
+        sys.exit(1)
+
+    if not args.output_path:
+        output_dir = args.vc_path.parent / ".vc" / vci_name
```
eschmidscs, line 55: this line currently fails. it should probably be args.vci_path.parent.

vunit/vc/compliance_test.py
```diff
@@ -0,0 +1,139 @@
...
+        if not output_dir.exists():
+            output_dir.mkdir(parents=True)
+
+        output_path = output_dir / ("tb_%s_compliance_template.vhd" % args.vc_handle_t)
+    elif args.output_path.is_dir():
+        output_dir = args.output_path / vci_name
+        if not output_dir.exists():
+            output_dir.exists(parents=True)
```
eschmidscs, line 63: this line currently fails. it should probably be mkdir instead of exists.

eschmidscs:
It would be helpful if the compliance_test.py would add a correct copyright header.

eschmidscs:
I am trying to get #703 to run with the compliance checks. Given the issues already noted, it seems I am getting somewhere.

The show stopper right now is the _p_initial_ that I added to the handle, currently an unconstrained std_logic_vector.
It is unconstrained in analogy to the std_logic_checker where the value port is unconstrained, thus the std_logic_driver should have it unconstrained as well, resulting in _p_initial_ being unconstrained.

The unconstrained vector cannot work with the compliance testbenches, because they don't constrain the handle.
The easy fix is to add a length parameter to the handle and have all vectors constrained. This is ok for me.
But maybe you want to have unconstrained vectors working?

LarsAsplund:
@eschmidscs I will have a test run of the compliance tests on your signal driver to see what conclusions I can make.

LarsAsplund:
@eschmidscs I updated to support unconstrained handle fields. Please rebase and try with #703 again

eschmidscs:
@LarsAsplund Ok, done this. Seems to work fine. Maybe you check #703 if the result is what you wanted.

> It would be helpful if the compliance_test.py would add a correct copyright header.

This would still be handy. I kept running iterations of calling the script, patching back my changes and running the testbenches. If the copyright would already be there, the patch work would be reduced by 50%.

LarsAsplund:
@eschmidscs The reason that no header is added is that the plan is for people to create independent repos for their VCs which are maintained by them under their license of choice. What we could do is to add the ability to specify a text file containing the header and then have the VUnit header as the default if no file is specified

eschmidscs:
@LarsAsplund Ok, that's a reason. Could you go fancy and add both the additional file you propose and, if only the flag but no file is specified, just insert your standard header?

LarsAsplund:
@eschmidscs That should be possible. We could also interpret the text file as an f-string. That would allow access to some convenient strings. For example,

```
{vc_header.vunit_license}
--
-- Copyright (c) {vc_header.year} John Doe, john.doe@gmail.com
```

would result in

> -- This Source Code Form is subject to the terms of the Mozilla Public
-- License, v. 2.0. If a copy of the MPL was not distributed with this file,
-- You can obtain one at http://mozilla.org/MPL/2.0/.
--
-- Copyright (c) 2021, John Doe john.doe@gmail.com

Or maybe a standard header function.

```
{vc_standard_header(name="John Doe", mail="john.doe@gmail.com")}
```

That would give the same output as above. There would also be optional arguments to specify license and year.

eschmidscs:
@LarsAsplund That looks nice! I like the f-string idea.

PR #530, opened 2019-08-21
Improve `remove_comments` function in the VHDL parser

### LINKED ISSUE #529: Two don't cares ('-') adjacent in a literal is parsed as a comment
The `remove_comments(code)` function is overly simplistic. It matches any two adjacent `-` characters until the end of the line. `'-'` is also used as a "don't care" value for `std_logic`. If there is an `std_logic_vector` literal where there are two don't care values in a row (e.g. `'----'`), it is parsed as a comment and the rest of the line is removed, causing the parse to fail.

https://github.com/VUnit/vunit/blob/12be97328976d592c8953e4943887432c3ec695f/vunit/vhdl_parser.py#L991-L1007

ktbarrett:
Changed regex for matching comments to not match when the `--` pattern is found in a string. This fixes #529.

### DIFF
--- a/vunit/vhdl_parser.py
+++ b/vunit/vhdl_parser.py
@@ -988,7 +988,7 @@ class VHDLReference(object):
         return self.name_within == "all"
 
 
-VHDL_REMOVE_COMMENT_RE = re.compile(r'^([^\"\-]*(?:\"[^\"]*\"[^\"\-]*)*)--.*')
+VHDL_REMOVE_COMMENT_RE = re.compile(r'^([^\"\-]*(?:\"(?:\"\"|[^\"])*\"[^\"\-]*)*)--.*', re.MULTILINE)
 
 
 def _comment_repl(match):
--- a/vunit/vhdl_parser.py
+++ b/vunit/vhdl_parser.py
@@ -996,7 +996,7 @@ def _comment_repl(match):
     Replace comment with equal amount of whitespace to make
     lexical position unaffected
     """
-    return match.group(1) + " "*len(match.group(2))
+    return match.group(1) + (" " * len(match.group(2)))
 
 
 def remove_comments(code):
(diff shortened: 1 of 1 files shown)


vunit/vhdl_parser.py
```diff
@@ -988,16 +988,15 @@ def reference_all_names_within(self):
         return self.name_within == "all"
 
 
-VHDL_REMOVE_COMMENT_RE = re.compile(r'--[^\n]*')
+VHDL_REMOVE_COMMENT_RE = re.compile(r'^([^\"\-]*(?:\"[^\"]*\"[^\"\-]*)*)--.*')
```
go2sh:
```suggestion
VHDL_REMOVE_COMMENT_RE = re.compile(r'^([^\"\-]*(?:\"(?:\"\"|[^\"])*\"[^\"\-]*)*)--.*')
```
A correct implementation would look like this, since VHDL allows to escape the `"` inside a string literal with `""`. So `""""` means `"`.

go2sh:
An the test break.

ktbarrett:
The regex removes newlines and the failing tests assert that certain things are found on certain lines. If that's important behavior (is it really?) the regex will also have to be modified to stop on newlines. Doing this from the WebUI so bear with me.

ktbarrett:
The tests that were previously failing pass now, however the GHDL docker test is broken (unresponsive and times out).

1138-4EB:
> The tests that were previously failing pass now, however the GHDL docker test is broken (unresponsive and times out).

Docker tests are broken for this PR only. Hence, it is related to the modifications you introduced. Just to be sure about it, the three failing tests were manually restarted, and the three of them failed again. Moreover, other commits/PRs have been tested before and after this PR, and all of them passed.

ktbarrett:
Running it locally is making my laptop hot. There must be some input that regex blows up on...  Oh well.

PR #540, opened 2019-09-16
Speedup adding existing files

dstadelm:


1138-4EB:
Hi @dstadelm! It seems that linting issues are easy to fix (remove a blank line and add a docstring). Nonetheless, would might adding a brief description about what you are trying to achieve? The title mentions *speedup*. It'd be good to know what performance gain you expect.

PR #543, opened 2019-09-17
style: merge data_types/src/types/* into data_types/src/types.vhd

umarcor:
Currently, there are two files with 5 lines each. When byte_vector is added in #507, there will be three. Apart from being short, the content is tightly related. As proposed in [#507](https://github.com/VUnit/vunit/pull/507#discussion_r323873269), a single `types_pkg` package is defined, and all the types are put there.

Moreover, as commented in [#507](https://github.com/VUnit/vunit/pull/507#discussion_r323408031), the definition of `byte_t` is moved from `verification_components/**/memory_pkg.vhd` to this `data_types/**/types.vhd`.

---

Also, in #507, `type storage_mode_t is (internal, extfnc, extacc);` is defined in `string_ptr_pkg.vhd`. This type is to be later used by `integer_vector_ptr_pkg.vhd` (#476) and `memory_pkg.vhd` (#470). Therefore, it makes sense to make `byte_t` and `storage_mode_t` available in a similar fashion.


umarcor:
@LarsAsplund, I updated this issue and I believe it is ready to merge. However, there is something strange going on with Travis.

- The latest commit in this PR is https://github.com/VUnit/vunit/pull/543/commits/fd89519320dea9e7e6b1089fb05c96ad83908d36.
- The corresponding Travis run on my fork is https://travis-ci.com/dbhi/vunit/builds/128416931, which is all successful.
- The commit is automatically merged with master, producing https://github.com/VUnit/vunit/commit/32a89e24bba4cf704305a946a34334bf96851ebc.
- The corresponding Travis run (which is shown below) is https://travis-ci.org/VUnit/vunit/builds/587195514, and it is failing just as the latest commit pushed by @kraigher: https://travis-ci.org/VUnit/vunit/builds/587167948.

This is very strange because this PR is not rebased on top of master yet. Therefore, even if the latest commit in master is broken, this PR should succeed (as it does on my fork).

vunit/vhdl/verification_components/src/memory_pkg.vhd
```diff
@@ -9,12 +9,15 @@
...
 
-use work.integer_vector_ptr_pkg.all;
+use work.types_pkg.all;
 use work.string_ptr_pkg.all;
+use work.integer_vector_ptr_pkg.all;
 use work.logger_pkg.all;
 
 package memory_pkg is
 
+  alias byte_t is byte_t;
```
LarsAsplund: Is this really needed?

umarcor: I think so. Otherwise, designs that use `memory` need to use `types_pkg` explicitly. For example, `tb_memory.vhd`: https://travis-ci.com/dbhi/vunit

LarsAsplund: I think we can expect users of memory_pkg to also include common packages on which it depends. Otherwise we would also have to alias things like `std_logic_vector`. If the burden on the user to include all the necessary packages becomes too much we should provide contexts to ease that burden.

umarcor: Ok. The alias is removed now.

PR #547, opened 2019-09-21
Vhdl 2019

LarsAsplund:
Enable VHDL-2019 for Riviera-PRO


vunit/activehdl_interface.py
```diff
@@ -104,13 +104,29 @@ def compile_source_file_command(self, source_file):
         LOGGER.error("Unknown file type: %s", source_file.file_type)
         raise CompileError
 
+    @staticmethod
+    def _std_str(vhdl_standard):
+        """
+        Convert standard to format of Active-HDL command line flag
+        """
+        if vhdl_standard == "2008":
```
1138-4EB:
```py
        if vhdl_standard in ['93', '2002', '2008']:
            return "-%s" % vhdl_standard
        raise ValueError("Invalid VHDL standard %s" % vhdl_standard)
```

vunit/builtins.py
```diff
@@ -44,13 +45,15 @@ def _add_files(self, pattern):
         """
         Add files with naming convention to indicate which standard is supported
         """
-        supports_context = self._simulator_class.supports_vhdl_2008_contexts() and self._vhdl_standard == "2008"
+        supports_context = self._simulator_class.supports_vhdl_contexts() and self._vhdl_standard in ["2008", "2019"]
```
1138-4EB: Instead of having `in ["2008", "2019"]` or `not in ["2008", "2019"]`, can we find some *cleaner* solution to *group* them?

- `self._vhdl_standard >= "2008"`. I honestly don't know whether this is feasible.
- `self._vhdl_standard_tag in ["2008p"]`. Alternatively, by adding a helper `check_vhdl_standard_tag(required_min_tag)` function: `check_vhdl_standard_tag("2008p")`.

vunit/modelsim_interface.py
```diff
@@ -142,13 +142,29 @@ def compile_source_file_command(self, source_file):
         LOGGER.error("Unknown file type: %s", source_file.file_type)
         raise CompileError
 
+    @staticmethod
+    def _std_str(vhdl_standard):
+        """
+        Convert standard to format of Modelsim command line flag
+        """
+        if vhdl_standard == "2008":
```
1138-4EB: See suggestion [above](https://github.com/VUnit/vunit/pull/547/files#r327522796).

vunit/rivierapro_interface.py
```diff
@@ -135,13 +135,33 @@ def compile_source_file_command(self, source_file):
         LOGGER.error("Unknown file type: %s", source_file.file_type)
         raise CompileError
 
+    @staticmethod
+    def _std_str(vhdl_standard):
+        """
+        Convert standard to format of Riviera-PRO command line flag
+        """
+        if vhdl_standard == "2019":
```
1138-4EB: Same suggestion as [above](https://github.com/VUnit/vunit/pull/547/files#r327522796). However, an exception is required here, since `2019` is converted to `-2018`.

vunit/vhdl_parser.py
```diff
@@ -43,6 +43,7 @@ class VHDLDesignFile(object):  # pylint: disable=too-many-instance-attributes
     """
     Contains VHDL objects found within a file
     """
+
```
1138-4EB: Are these spaces additions required? If so, it might be worth making a commit with these only.

LarsAsplund: This is the result of some code formatting settings I have. While we don't have any such rules I think both autopep8 and pydev's formatter will add this line so I guess we can call this good practice. I think we should consider a standard formatter. Meanwhile we could apply the boy scout rule

vunit/builtins.py
LarsAsplund: I did consider it for a while but postponed it for the future. Do you think we need to fix it before the release?

1138-4EB: No, I don't think we need to fix this before the release. I just tried to *visualize* how to implement such enhancement, and I wasn't sure about it.

Related to the comments about standard names last week, I think we should have all of the standard names, aliases and these *tags* defined in a single place. Then, when `from_argv` or `main` are executed, functions should check them ASAP, so that all other internal steps can rely on some fixed/known values. Functions that need to know the version, can check the *parsed version*, and functions that need to know the tag can check the *parsed tag*.

vunit/vhdl_parser.py
1138-4EB: Your are correct. I executed `autopep8 vunit/vhdl_parser.py` and it introduced these spaces, among other changes.

I think it'd be good to have any standard formatter. Being able to execute it either in a container or as a git hook would be really helpful.

vunit/builtins.py
kraigher: I think we should have a vhdl_standard.py file defining a type for the vhdl standard with <= operators and similar instead of using strings. When adding support for VHDL 2019 it would have been a good opportunity to do it.

kraigher:
I pushed a commit that adds a new type for VHDL standard that is used everywhere internally. Only in ui.py methods is a string allowed.

kraigher:
And 1993 is also a supported value now, as well as 02, 08, 19

vunit/builtins.py
```diff
@@ -78,35 +77,35 @@ def _add_array_util(self):
         """
         Add array utility
         """
-        if self._vhdl_standard != '2008':
-            raise RuntimeError("Array util only supports vhdl 2008")
+        if self._vhdl_standard >= VHDL.STD_2008:
```
1138-4EB: Shouldn't it be `<=` instead of `>=`?

vunit/builtins.py
```diff
@@ -78,35 +77,35 @@ def _add_array_util(self):
...
+            raise RuntimeError("Array util only supports vhdl 2008 and later")
```
1138-4EB: - `and later`
- `or later`
- `and/or later`

vunit/test/unit/test_vhdl_standard.py
```diff
@@ -0,0 +1,65 @@
...
+def test_error_on_invalid_standard():
+    _assert_is_invalid("2001")
+    _assert_is_invalid("002")
+    _assert_is_invalid("993")
+    _assert_is_invalid("2")
+    _assert_is_invalid("3")
+
+
+def test_equality():
+    assert VHDL.standard("2008") == VHDL.standard("2008")
```
1138-4EB: Maybe check `VHDL.standard("2008") == VHDL.standard("08")` instead?

kraigher: There is such a check for 1993.

vunit/builtins.py
kraigher: yes the logic is inverted.

vunit/test/unit/test_vhdl_standard.py
1138-4EB: Yes. What I meant is that `x == x` will always be true, independently of what `x` is. The only possibility I can imagine for it to fails is if calls `VHDL.standard` where sensitive to some runtime param (system time, number of times the func was executec, etc.). Is this what we are trying to check here?

kraigher: X1 == X2 for a class without __eq__ python compares the object instance id returned by id(X1) and id(X2) which is not equal in the case you commented since the rhs and lhs are different object instances even though they got the same __init__ argument.

PR #548, opened 2019-09-23
Compile only required subset

dstadelm:
This introduces a new command line switch -m / --minimal 
Using this switch will allow to only compile the files that are required for running a given test bench, or a set of test benches, or a set of filtered test benches.

### DIFF
--- a/vunit/project.py
+++ b/vunit/project.py
@@ -9,24 +9,18 @@
 """
 Functionality to represent and operate on a HDL code project
 """
-
-
-from os.path import join, basename, dirname, splitext, isdir, exists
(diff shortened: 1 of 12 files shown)


kraigher:
Is it a lot of work for you to make a separate PR for the refactoring and a PR adding the new feature? It would make reviewing the code easier and I can approve one change at a time.

dstadelm:
Sure thing, no problem at all.

FYI
I had to refactor the SourceFile, VerilogSourceFile and VHDLSourceFile to a separate file as pycodecheck was complaining as the file project.py was getting too long.

PR #549, opened 2019-09-24
Refactoring for subset compile

dstadelm:
This refactores the functions get_files_in_compile_order and get_dependencies_in_compile_order in the first commit. This was done as quite a lot of code is duplicate code. 
In the second commit the SourceFile, VHDLSourceFile and VerilogSourceFile classes are refactored to a seperate module. As pycodestyle complained that the file project.py is too long.

### DIFF
--- a/vunit/project.py
+++ b/vunit/project.py
@@ -9,24 +9,18 @@
 """
 Functionality to represent and operate on a HDL code project
 """
-
-
-from os.path import join, basename, dirname, splitext, isdir, exists
-from copy import copy
-import traceback
+from os.path import join, basename, dirname, isdir, exists
 import logging
 from collections import OrderedDict
 from vunit.hashing import hash_string
 from vunit.dependency_graph import (DependencyGraph,
                                     CircularDependencyException)
-from vunit.vhdl_parser import VHDLParser, VHDLReference
-from vunit.cached import file_content_hash
+from vunit.vhdl_parser import VHDLParser
 from vunit.parsing.verilog.parser import VerilogParser
-from vunit.parsing.encodings import HDL_FILE_ENCODING
 from vunit.exceptions import CompileError
-from vunit.simulator_factory import SIMULATOR_FACTORY
-from vunit.design_unit import DesignUnit, VHDLDesignUnit, Entity, Module
 from vunit import ostools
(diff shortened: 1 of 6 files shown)


1138-4EB:
Hi @dstadelm! It seems that this PR is a subset of #550. Could you, please, make it explicit? I'd suggest adding `Ref #549` or `Ref #550` in both comments, and adding some sentence in #550 explaining that it should not be reviewed until this is merged. Alternatively, you can edit the title to number (`[1/2]`, `[2/2]`, etc.) a series of related PRs.

Regarding this PR, would you mind either elaborating your first comment or further splitting these commits? The main motivation seems to be to reduce duplicate code. However, the proposed implementation is 10% larger. It might be legitimate, but it is hard to see it in the current state. For example, it seems that 5 lines of code are split to a new `_get_affected_files` function. Should that be a commit on it's own, it'd be easier to see where is it called from and how it helps with reducing code duplication.

vunit/project.py
```diff
@@ -370,59 +370,74 @@ def _get_compile_timestamps(self, files):
                 timestamps[source_file] = ostools.get_modification_time(hash_file_name)
         return timestamps
 
-    def get_files_in_compile_order(self, incremental=True, dependency_graph=None):
+    def get_files_in_compile_order(self, incremental=True, dependency_graph=None, files=None):
         """
         Get a list of all files in compile order
-        incremental -- Only return files that need recompile if True
+        :param incremental: -- Only return files that need recompile if True
```
1138-4EB: Is `--` a typo?

Also, please, add descriptions for `dependency_graph` and `files` too.

dstadelm:
Shure thing. I'll create a new pull request.

Regards
Dave

1138-4EB:
> Shure thing. I'll create a new pull request.

It is up to you, of couse; but it is not required to close and open new PRs for these changes. You can just push commits or force-push this branch. The advantage of doing so is that it is easier for us to see previous comments/discussions. It is useful to see different versions of a contribution, because it helps understand which decisions were made.

dstadelm:
Ok, I didn't know about that, especially the force push. 

My new pull request does not comply with one pylint instance check. I don't know why, because it is complaining about code I didn't touch....

PR #551, opened 2019-09-24
refactoring for subset compile

dstadelm:
This pull request is the base for following pull request to compile_only_rquired_subset


1138-4EB:
Refs: #540, #541, #548, #549, #550.
Base for: #552.

> I do not understand why the checker fails. I've not changed the code where the checker fails....

Although some of the errors seem not to be related, others correspond to the files you introduced. See https://travis-ci.org/VUnit/vunit/jobs/588929398#L283

I suggest you to start from `master` and add/cherry-pick the commits one-by-one, running all the tests each time. It seems that the issues are related to `py37-lint`. Hence, you can run the corresponding test suite locally with `tox -e py37-lint`.

1138-4EB:
@dstadelm, it seems that something changed in the last 16h that broke our main branch. These tests were executed 16h ago, and all of them passed: https://travis-ci.org/VUnit/vunit/builds/588730072. I restarted `py37-lint` now, and, as you see, it is failing with similar messages as the ones you get.

@kraigher, @LarsAsplund, we need to investigate if linting criteria was updated this afternoon.

kraigher:
Most likely a new version of pylint is available that checks more stuff than previously. This happens from time to time. We just fix the new things it finds on master. I can push a fix and then this PR can be rebased.

1138-4EB:
> Most likely a new version of pylint is available that checks more stuff than previously. This happens from time to time. We just fix the new things it finds on master. I can push a fix and then this PR can be rebased.

Exactly! pylint was released 8h ago: https://github.com/PyCQA/pylint/releases Thanks for the hint.

I'm already having a look at it. So, if you don't mind, I'd like to push the fix. Unless you have already written it, of course.

kraigher:
I pushed a fix for the broken pylint to master now, once you see that the master build goes green you should rebase the pull request.

vunit/project.py
```diff
@@ -433,6 +434,20 @@ def comparison_key(source_file):
         sorted_files = sorted(affected_files, key=comparison_key)
         return sorted_files
 
+    def _get_affected_files(self, target_files, get_depend_func):
```
1138-4EB: It seems that lines 384-396 and 423-435 of the resulting `project.py` file are redundant. Does it make sense to move those snippets into this new function (maybe renamed to `_get_affected_files_in_compile_order`)?

vunit/project.py
```diff
@@ -448,6 +438,17 @@ def _get_affected_files(self, target_files, get_depend_func):
             self._handle_circular_dependency(exc)
             raise CompileError
 
+    def _get_compile_order(self, dependency_graph):
```
1138-4EB: See [previous comment](https://github.com/VUnit/vunit/pull/551#discussion_r327845879) about merging `_get_compile_order` and `_get_affected_files`.

vunit/project.py
```diff
@@ -518,6 +508,17 @@ def update(self, source_file):
         LOGGER.debug('Wrote %s content_hash=%s', source_file.name, new_content_hash)
 
 
+def _get_sorted_files(files, compile_order):
```
1138-4EB: See previous comments about merging `_get_compile_order`, `_get_affected_files` and `_get_sorted_files`.

vunit/project.py
```diff
@@ -370,15 +370,17 @@ def _get_compile_timestamps(self, files):
...
         param: incremental: Only return files that need recompile if True
         """
         if dependency_graph is None:
             dependency_graph = self.create_dependency_graph()
 
-        files = self.get_source_files_in_order()
+        if files is None:
+            files = self.get_source_files_in_order()
+
         files_to_recompile = self._get_files_to_recompile(files, dependency_graph, incremental)
```
1138-4EB: What about the following syntax?

```py
files_to_recompile = self._get_files_to_recompile(
    (files or self.get_source_files_in_order()),
    dependency_graph,
    incremental
)
```

vunit/project.py
dstadelm: Then I have to pass dependency_graph and dependeny_graph.dependent/dependencies or I write a closure function passing dependent or dependencies returning the function you propose. Do you have a preference?

1138-4EB: Although I might have a preferene, I think it is not relevant here. It was a somehow 
rhetorical question: are you going to later (in some other PR) use any of the three individual functions from somewhere else in the codebase and/or do you want to expose any of them? If the answer to any of these is yes, keeping them separated is preferred/required. However, should it be an *atomic* chain (external processes always execute A, then B, then C), the creation of three functions might be better implemented as three comments/sections in a single function. Nevertheless, we are not doing embedded programming here, so having three function calls is irrelevant from a performance perspective.

kraigher:
This PR is approved but there is a merge conflict preventing us from merging it.

vunit/project.py
dstadelm: Originally I did use the three functions individually. But now in my final solution I actually do not. So I would agree that it could be preferable to have one function. On the other hand the three functions actually have three 'responsibilities' and therefor I personally find it easier to understand when reading the code. 

I completely agree with you and find the discussion very interesting. My background is also more an embedded one, but here I was actually trying to follow Uncle Bob's advice from 'Clean Code'.

Alternatively it would also be possible to write one function which orchestrates (wraps) these three functions so that there is only one function call in get_files_in_compile_order and get_dependencies_in_compile_order. Again this is not ideal from performance perspective, but would keep the functions small which makes it easier to understand the intent.

1138-4EB: I believe that this discussion we had is enough for anyone considering to write the wrapper/orchestration function in the future. Thanks!

It is not required for now, and you/we would better focus on further features/contributions. Just resolve the conflicts, and we can move forward.

kraigher: Clean code is a good book to use as a guidance. Small function with a clearly defined responsibility is worth striving for.

dstadelm:
I've merged master and also made the changes suggested by @1138-4EB. The remaining subset of functions is required in the upcoming pull request for the minimal subset compilation.

PR #552, opened 2019-09-24
Compile only required subset

dstadelm:
This pull request is based on pull request #551 therefore it should only be reviewed when #551 has been merged


vunit/ui.py
```diff
@@ -1888,6 +1896,10 @@ def __init__(self, source_file, project, ui):
         self._project = project
         self._ui = ui
 
+    @property
+    def project_source_file(self):
```
kraigher: This exposes a private class on the public API. Use private access instead with a pylint disable comment.

vunit/simulator_interface.py
```diff
@@ -163,13 +163,14 @@ def add_simulator_specific(self, project):
         Hook for the simulator interface to add simulator specific things to the project
         """
 
-    def compile_project(self, project, printer=NO_COLOR_PRINTER, continue_on_error=False):
+    def compile_project(self, project, printer=NO_COLOR_PRINTER, continue_on_error=False, target_files=None):
```
kraigher: The new functionality misses unit tests in test_project.py

vunit/ui.py
```diff
@@ -1032,9 +1032,17 @@ def _compile(self, simulator_if):
         """
         Compile entire project
         """
+        # get test benches
+        if self._args.minimal:
```
kraigher: The new functionality misses some kind of test in test_ui.py

vunit/test_bench.py
```diff
@@ -281,10 +281,15 @@ class Test(object):
          Implicit tests are those when there are no tests in the test bench, just the test suite
     """
 
-    def __init__(self, name, location):
+    def __init__(self, name, location, file_name):
         self._name = name
         self._location = location
         self._attributes = []
+        self._file_name = file_name
```
kraigher: The location should already contain the file name.

dstadelm:
There are now some more tests. I haven't written one that explicitly works on __"def compile()"__ The only thing that would have to be tested would be that the switch is applied correctly. But I'm not sure how to achieve this...

vunit/simulator_interface.py
```diff
@@ -214,13 +215,19 @@ def __compile_source_file(self, source_file, printer):
 
         return True
 
-    def compile_source_files(self, project, printer=NO_COLOR_PRINTER, continue_on_error=False):
+    def compile_source_files(self, project, printer=NO_COLOR_PRINTER, continue_on_error=False, target_files=None):
         """
         Use compile_source_file_command to compile all source_files
+        param: target_files: Given a list of SourceFiles only these and dependent files are compiled
         """
         dependency_graph = project.create_dependency_graph()
```
kraigher: If this was contained in a single call to Project class there would not need to be an extra test of the SimulatorInterface class

vunit/test/unit/test_simulator_interface.py
```diff
@@ -53,6 +53,33 @@ def test_compile_source_files(self):
 """)
         self.assertEqual(project.get_files_in_compile_order(incremental=True), [])
 
+    def test_compile_source_files_minimal_subset(self):
```
kraigher: This is a lot of code just to test the if-statement in the SimulatorInterface class selecting the Project method to call. Maybe this test can be removed.

vunit/ui.py
```diff
@@ -1032,9 +1032,21 @@ def _compile(self, simulator_if):
...
+            target_files = self._get_testbench_files(simulator_if)
+        else:
+            target_files = None
+
         simulator_if.compile_project(self._project,
                                      continue_on_error=self._args.keep_compiling,
-                                     printer=self._printer)
+                                     printer=self._printer, target_files=target_files)
+
+    def _get_testbench_files(self, simulator_if):
```
kraigher: I think this function misses a docstring explaning what it does. Such as "Return the list of all files for the currently selected tests to run"

kraigher:
Just some minor comments, I think one test you did was maybe overkill and there is a docstring missing. If you fix this I am ready to press the merge button. Thank you for your good work!

1138-4EB:
@dstadelm, I think that you are not being shown in https://github.com/VUnit/vunit/graphs/contributors because the email address you use in your commits is not recognized by GitHub. You can add it as a secondary address to your GitHub profile.

PR #553, opened 2019-09-24
minor cleanup

1138-4EB:
These are just some little simplifications that I was writing as an alternative fix to https://github.com/VUnit/vunit/commit/03c2b9b95ff890e9b8d701d02dda55693e351a96.


kraigher:
Looks OK to me, you already have the permission to merge. The only feedback is that I tend to prefer to use keyword arguments (name=value) when calling functions with many parameters.

Issue #353, opened 2018-07-06
Combine multiple Vunit projects


FranzForstmayr:
I've read this post on SO:
https://stackoverflow.com/questions/42980036/how-to-combine-multiple-vunit-run-py-files-into-a-single-vunit-run

I want to start a discussion, if it is possible to provide a Pyhton API to import another vunit file, without changing old project's "run.py". Has anyone tried this before? Are there any risks?

If you start a "run.py" from a top project, all imported files should be loaded into a sub project. I would like to have the opportunity to compile the project into a specific library. The test runner will start the underlying tests, after that, the tests of the top module are executed.

kraigher:
In principle a run.py file cannot be imported, it is not a data file but rather an executable script. The data of the sub project run.py has to be factored out into a separate file which can be imported into the parent run.py file. In principle what you are asking for is a "module" system which VUnit does not implement. Such a module system should really not be part of VUnit but a separate system on top of VUnit since it should also handle synthesis and other parts of the HDL flow as well. A module system that only handles one aspect of the flow such as simulation is not really useful in my opinion. This is why I do  not want VUnit to have a module system and rather have it as a separate project. Internally at my company we have such a module system which uses VUnit underneath to run tests. 

Examples of open source module systems are fusesoc which has been integrated with VUint to some degree.

FranzForstmayr:
I'm aware about the script/import issue. I thought about some deeper python tricks, like parse the file with ast.

 I want to test a bigger signal chain in vivado. Let's assume you have a input signal, make some operations in module1, maybe a fft in module2 and finally format the data in a last module3. The top module consists of a testbench, which is instantiating the underlying modules. The underlying modules are vunit projects at it's own. 
This is of course possible in Vunit, when i manually include the source files into my top project. I just wanted the possibility to do this easily, i don't want to change existing structures of Vunit. 
 
Are you using Fusesoc? Or do you have other preferences?

kraigher:
I am not using fusesoc but have a comapny internal module system.

1138-4EB:
@FranzForstmayr, I think that you might find interesting the conversation that @dstadelm started on Jun 04 15:46 in [Gitter](https://gitter.im/VUnit/vunit). It is related to this SO question: https://stackoverflow.com/questions/42980036/how-to-combine-multiple-vunit-run-py-files-into-a-single-vunit-run

PR #558, opened 2019-10-06
add_submodule method to import vunit projects

FranzForstmayr:
This PR enable vunit to import separate managed vunit projects at once. 

The vunit class creating has been modified therefore and acts now as a singleton. So every further vunit call returns the same instance, so the project's source files get _merged_ together.

Typical use cases are a top level design like a CPU which imports several peripheries as a submodule. Each periphery is managed separately as vunit project. When importing the module there's no need to track the source files (changes) of the top project anymore. 
Another use case would be libraries. So instead of importing several files from a vhdl library only a run.py from the library gets imported.

I opened an issue a while ago, #353 but it is already closed. I needed the feature anyway so i decided to give it a try.

Testers are welcome to try the new feature.

### DIFF
--- a/examples/vhdl/submodule/run.py
+++ b/examples/vhdl/submodule/run.py
@@ -1,0 +1,34 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# Copyright (c) 2014-2019, Lars Asplund lars.anders.asplund@gmail.com
+
+"""
+Submodule support
+-------------
+This script shows how to import smaller submodules in a bigger design. 
+
+The submodules contains a own run.py script in which the sources are defined.
+With 'add_submodule' you can add the complete project into your top design.
+Usecase could be a CPU design which uses some periphery which is managed in 
+seperate projects.
+"""
+
+from os.path import join, dirname
+from vunit import VUnit
+
+root = dirname(__file__)
+
+vu = VUnit.from_argv()
+
+vu.add_json4vhdl()
+
+lib = vu.add_library("lib")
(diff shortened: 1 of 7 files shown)


kraigher:
For this type of change it is good to ask before starting the implementation. I think adding a submodule concept to VUnit is a good idea but I do not think this is the right approach.  I do not like that there is shared global state in the VUnit object. It will actually break existing code that I know of and will also break the VUnit acceptance tests.

Another concern is a submodule concept should offer modularity and isolation. By having everything being inside one global VUnit object submodules can change settings of other submodules. I would want a solution where submodule A cannot touch anything in submodule B.

Also a run.py file is not a good carrier for defining a submodule since it will contain parsing of command line arguments and other side effects related to the environment. The run.py file is meant to serve as an executable and a means to run the tests. I think submodules should be declared outside of run.py files in files that just serve to declare the submodule.

I think the right aproach is creating a VUnitModule class or similiar which is defined in a file separate from the run.py file. If this approach was taken the user might not even need a run.py file in many cases and VUnit could come with a built-in "run.py" that just accepts a module file.

1138-4EB:
I think that this 'submodule' concept is related to #511.

FranzForstmayr:
Thanks for the feedback, I already thought there will be some negative points, but I had finished the code, so I wanted to try.
Can you  please the explain the VUnitModule a little bit more? 

> It will actually break existing code that I know of and will also break the VUnit acceptance tests.

kraigher:
I think the right way forward is to create an issue for making a submodule system for VUnit. In this issue we need to gather all currently known requirements. Then we can discuss a solution.

GlenNicholls:
Is this issue going to continue being tracked here or will it move elsewhere once requirements are solidified?

kraigher:
I would need to be tracked elsewhere. This goes in the direction of making VUnit into a package manager which I believe should be a separate project. A package manager also needs to work for synthesis.

Issue #562, opened 2019-10-08
vunit_restart does not run files from `modelsim.init_files.after_load`


felixn:
We have a use-case where we use the `modelsim.init_files.after_load` feature to set up some required simulation parameters (via ModelSim `force` commands in a .do file).

This works fine in batch mode and in `--gui` mode for the first simulation run, but when using `vunit_restart` the after_load files are not re-run (`_vunit_source_init_files_after_load` is not rerun).
`vunit_load` re-runs these files, but it also re-runs vsim - throwing away all changes to the waveform viewer and so on.

Can we add the possibility to re-run the after_load files with vunit_restart?
I know there are cases where re-running the after_load files is not desired (e.g. if they're used to set up signals in the waveform viewer), but maybe as an optional parameter to vunit_restart?

If we can agree on an approach, I can work on implementing it.

kraigher:
Maybe there needs to be a new hook that is called `before_run`

felixn:
Closed by #564

PR #564, opened 2019-10-09
Add .init_files.before_run hook (related to issue #562)

felixn:
See issue #562

### DIFF
--- a/vunit/ui.py
+++ b/vunit/ui.py
@@ -136,8 +136,15 @@ The following simulation options are known.
 
 ``modelsim.init_files.after_load``
    A list of user defined DO/TCL-files that is sourced after the design has been loaded.
+   They will be executed during ``vunit_load``, after the top level has been loaded
+   using the ``vsim`` command.
    During script evaluation the ``vunit_tb_path`` variable is defined
    as the path of the folder containing the test bench.
+   Must be a list of strings.
+
+``modelsim.init_files.before_run``
+   A list of user defined DO/TCL-files that is sourced before the simulation is run.
+   They will be executed at the start of ``vunit_restart``.
    Must be a list of strings.
 
 ``modelsim.init_file.gui``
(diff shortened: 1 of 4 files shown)


vunit/ui.py
```diff
@@ -136,10 +136,17 @@
...
    A list of user defined DO/TCL-files that is sourced after the design has been loaded.
+   They will be executed during ``vunit_load``, after the top level has been loaded
+   using the ``vsim`` command.
    During script evaluation the ``vunit_tb_path`` variable is defined
    as the path of the folder containing the test bench.
    Must be a list of strings.
 
+``modelsim.init_files.before_run``
+   A list of user defined DO/TCL-files that is sourced before the simulation is run.
+   They will be executed at the start of ``vunit_restart``.
```
kraigher: This is a bit misleading. The files will be sources in vunit_run which incidentally is called by vunit_restart.

vunit/ui.py
```diff
@@ -158,10 +165,17 @@
...
    A list of user defined DO/TCL-files that is sourced after the design has been loaded.
+   They will be executed during ``vunit_load``, after the top level has been loaded
+   using the ``vsim`` command.
    During script evaluation the ``vunit_tb_path`` variable is defined
    as the path of the folder containing the test bench.
    Must be a list of strings.
 
+``rivierapro.init_files.before_run``
+   A list of user defined DO/TCL-files that is sourced before the simulation is run.
+   They will be executed at the start of ``vunit_restart``.
```
kraigher: This is a bit misleading. The files will be sources in vunit_run which incidentally is called by vunit_restart.

vunit/ui.py
felixn: You're right, I meant to write vunit_run. It's more explicit now, I hope it's clear enough.

vunit/ui.py
felixn: Same as above.

PR #567, opened 2019-10-12
Update release procedure

kraigher:


docs/contributing.rst
```diff
@@ -189,17 +189,17 @@ Making releases
...
-- If required, bump version in ``vunit/about.py``.
-- Add corresponding release notes in ``docs/release_notes/X.Y.Z.rst``. The
-  release notes files in ``docs/release_notes/`` are used to
-  automatically generate the :ref:`release notes <release_notes>`.
-- Commit changes.
-- Execute ``python tools/new_release.py``. To generate a release candidate, use cli arg ``--rc``.
-- Push the new version/tag: ``git push vX.Y.Z``.
-- Bump version in ``vunit/about.py``.
-- Commit changes.
+In create a new tagged release commit:
```
1138-4EB: Typo: should be "To create..." or "In order to commit..."

docs/contributing.rst
```diff
@@ -189,17 +189,17 @@ Making releases
...
+
+- Create corresponding release notes in ``docs/release_notes/X.Y.Z.rst``.
+   - The release notes files in ``docs/release_notes/`` are used to
+     automatically generate the :ref:`release notes <release_notes>`.
+- Execute ``python tools/release.py create X.Y.Z``.
+   - Will make commit with the new ``about.py`` version and release notes and tag it.
+   - Will make another commit setting ``about.py`` to the next pre release candidate version.
+- Push the changes AND tags to origin master.
```
1138-4EB: It might be desirable to suggest pushing the tag only first. This will trigger a build that can be used to check that everything is correct. If something goes wrong, the tag can be force-pushed without doing so in master. When everything is good, master can be safely updated.

kraigher:
I updated the PR based on your comments.

PR #570, opened 2019-10-15
Move new_log_handler to public section and add default parameters

### LINKED ISSUE #566: No public API for logging to file?
I couldn't find a way to log to a file in the logging library without using private pats of the API. Am I missing something?

The following works:
```vhdl
library vunit_lib;

entity tb_logging is
  generic (runner_cfg : string);
end entity tb_logging;

architecture tb of tb_logging is
begin
  main : process
    constant log_file_name: string := output_path(runner_cfg) & "schmift.txt";
    variable file_handler: log_handler_t;
    variable logger: logger_t;
  begin

    while test_suite loop
      if run("stdout") then
        warning("A warning");
      elsif run("file") then
        file_handler := new_log_handler(log_file_name, format => verbose, use_color => false);
        logger := get_logger("custom");
        set_log_handlers(logger, (0 => file_handler));
        show_all(logger, file_handler);
        warning(logger, "A warning");
      elsif run("both") then
        file_handler := new_log_handler(log_file_name, format => verbose, use_color => false);
        logger := get_logger("custom");
        set_log_handlers(logger, (0 => file_handler, 1 => display_handler));
        show_all(logger, file_handler);
        warning(logger, "A warning");
      end if;
    end loop;

    test_runner_cleanup(runner);
  end process;
end architecture;
```

But the `new_log_handler()` function is in the private section of log_handler_pkg.vhd:
https://github.com/VUnit/vunit/blob/214de988c32018d19be4101667185747b0bb69fe/vunit/vhdl/logging/src/log_handler_pkg.vhd#L58-L81

Is there another way to log to files? Or can we make `new_log_handler()` part of the public API?

felixn:
Move new_log_handler to public section and add default parameters (closes #566)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,6 @@ docs/_build
 docs/examples.rst
 docs/release_notes.rst
 vunit/test/acceptance/*_out
+vunit/test/unit/test_report_output.txt
+env/
+venv/
--- a/vunit/vhdl/logging/src/log_handler_pkg-body.vhd
+++ b/vunit/vhdl/logging/src/log_handler_pkg-body.vhd
@@ -61,8 +61,8 @@ package body log_handler_pkg is
   end;
 
   impure function new_log_handler(file_name : string;
-                                  format : log_format_t;
-                                  use_color : boolean) return log_handler_t is
+                                  format : log_format_t := verbose;
+                                  use_color : boolean := false) return log_handler_t is
     constant id : natural := get(next_log_handler_id, 0);
   begin
     set(next_log_handler_id, 0, id + 1);
--- a/vunit/vhdl/logging/src/log_handler_pkg.vhd
+++ b/vunit/vhdl/logging/src/log_handler_pkg.vhd
@@ -55,6 +55,10 @@ package log_handler_pkg is
                        variable format : out log_format_t;
                        variable use_color : out boolean);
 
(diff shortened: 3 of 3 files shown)


1138-4EB:
I'm not sure about adding defaults. There are two optional arguments only, and the second one is a boolean. It's not much work to actually define them. Moreover, I'd personally set use_color to true, as I find it easier to disable it if the output is not desired, rather than thinking that the feature is not supported because it defaults to not being used.

kraigher:
I think for file logging you almost never want color logging so I actually vote for default off.

felixn:
I chose to add default parameters because I think it makes the API easiest to use for the user. Other functions like `set_format(` also have sensible default values.
I agree with kraigher, log files with color escape sequences are basically unreadable in a text editor - that's why I chose off. (`set_format(` also has use_color with false as default)
For format, I chose verbose because that's the default for the display handler as well.
Alternatively, csv makes sense for log files, but I think the user expects the log files to behave same as the displayed messages by default and I think log files are rarely machine-read (which would be best with csv).

1138-4EB:
I normally use `verbose, true` to split logs from sub-components that I then inspect with `cat`; and I use `csv, false` to later machine-read the output. Nonetheless, this is a minor issue which should not prevent the merge.

PR #575, opened 2019-10-23
fix typos

arcturus140:


docs/logging/user_guide.rst
```diff
@@ -261,7 +261,7 @@ checks and consumes one log message from the list of recorded log calls to
 check a sequence of log messages.
 
 The ``mock`` procedures ensures that all future calls to the logger
```
1138-4EB: BTW, could you fix 'procedures ensures' to either 'procedure ensures' or 'procedures ensure'?

docs/logging/user_guide.rst
```diff
@@ -261,7 +261,7 @@ checks and consumes one log message from the list of recorded log calls to
...
-are recorded and not sent to a any log handler or cause simulation
+are recorded and not sent to a log handler or cause simulation
```
1138-4EB: Why keep 'a' instead of 'any'? I'm not a native speaker, so I have no preference. I'm just curious about it.

arcturus140: Not a native speaker myself but you had «to a any log» at least remove one of them. Use 'a' to specify something. 

Any can be used to especially pronounce the fact that there is 'none at all' to be used. It is better fit for edge cases, such as hint, attention, warning...

PR #578, opened 2019-10-28
post_run: add report data to 'Results' object/class

### LINKED ISSUE #445: Feature request: Access to TestReport object in post_run callback
It would be nice to have access to the **TestReport** object within the `post_run` callback of class vunit.ui.**VUnit**, as the current **Results** object only allows for merging coverage reports. This would be useful for programmatically constructing different reports, rather than re-parsing the XML which may or may not have been included in the CLI arguments.

Currently, both the **TestReport** object for printing to the screen and the **Results** object for the callback are both constructed in the `_main_run` function of _ui.py_, before `post_run` is invoked with just the **Results** object.

As **Results** seems to only be a wrapper to pass to the callback, I propose extending this object to take a second constructor argument of the existing **TestReport** object within `_main_run`.

This should not break any existing interfaces, and provided this is an acceptable approach to the project maintainers I am happy to submit a pull request.

### LINKED ISSUE #478: Programatic procedure to retrieve the output path of a test
Ref #467.

> @umarcor:
> Is it possible to read `test_name_to_path_mapping.txt` directly as some Python type (dict or list)?

> @kraigher:
> The file was intended for human use. The format is `folder_name` `SPACE` `test_name`.
I do not recommend that you use it. It is better we add this to `--export-json` instead.

> @kraigher:
> Currently the commands are logged if you use `--log-level=DEBUG` for human use. If an external tool needs this information we need to add it to `--export-json`.

> @umarcor:
> IMHO, there should be a simple and realiable programatic procedure to retrieve the output/log of a test. It don't really mind if I need to set some stdout in the python script, or get it as a return field from `vu.main()`, or parse some output file. I am trying to process `test_name_to_path_mapping.txt` just because I didn't find a better way to 'see' the content in `output.txt`.
>
> About adding it to `--export-json`, although useful, I'd rather avoid it for this specific use case. Using JSON is a better approach when it is to be used from Python, Go, etc. But, with Bash or C I find it easier not to require a JSON parsing library. The current format, as you described it, is straighforward, as one only needs to search for the first space.

### LINKED ISSUE #574: post_run for Test / TestBench
Are there any arguments against adding a `post_run` callback to Test / TestBench?
Similar to `post_check`, but called regardless of whether test passed or failed.

My use case: Processing log files written to output_path, even for failing tests. For example, archiving log files from automated simulation runs.
I know I could do this without a callback by parsing the mapping text file to get to the output directories, but the callback would be a much easier solution.
Or is there another easier way that I'm not seeing?

And I believe there could be other use cases where this callback would be beneficial.

If there is general consensus that this suggestion is useful, I could work on implementing it.

umarcor:
Close #445
Close #478
Close #574

In this PR, parameters `output_path` and `report` are passed from `main` to the `Results` object built and passed to `post_run` callbacks. Instead of exposing the `TestReport` object, as suggested in #445, a subset of info is extracted for each test: name, status, time and path. The path is relative to `join(output_path, 'test_output')`.

Usage example:

```py
def post_func(results):
    print(results.output_path)
    for key, val in results.get_report().items():
        print(key , val)

vu.main(post_run=post_func)
```

@felixn, you can build the path to each test log as follows: `join(results.output_path, 'test_output', val.path, 'output.txt')`. Alternatively: `join(results.output_path, 'test_output', results.get_report().items()[<test_name>], 'output.txt')`.

/cc @richjyoung

### DIFF
--- a/vunit/ui/results.py
+++ b/vunit/ui/results.py
@@ -8,14 +8,19 @@
 UI class Results
 """
 
+from os.path import join, basename, normpath
+from .common import TEST_OUTPUT_PATH
+
 
 class Results(object):
     """
-    Gives access to results after running tests.
+    Gives access to results after running tests
     """
 
-    def __init__(self, simulator_if):
+    def __init__(self, output_path, simulator_if, report):
+        self._output_path = output_path
         self._simulator_if = simulator_if
+        self._report = report
 
     def merge_coverage(self, file_name, args=None):
         """
@@ -24,5 +29,84 @@ class Results(object):
         :param file_name: The resulting coverage file name.
         :param args: The tool arguments for the merge command. Should be a list of strings.
         """
-
         self._simulator_if.merge_coverage(file_name=file_name, args=args)
+
+    def get_report(self):
+        """
+        Get a report (dictionary) of tests: status, time and output path
+
+        :returns: A :class:`Report` object
+        """
+        report = Report(self._output_path)
+        for (
+            test
+        ) in self._report._test_results_in_order():  # pylint: disable=protected-access
+            obj = test.to_dict()
+            report.tests.update(
+                {
+                    test.name: TestResult(
+                        join(self._output_path, TEST_OUTPUT_PATH),
+                        obj["status"],
+                        obj["time"],
+                        obj["path"],
+                    )
+                }
+            )
+        return report
+
+
+class Report(object):
+    """
+    Gives access to test results and paths after running tests
+
+    :data output_path: Absolute path to the output path (see :ref:`cli`)
+    :data tests: Dictionary of :class:`TestResult` objects
+    """
+
+    def __init__(self, output_path):
+        self.output_path = output_path
+        self.tests = {}
+
+
+class TestResult(object):
+    """
+    Gives access to a subset of the results of a test
+
+    :data status: Result status (passed, failed or skipped)
+    :data time: Simulation time
+    :data path: Absolute path of the test output
+
+    :example:
+
+    .. code-block:: python
+
+       def post_func(results):
+           report = results.get_report()
+           print(report.output_path)
+           for key, test in report.tests.items():
+               print(key)
+               print(test.status)
+               print(test.time)
+               print(test.path)
+               print(test.relpath)
+
+       vu.main(post_run=post_func)
+    """
+
+    def __init__(self, test_output_path, status, time, path):
+        self._test_output_path = test_output_path
+        self.status = status
+        self.time = time
+        self.path = path
+
+    @property
+    def relpath(self):
+        """
+        If the path is a subdir to the default TEST_OUTPUT_PATH, return the subdir only
+        """
+        base = basename(self.path)
+        return (
+            base
+            if normpath(join(self._test_output_path, base)) == normpath(self.path)
+            else self.path
+        )
--- a/docs/py/vunit.rst
+++ b/docs/py/vunit.rst
@@ -36,6 +36,10 @@ Results
 
 .. autoclass:: vunit.ui.results.Results()
 
+.. autoclass:: vunit.ui.results.Report()
+
+.. autoclass:: vunit.ui.results.TestResult()
+
 .. |compile_option| replace::
    The name of the compile option (See :ref:`Compilation options <compile_options>`)
 
--- a/tests/unit/test_test_report.py
+++ b/tests/unit/test_test_report.py
@@ -10,9 +10,11 @@ Test the test report functionality
 
 from unittest import TestCase
 from xml.etree import ElementTree
-from os.path import join, dirname
+from os.path import basename, dirname, join
 import os
 from vunit.test.report import TestReport, PASSED, SKIPPED, FAILED
+from vunit.ui.common import TEST_OUTPUT_PATH
+from vunit.ui.results import Results
 
 
 class TestTestReport(TestCase):
@@ -274,14 +276,49 @@ Elapsed time was 3.0 seconds
             ),
         )
 
-    def _report_with_all_passed_tests(self):
+    def test_dict_report_with_all_passed_tests(self):
+        opath = dirname(dirname(self.output_file_name))
+        test_path = join(opath, TEST_OUTPUT_PATH, "unit")
+        output_file_name = join(test_path, basename(self.output_file_name))
+        results = Results(
+            opath, None, self._report_with_all_passed_tests(output_file_name),
+        )
+        report = results.get_report()
+        for key, test in report.tests.items():
+            self.assertEqual(
+                basename(test.path), test.relpath,
+            )
+        test0 = report.tests["passed_test0"]
+        test1 = report.tests["passed_test1"]
+        self.assertEqual(
+            {
+                "passed_test0": {
+                    "status": test0.status,
+                    "time": test0.time,
+                    "path": test0.path,
+                },
+                "passed_test1": {
+                    "status": test1.status,
+                    "time": test1.time,
+                    "path": test1.path,
+                },
+            },
+            {
+                "passed_test0": {"status": "passed", "time": 1.0, "path": test_path},
+                "passed_test1": {"status": "passed", "time": 2.0, "path": test_path},
+            },
+        )
+
+    def _report_with_all_passed_tests(self, output_file_name=None):
         " @returns A report with all passed tests "
+        if not output_file_name:
+            output_file_name = self.output_file_name
         report = self._new_report()
         report.add_result(
-            "passed_test0", PASSED, time=1.0, output_file_name=self.output_file_name
+            "passed_test0", PASSED, time=1.0, output_file_name=output_file_name
         )
         report.add_result(
-            "passed_test1", PASSED, time=2.0, output_file_name=self.output_file_name
+            "passed_test1", PASSED, time=2.0, output_file_name=output_file_name
         )
         report.set_expected_num_tests(2)
         return report
--- a/vunit/test/report.py
+++ b/vunit/test/report.py
@@ -14,6 +14,7 @@ from sys import version_info
 import os
 import socket
 import re
+from os.path import dirname
 from vunit.color_printer import COLOR_PRINTER
 from vunit.ostools import read_file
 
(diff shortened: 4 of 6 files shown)


vunit/ui/results.py
```diff
@@ -26,3 +31,20 @@ def merge_coverage(self, file_name, args=None):
...
 
         self._simulator_if.merge_coverage(file_name=file_name, args=args)
+
+    def get_report(self):
+        """
+        Get a report (dictionary) of tests: status, time and output path
+        """
+        return {
+            test.name: {
+                "status": test._status.name,  # pylint: disable=protected-access
```
kraigher: Add a method to_dict to test instead of using private access

vunit/ui/results.py
```diff
@@ -26,3 +31,20 @@ def merge_coverage(self, file_name, args=None):
...
+                "time": test.time,
+                "path": re.sub(
```
kraigher: Use os. path api for manipulation of paths

felixn: Thanks for working on this!

In my opinion, path should be the full path to the output directory of individual paths instead of a relative path. That would make working with the results/paths easier, because I think the common case requires the full path anyways.

umarcor: I'm not sure about it. The point with the regular expression is that no replacement (removal) is done if the path is, for any reason, not a child of output_path. In order to do so with `os.path`, it seems that a helper function needs to be written. This is because, [relpath](https://docs.python.org/3/library/os.path.html#os.path.realpath) will return a relative path, even if the test path is not a subset of output path. Therefore, some combination with [commonprefix](https://docs.python.org/3/library/os.path.html#os.path.commonprefix) is required. However, `commonprefix` is buggy, and a better solution was introduced in Python 3.5: [commonpath](https://docs.python.org/3/library/os.path.html#os.path.commonpath). See https://stackoverflow.com/questions/7287996/python-get-relative-path-from-comparing-two-absolute-paths. Alternatively, in Python 3.4 pathlib was added https://docs.python.org/3/library/pathlib.html.

Overall, I think it is not worth implementing a solution based on `os.path` which is compatible with both Python 2.7 and Python >=3.5. We'd better keep the simple regex based approach and revisit this when Python 2 is deprecated. What do you think?

vunit/ui/results.py
umarcor: I added it now. However, private access is required for `_test_results_in_order()`. Should I add `to_dict` to `TestReport` too?

vunit/ui/results.py
umarcor: The motivation is to reduce duplication, since all the tests will normally be generated in the same dir. I think that the relevant information in the path is the hash that is appended to the test name, the piece that you cannot know beforehand. With this format, `path name` matches the content of `test_name_to_path_mapping.txt`.

Nonetheless, we can provide a helper function to build the path (`results.get_full_path(val, <location of desired artifact>`). Alternatively, we can provide both the full path and the hash (without the lib name). @kraigher, wdyt? If we provide the *raw* full path, the discussion above about regexp or `os.path` doesn't apply anymore.

felixn: My thinking is, if we're interested in the output path, then it's because we're trying to do something with the files in there, in which case we want to work with the full path anyways. I can't think of a good usecase where we only want the name of the folder (e.g. `lib.tb_example_many.01.test_fail_eb05d78caf9a3f67d98b38ccc6f078898ace39e9`), but in any case that's easy to get with `os.path.basename`. And that feels cleaner to me than having to build the path with `join(results.output_path, 'test_output', ...)`

felixn: @kraigher @1138-4EB :
Do you have an opinion about the remaining open topics (full path/relative path, `to_dict` for `TestReport`, ...)? I'd love to get this pull request merged.

vunit/test_report.py
```diff
@@ -322,3 +323,16 @@ def to_xml(self, xunit_xml_format):
             skipped = ElementTree.SubElement(test, "skipped")
             skipped.attrib["message"] = "Skipped"
         return test
+
+    def to_dict(self):
+        """
+        Convert a subset of the test result to a dictionary
+        """
+        return (
+            self.name,
```
kraigher: Just return the dict. The name is already an attribute.

vunit/ui/results.py
```diff
@@ -24,5 +29,19 @@ def merge_coverage(self, file_name, args=None):
...
+        """
+        Get a report (dictionary) of tests: status, time and output path
+        """
+        report = {}
+        for (
+            test
+        ) in self._report._test_results_in_order():  # pylint: disable=protected-access
+            name, obj = test.to_dict()
+            obj["path"] = re.sub(
+                "^%s" % join(self.output_path, "test_output") + sep, "", obj["path"]
```
kraigher: Duplication of information with hard-coded "test_output"

kraigher: Also at least wrap the regex operation in a function that documents what it wants to achive.

umarcor: > Duplication of information with hard-coded "test_output"

I don't understand what you mean.

kraigher: There is somewhere else in the code where the test output path was decided. If that path changed the code in the PR would be wrong since it duplicates the knowledge of where the test output path is.

vunit/ui/results.py
umarcor: > I can't think of a good usecase where we only want the name of the folder

When using different docker containers to execute multiple tasks in a pipeline, it is desirable to know the relative paths because absolute location inside the containers/VMs can be very different.

Anyway I reworked the returned object, which is a class now and it provides a helper function:

```py
def post_func(results):
    report = results.get_report()
    print(report.output_path)
    for key, val in report.tests.items():
        print(key, val)
        print(report.get_full_path(val["path"]))
        print(report.get_full_path(val["path"], "output.txt"))

vu.main(post_run=post_func)
```

vunit/ui/results.py
umarcor: Now, `TEST_OUTPUT_PATH = "test_output"` is defined in `vunit.ui.common` and it is imported by both `vunit.ui.vunit` and `vunit.ui.results`.

umarcor:
> Also there are no new tests added for this.

I didn't find tests in `vunit/test/unit` for `merge_coverage`. Should I create a new `vunit/test/unit/test_results.py` file? Or can I add them to `vunit/test/unit/test_test_report.py`?

kraigher:
I think we should not expose dicts for individual test results on the public api since it is harder to change without breaking in the future. Also see my other line comments.

umarcor:
> I think we should not expose dicts for individual test results on the public api since it is harder to change without breaking in the future.

I added and used class `ReportResult` now. Let me know if that's what you meant.

> Also see my other line comments.

I don't see any other comments. Maybe you have some unfinished review?

vunit/ui/results.py
```diff
@@ -24,5 +29,62 @@ def merge_coverage(self, file_name, args=None):
...
+
+    def get_full_path(self, test_path, file_path=""):
+        """
+        Return the full path to a test subdir or to a file within it
+        """
+        return join(self.output_path, TEST_OUTPUT_PATH, test_path, file_path)
+
+
+class ReportResult(object):
+    """
```
kraigher: TestResult is a better name

vunit/ui/results.py
```diff
@@ -24,5 +29,62 @@ def merge_coverage(self, file_name, args=None):
...
+class Report(object):
+    """
+    Gives access to test results and paths after running tests
+    """
+
+    def __init__(self, output_path):
+        self.output_path = output_path
+        self.tests = {}
+
+    def get_full_path(self, test_path, file_path=""):
```
kraigher: This method belongs on the individual test result (ReportResult) class.

vunit/ui/results.py
umarcor: `TestResult` is already used in `vunit/test_report.py`. I tried not to repeat the name of the class, even though there is no collision because this is located in submodule `vunit.ui`. Should I change it anyway?

vunit/ui/results.py
```diff
@@ -24,5 +30,41 @@ def merge_coverage(self, file_name, args=None):
...
+        class Report(object):
+            """
+            Gives access to test results and paths after running tests.
+            """
+
+            def __init__(self, output_path):
+                self.output_path = output_path
+                self.tests = {}
+
+            def get_full_path(self, test_path, file_path=""):
```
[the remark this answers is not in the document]
umarcor: Since moving `get_full_path` requires to store `output_path` in each `ReportResult` object, I a took a different approach: `path` is saved raw in `ReportResult` (as @felixn preferred), and `report.get_relpath(test_path)` is provided.

vunit/ui/results.py
```diff
@@ -24,5 +30,41 @@ def merge_coverage(self, file_name, args=None):
...
+
+        def get_relative_test_path(test_path):
+            """
+            If the test_path is relative to the default TEST_OUTPUT_PATH, return the subdir only
+            """
+            return re.sub(
+                "^%s" % normpath(join(self._output_path, TEST_OUTPUT_PATH)) + sep, "", normpath(test_path)
+            )
+
+        class Report(object):
```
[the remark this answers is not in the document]
umarcor: Classes `Report` and `ReportResult` (maybe renamed to `TestResult`) are not nested any more.

Also, I added them to the website via sphinx.

vunit/ui/results.py
umarcor: See https://github.com/VUnit/vunit/pull/578#discussion_r345582338

vunit/ui/results.py
kraigher: Yes I think it is a better name. As you say it is not a name conflict.

vunit/ui/results.py
```diff
@@ -24,5 +29,76 @@ def merge_coverage(self, file_name, args=None):
...
+
+    :data output_path: Absolute path to the output path (see :ref:`cli`)
+    :data tests: Dictionary of :class:`ReportResult` objects
+    """
+
+    def __init__(self, output_path):
+        self.output_path = output_path
+        self.tests = {}
+
+    def get_relpath(self, path):
```
kraigher: I still think this should be property of the individual tests result since the path is associated with an individual test. The fact that the test object needs a private copy of the output path is no reason not to do it.

PR #582, opened 2019-11-03
Create sim_if and test submodules

1138-4EB:
Following the same approach as with `ui.py`, in this PR:

- A `vunit.sim_if` submodule is created.
- Submodule `test` is renamed to `tests`.
- A new submodule names `test` is created, and all the `vunit/test_*.py` sources are moved there.
- `vunit.ui.vunit` is merged into `vunit/ui/__init__.py` and unused imports are removed.


kraigher:
I am basically ok with this as it makes the structure better. Regarding test vs tests I think tests should really be outside the vunit folder. Moving it outside also makes it less confusing.

1138-4EB:
> Regarding test vs tests I think tests should really be outside the vunit folder. Moving it outside also makes it less confusing.

Good idea! It is done now. BTW, `builtins` imported `simulator_check` from `vunit.test.common` (which is now `tests.common`). However, I think that `vunit` should not depend on `tests`. Hence, I moved three helper functions from `tests.common` to `vunit.sim_if.common`.

Issue #584, opened 2019-11-11
Active-HDL 10.5a not adding package generics


aappl:
When using Active-HDL 10.5a as the simulator, the OSVVM packages that contain package generics are not added. The packages missed are "ScoreboardGenericPkg.vhd", "ScoreboardPkg_int.vhd", and "ScoreboardPkg_slv.vhd". When running with 10.5 the files are included.

To reproduce:
- Set environment variables to select Active-HDL 10.5a.
- Add the add_osvvm() call to your run.py.
- Run your run.py file with the -f flag.
- See that the OSVVM files listed do not include the ones stated above.

It looks like the VersionConsumer class in activehdl_interface.py doesn't handle non-digits in the version number.

LarsAsplund:
I don't have both versions but does it works for you if you change the identified regexp to `_version_re = re.compile(r"(?P<major>\d+)\.(?P<minor>\d+)[a-zA-Z]?\.\d+\.\d+")`?

aappl:
Yes, that does seem to fix it for 10.5a.

LarsAsplund:
Good. One more question before we can update the code. Is 10.5a a later release then 10.5 or the other way around?

aappl:
10.5a is the later version.

PR #585, opened 2019-11-14
Fix version number bug in activehdl_interface.py

aappl:
FIxed bug relating to issue: Active-HDL 10.5a not adding package generics #584


vunit/activehdl_interface.py
```diff
@@ -453,7 +453,7 @@ def __init__(self):
         self.major = None
         self.minor = None
 
-    _version_re = re.compile(r"(?P<major>\d+)\.(?P<minor>\d+)\.\d+\.\d+")
+    _version_re = re.compile(r"(?P<major>\d+)\.(?P<minor>\d+)[a-zA-Z]?\.\d+\.\d+")
```
LarsAsplund: Now that we've had a bug it would be appropriate to add a test case that would have detected this bug. Since the version number is used to compare if one release is newer than another we also need  a way to determine the order between versions like 10.5 and 10.5a. Currently the "a" is just discarded. Are you up to making these additions?

aappl: Yes I can make these additions. Could you help me to understand why we need to check the letter portion of the version? From the code it seems like we are checking if the version is 10.1 or greater. As a version with a letter is a later release than one without, can we not safely assume that any version with a letter that is greater than 10.1 (e.g. 10.1a, 10.2a, etc) will also be valid? And therefore the letters are just sub-versions like the 3rd and 4th digits, which are also discarded? Thanks.

LarsAsplund: Whether or not to include the letter really depends on what it means. I'm used to the second digit indicating a new functional feature (which VUnit may depend upon). If 10.1a can have a new feature that 10.1 doesn't have we need a way to separate those versions. Not in this particular case but if we were to depend on a feature released in, let say 12.2c, sometime in the future. OTOH, if the letter denotes a bug fix release we can ignore it just like the 3rd and 4th digit. Do you know? If the letter denotes a new feature release I think we should take it into account since it's easy to miss that once it really matters.

1138-4EB:
@aappl the two files that you modified have been moved in `master`. `vunit/activehdl_interface.py` is now `vunit/sim_if/activehdl.py` and `vunit/test/unit/test_activehdl_interface.py` is now `tests/unit/test_activehdl_interface.py`. Just in case you have any issue when rebasing this PR.

vunit/activehdl_interface.py
aappl: I do not know if it means feature or bug fix, but its position next to the second digit does make it seem like a feature. I will add some code to capture it in the VersionConsumer as an extra attribute and check it with the tests. I don't know how I would use the new attribute in the supports_vhdl_package_generics class method as it will work with or without the new attribute. Would it be ok to just leave that method as it is, or do you have something in mind?

LarsAsplund: What you can do is to create something like

``` python
class Version(object):
    def __init__(self, major=None, minor=None, minor_letter=None):
        self.major = major
        self.minor = minor
        self.minor_letter = minor_letter
```

and then let `VersionConumer` inherit from that and just add the `__call__` method. To enable comparison between these classes you can use `@total_ordering` on the `Version` class, see https://github.com/VUnit/vunit/blob/master/vunit/vhdl_standard.py for inspiration. With that in place a test case like this should pass:

``` python
def test_comparison(self):
    consumer = VersionConsumer()
    consumer("some version line with 10.5a")
    self.assertTrue(consumer > Version(10, 5))
    self.assertFalse(consumer > Version(10, 5, "a"))
```

LarsAsplund:
@aappl Thanks for your contribution!

aappl:
@LarsAsplund You're very welcome!

PR #586, opened 2019-11-14
update .gitignore

umarcor:
Some minor fixes to `.gitignore`, plus ensuring that all temporal output directories end in `*_out`.

### DIFF
--- a/tests/acceptance/test_artificial.py
+++ b/tests/acceptance/test_artificial.py
@@ -26,7 +26,7 @@ class TestVunitArtificial(unittest.TestCase):
 
     def setUp(self):
         # Spaces in path intentional to verify that it is supported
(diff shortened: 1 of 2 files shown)


tests/acceptance/test_artificial.py
```diff
@@ -26,7 +26,7 @@ class TestVunitArtificial(unittest.TestCase):
 
     def setUp(self):
         # Spaces in path intentional to verify that it is supported
-        self.output_path = join(dirname(__file__), "artificial out")
+        self.output_path = join(dirname(__file__), "artificial_out")
```
felixn: The comment in the line above makes me think that it was a concious decision to use a space instead of an underscore.
Why not simply vunit/test/acceptance/artificial\ out in .gitignore?

PR #593, opened 2019-11-26
Use integer_array_t instead of array_t

1138-4EB:
In this PR, examples 'array' and 'array_axis_vcs' are modified to use `integer_array_t` instead of `array_t`. Also, a deprecation warning is added, which is shown when `add_array_util()` is used.

This PR is expected to be merged in the next release, and the following removal branch should be merged in future releases: https://github.com/1138-4EB/vunit/compare/array_t...1138-4EB:rm-array_pkg?expand=1

Ref https://github.com/1138-4EB/vunit/commit/9020d6ed6802d70a7f8561734eab29f67e044489#r36134178

### DIFF
--- a/examples/vhdl/array/src/test/tb_sobel_x.vhd
+++ b/examples/vhdl/array/src/test/tb_sobel_x.vhd
@@ -10,59 +10,76 @@ use ieee.numeric_std.all;
 
 library vunit_lib;
 context vunit_lib.vunit_context;
-use vunit_lib.array_pkg.all;
-
 
 library osvvm;
-use osvvm.RandomPkg.all;
+use osvvm.RandomPkg.RandomPType;
 
 entity tb_sobel_x is
   generic (
     runner_cfg : string;
-    tb_path : string);
+    tb_path    : string
+  );
 end entity;
 
 architecture tb of tb_sobel_x is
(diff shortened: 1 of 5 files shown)


examples/vhdl/array/src/test/tb_sobel_x.vhd
```diff
@@ -10,79 +10,94 @@ use ieee.numeric_std.all;
...
+            )
+          );
         end loop;
       end loop;
 
     end procedure;
 
     variable rnd : RandomPType;
 
-    procedure randomize(variable arr : inout array_t) is
```
kraigher: Keep this but make it return an image and take parameters for dimensions.

examples/vhdl/array/src/test/tb_sobel_x.vhd
```diff
@@ -10,79 +10,94 @@ use ieee.numeric_std.all;
...
     end procedure;
 
     procedure test_random_image(width, height : natural) is
     begin
-      image.init_2d(width => width, height => height,
-                    bit_width => input_tdata'length,
-                    is_signed => false);
-      randomize(image);
-      sobel_x(image, result => reference_image);
+      image := new_2d(
```
kraigher: Create with randomize function.

PR #595, opened 2019-11-27
json4vhdl: use base16 encodings

### LINKED ISSUE #588: Constraints on generic types
> https://github.com/VUnit/vunit/pull/557#discussion_r347529855
> vunit does not support real to be passed as generic.

> https://github.com/VUnit/vunit/pull/557#discussion_r349299021
> It's not VUnit but the simulator that sets the limitations on what generic types that are supported. GHDL used to be limited to strings only (to support VUnit runner_cfg).

> https://github.com/VUnit/vunit/pull/557#discussion_r349309165
> Shall we update that constraint? I just tested the following in GHDL and it seems to work:
> 
> ```vhdl
> entity ent is
>  generic (
>     a: string := "hello";
>     b: natural range 0 to 100 := 5;
>     c: integer := -15;
>     d: real := 2.75
>  );
>  end;
> ```
> 
> A proper test to see which types of generics are supported by each simulator might better fit in https://github.com/VHDL/Compliance-Tests. However, in VUnit we can test (and hopefully use) at least those four. Of course, this is out of the scope of this PR.

> https://github.com/VUnit/vunit/pull/557#discussion_r349320189
> I think we should keep it so that we can run the acceptance test on all supported simulators

---

This issue is related to using JSON to pass *complex* generics, as it is done in example `json2vhdl`. As commented in https://github.com/VUnit/vunit/commit/45870be9e7f5b81a98a2325e0a47b77139604cbb#r36112229, it might be desirable to split/extend the example to three separate tests:

- Pass path to a JSON file.
- Pass not-too-complex JSON string.
- Pass a complex JSON string.

eine:
Close #588

This is a proof of concept to encode stringified JSON generics in Python and decode them in VHDL. Example json4vhdl is reworked to handle three separate tests: stringified JSON, stringified and b16encoded JSON, and path to a JSON file.

In the current implementation, a package named `encodings` is added to VUnit. This includes `b16decode` and `b16encode` only, for now. Nonetheless, as commented in #588, this package might be merged into JSON-for-VHDL, instead of being added to VUnit.

### DIFF
--- a/examples/vhdl/json4vhdl/run.py
+++ b/examples/vhdl/json4vhdl/run.py
@@ -14,17 +14,29 @@ This is an alternative to composite generics, that supports any depth in the con
 """
 
 from pathlib import Path
-from vunit import VUnit, read_json, encode_json
+from vunit import VUnit
+from vunit.json4vhdl import read_json, encode_json, b16encode
 
 TEST_PATH = Path(__file__).parent / "src" / "test"
 
 VU = VUnit.from_argv()
 VU.add_json4vhdl()
 
-VU.add_library("test").add_source_files(TEST_PATH / "*.vhd")
+LIB = VU.add_library("test")
+LIB.add_source_files(TEST_PATH / "*.vhd")
 
 TB_CFG = read_json(TEST_PATH / "data" / "data.json")
 TB_CFG["dump_debug_data"] = False
-VU.set_generic("tb_cfg", encode_json(TB_CFG))
+JSON_STR = encode_json(TB_CFG)
+JSON_FILE = Path("data") / "data.json"
+
+TB = LIB.get_test_benches()[0]
+
+TB.get_tests("stringified*")[0].set_generic("tb_cfg", JSON_STR)
+TB.get_tests("b16encoded stringified*")[0].set_generic("tb_cfg", b16encode(JSON_STR))
+TB.get_tests("JSON file*")[0].set_generic("tb_cfg", JSON_FILE)
+TB.get_tests("b16encoded JSON file*")[0].set_generic(
+    "tb_cfg", b16encode(str(TEST_PATH / JSON_FILE))
+)
 
 VU.main()
--- a/examples/vhdl/json4vhdl/src/test/tb_json_gens.vhd
+++ b/examples/vhdl/json4vhdl/src/test/tb_json_gens.vhd
@@ -10,71 +10,73 @@ context JSON.json_ctx;
 
 entity tb_json_gens is
   generic (
-    runner_cfg  : string;
-    tb_path     : string;
-    tb_cfg      : string;
-    tb_cfg_file : string := "data/data.json"
+    runner_cfg : string;
+    tb_path    : string;
+    tb_cfg     : string
   );
 end entity;
 
 architecture tb of tb_json_gens is
 
-  -- tb_cfg contains stringified content
-  constant JSONContent     : T_JSON := jsonLoad(tb_cfg);
-
-  -- tb_cfg is the path of a JSON file
-  constant JSONFileContent : T_JSON := jsonLoad(tb_path & tb_cfg_file);
-
-  -- record to be filled by function decode
-  type img_t is record
-    image_width     : positive;
-    image_height    : positive;
-    dump_debug_data : boolean;
-  end record img_t;
-
-  -- function to fill img_t with content extracted from a JSON input
-  impure function decode(Content : T_JSON) return img_t is
-  begin
-    return (image_width => positive'value( jsonGetString(Content, "Image/0") ),
-            image_height => positive'value( jsonGetString(Content, "Image/1") ),
-            dump_debug_data => jsonGetBoolean(Content, "dump_debug_data") );
-  end function decode;
-
-  constant img : img_t := decode(JSONContent);
-
-  -- get array of integers from JSON content
-  constant img_arr : integer_vector := jsonGetIntegerArray(JSONContent, "Image");
-
 begin
   main: process
+
+    procedure run_test(JSONContent : T_JSON) is
+      -- get array of integers from JSON content
+      constant img_arr : integer_vector := jsonGetIntegerArray(JSONContent, "Image");
+    begin
+      -- Content extracted from the JSON
+      info("JSONContent: " & lf & JSONContent.Content);
+
+      -- Integer array, extracted by function jsonGetIntegerArray with data from the JSON
+      for i in 0 to img_arr'length-1 loop
+        info("Image array [" & integer'image(i) & "]: " & integer'image(img_arr(i)));
+      end loop;
+
+      -- Image dimensions as strings, get from the content from the JSON file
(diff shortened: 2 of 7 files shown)


vunit/vhdl/encodings/test/tb_encodings_pkg.vhd
```diff
@@ -0,0 +1,31 @@
...
+use vunit_lib.run_pkg.all;
+use vunit_lib.check_pkg.all;
+use vunit_lib.encodings_pkg.all;
+
+entity tb_encodings is
+  generic (runner_cfg : string);
+end;
+
+architecture a of tb_encodings is
+  constant str: string := "[""test"",[true,false,18,null,""hello""],[9,8],3324.34,832432,""world""]";
```
felixn: Maybe making str and enc generics of tb_encodings and passing them in from run.py tests more / closer to the json usecase? It would also make it very easy to use this to test maximum generic string lengths of various simulators.

1138-4EB: The tests in `tb_encodings` are to test encoding/decoding functions/procedures, which we should be able to do independently of the generic support of the simulator. I think that what you suggest belongs to example json4vhdl.

vunit/vhdl/encodings/src/encodings_pkg.vhd
```diff
@@ -0,0 +1,59 @@
...
+  function to_natural (c: character) return natural is
+    constant offset_0 : natural := character'pos('0');
+    constant offset_U : natural := character'pos('A') - 10;
+    constant offset_l : natural := character'pos('a') - 10;
+    variable num: integer := -1;
+  begin
+    num := character'pos(c);
+    case c is
+      when '0' to '9' => return num - offset_0;
+      when 'A' to 'F' => return num - offset_U;
```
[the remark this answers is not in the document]
1138-4EB: Actually, this is a naive implementation, since I was expecting someone to tell me that these conversion functions are already implemented somewhere in the codebase, or in any of the dependencies (say OSVVM). I must not be the first one converting a character representing an hex digit to a natural/integer...

vunit/vhdl/encodings/src/encodings_pkg.vhd
```diff
@@ -0,0 +1,59 @@
...
+    variable res: string (1 to str'length/2);
+  begin
+    for x in 1 to res'length loop
+      res(x) := character'val(to_natural(str(2*x-1)) *16 + to_natural(str(2*x)));
+    end loop;
+    return res;
+  end;
+
+  function to_character(num: natural) return character is
+  begin
```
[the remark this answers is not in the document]
1138-4EB: Done!

vunit/vhdl/encodings/test/tb_encodings_pkg.vhd
1138-4EB: I now extended the existing `tb_set_generic` acceptance test, to pass a long generic string. The content is created repeating `0123456789abcdef` `str_long_num` times. Now, it is set to 512; hence, the length of the string is 512*16=8192. We should set `str_long_num` to the minimum length that we want any simulator to support. Do you have any reference of which is a sensible value?

Moreover, if other encondings are added, I believe that this testbench can be extended to test the character sets used in other encodings (b32, b64).

vunit/vhdl/encodings/src/encodings_pkg.vhd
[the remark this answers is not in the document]
LarsAsplund: What you can do is to import the fixed package

``` vhdl
library ieee;
use ieee.fixed_pkg.all;
```

and then decode like this

``` vhdl
    function base16_decode(str: string) return string is
      constant str_i : string(1 to str'length) := str;
      variable result: string (1 to str'length / 2);
      variable character_code : ufixed(7 downto 0);
    begin
      for x in result'range loop
        character_code := from_hex_string(str_i(2 * x - 1 to 2 * x));
        result(x) := character'val(to_integer(character_code));
      end loop;

      return result;
    end;
```

vunit/vhdl/encodings/src/encodings_pkg.vhd
```diff
@@ -0,0 +1,60 @@
...
+    variable num: natural;
+  begin
+    for x in 1 to str'length loop
+      num := character'pos(str(x));
+      res(2*x-1 to 2*x) := to_character(num / 16) & to_character(num rem 16);
+    end loop;
+    return res;
+  end function;
+
+end;
```
LarsAsplund: All this being said I rather see that we try to push this to the JSON4VHDL repo. This means two things
1. CamelCase identifiers
2. The user doesn't really need to know what encoding we use to make it safe. jsonLoad could detect what format that is provided and do an initial base16 decoding if necessary.

vunit/vhdl/encodings/src/encodings_pkg.vhd
```diff
@@ -0,0 +1,60 @@
...
+
+  function to_character(num: natural) return character is
+  begin
+    if num<10 then
+      return character'val(offset_0 + num);
+    end if;
+    return character'val(offset_l + num);
+  end;
+
+  function b16encode(str: string) return string is
```
LarsAsplund: Can be rewritten as

``` vhdl
    function base16_encode(str: string) return string is
      constant str_i : string(1 to str'length) := str;
      variable result: string (1 to str'length * 2);
      variable character_code : string(1 to 4);
    begin
      for x in str_i'range loop
        character_code := to_hex_string(to_ufixed(character'pos(str_i(x)), 7, 0));
        result(2 * x - 1 to 2 * x) := lower(character_code(1 to 2));
      end loop;

      return result;
    end;

```

vunit/vhdl/encodings/src/encodings_pkg.vhd
```diff
@@ -0,0 +1,60 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this file,
+-- You can obtain one at http://mozilla.org/MPL/2.0/.
+--
+-- Copyright (c) 2014-2019, Lars Asplund lars.anders.asplund@gmail.com
+
+package encodings_pkg is
+
+  function b16decode(str: string) return string;
```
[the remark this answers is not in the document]
1138-4EB: Although it is true that I like brevity, in this case it was not my preference; I just copied the function names that are used in Python: https://docs.python.org/3/library/base64.html So, e.g. content encoded with `base64.b32encode(s)` in Python is read with `encodings_pkg.b32decode(c)` in VHDL.

vunit/vhdl/encodings/src/encodings_pkg.vhd
1138-4EB: See Paebbels/JSON-for-VHDL#9.

1138-4EB:
This PR is now based on Paebbels/JSON-for-VHDL#9.

eine:
Paebbels/JSON-for-VHDL#9 is now merged upstream.

PR #596, opened 2019-11-27
Remove Python 2.7 support

kraigher:


setup.py
```diff
@@ -72,6 +71,7 @@ def find_all_files(directory, endings=None):
         "Topic :: Software Development :: Testing",
         "Topic :: Scientific/Engineering :: Electronic Design Automation (EDA)",
     ],
+    python_requires=">=3.4",
```
1138-4EB: Shouldn't it be `>=3.5` or `>3.4`, for now?

PR #597, opened 2019-11-27
Python 2 End-Of-Life

umarcor:
This is a PR against feature branch `no-python27`:

- I don't think it is required to run acceptance tests twice (mcode and LLVM). Instead, LLVM is used for acceptance tests and mcode is used for vcomponents.
- `from __future__ import` statements are removed.


umarcor:
Closing because the content is already squashed manually.

PR #600, opened 2019-11-30
remove 'array_pkg'

eine:
Now that v4.3.0 is released, which deprecates `array_pkg` (see #593), I think we can remove it from the codebase.


LarsAsplund:
I'm ok with this but there is a merge conflict to resolve

Issue #603, opened 2019-12-02
Proposal for a more comprehensive approach to VUnit co-simulation


bradleyharden:
@LarsAsplund, @kraiger and @umarcor,

I would like to propose a broad strategy for co-simulation with VUnit. I really like what @umarcor has been working on, but I think his approach leaves most of the implementation to users. If I understand #568 correctly, users have to manually synchronize between VHDL and C/Python using clock signals run through a `string_ptr`. That approach is extremely powerful and flexible, but to me, it doesn't seem very easy to use or approachable for new users.

So far, I've found VUnit to be extremely easy to learn and use, because the VHDL libraries abstract away many of the low-level details. I never even knew `string_ptr_pkg` existed until I started reading the VUnit codebase out of interest. However, even with simple, easy-to-use libraries, I have struggled to convince my co-workers to learn and use VUnit. I fear it will be even harder to convince them to use VUnit for co-simulation if they have to implement most of the low-level details.

Based on the work I've already done for #521 and #522, and the discussion in #583, I would like to propose a more comprehensive approach to co-simulation. **Essentially, I would like to communicate with C/Python code in the same way that I currently communicate with concurrent actors in VHDL, using messages and queues**. I think this approach will require more work than what @umarcor has proposed, but I think the end result will be much more user-friendly. Furthermore, there is a lot of overlap between this proposal and my work on #521 and #522.

To support my proposal, I spent some time this past week prototyping various pieces. Here is an overview of a roadmap, as I see it:

1. Implement `ext_ptr.c` and `ext_ptr_pkg` - **Mostly complete**

This library is essentially a port of `string_ptr_pkg` and `integer_vector_ptr_pkg` to C. It serves the same purpose in C as those packages do in VHDL. It is only responsible for providing shared memory among concurrent actors. It is _not_ used for synchronizing between actors. I already have a complete, working implementation, _with tests_, [here](https://github.com/bradleyharden/vunit/tree/ext_pkg_prototypes/vunit/vhdl/data_types/src/ext_ptr).

2. Modify `string_ptr_pkg` and `integer_vector_ptr_pkg` to cast to and from `ext_ptr_t` - **Trivial**

This change would make it possible to create instances of `string_ptr` and `integer_vector_ptr` in VHDL that are stored externally. This is the infrastructure for implementing external queues. Because `ext_ptr_pkg` and `ext_ptr.c` mimic the structure of the two VHDL `ptr` packages, adding this would be trivial.

[...]

PR #606, opened 2019-12-06
Move VHPIDIRECT related sources (implementation and example) to separate repo

umarcor:
As commented in https://github.com/VUnit/vunit/issues/603#issuecomment-562176366, this is a draft to remove VHDL and C sources related to VHPIDIRECT from this repository. The prosopal is as follows:

- Rename `vunit/vhdl/data_types/src/external` to `vunit/vhdl/data_types/src/api`:
  - Remove subdir `ghdl`.
  - Remove `external_*-vhpi.vhd`.
  - Merge `external_*-novhpi.vhd` and `external_*-body.vhd`; rename to `external_*_pkg_.vhd`.
- Remove `examples/vhdl/external_buffer` (remove it from ` tests/acceptance/test_external_run_scripts.py` too).
- `builtins._add_data_types` is simplified accordingly:
  - `{'string': ['path/to/custom/file'], 'integer': ['path/to/custom/file']}` is supported. The default `vunit/vhdl/data_types/src/api/external_*_pkg_.vhd` is used when: any of the fields does not exist, is `False`, or is `None`. Otherwise, the user-provided file is used.
- `builtins._add_files` is enhanced in order to check if provided files actually exist.
- #581 is merged in this PR.
- A section about 'External VHDL API' is added to the docs. It includes a brief description, a diagram and a note referring users to VUnit/cosim.
  - At the same time, a step is added to workflow 'docs' so that the artifacts of the PRs are available as a zipfile.

All the content that has been removed (the example, `*-vhpi.vhd` sources and subdir `ghdl`) is expected to be added to a new `vunit/cosim` repository. In fact, this draft PR is a result of the discussion that is going on in #603.

Note that this does not revert the 'external modes' feature introduced in `v4.3.0`. Files in `vunit/vhdl/data_types/*.vhd` are not modified at all. Hence, this is just a reorganization of the sources that moves specific implementations (e.g. GHDL's VHPIDIRECT) to a separate repository.


umarcor:
A section about 'External VHDL API' is added to the docs. It includes a brief description, a diagram and a note referring users to VUnit/cosim.

At the same time, a step is added to workflow 'docs' so that the artifacts of the PRs are available as a zipfile. This approach might be extended in the future, to create/update a subdir in the `gh-pages` branch of a dummy repo. It would allow to browse them directly.

PR #610, opened 2020-01-09
Add get_simulator_name() to ui

LukasVik:
In my `run.py` I want to do different things depending on what simulator the user is running. That might be things like using different locations for precompiled simlib, not adding some sources when using ghdl, etc.

This PR is a suggestion for a simple interface that solves my use case. Feel free to provide feedback.

### DIFF
--- a/tests/unit/test_ui.py
+++ b/tests/unit/test_ui.py
@@ -1193,6 +1193,10 @@ end architecture;
             sorted(expected, key=lambda x: x.name),
         )
 
+    def test_get_simulator_name(self):
+        ui = self._create_ui()
+        self.assertEqual(ui.get_simulator_name(), "mock")
+
     def _create_ui(self, *args):
         """ Create an instance of the VUnit public interface class """
         with mock.patch(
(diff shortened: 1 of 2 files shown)


eine:
Hi @LukasVik! Taking https://github.com/VHDL/Compliance-Tests/blob/master/run.py as a reference, might this work?

```py
from vunit.sim_if.factory import SIMULATOR_FACTORY
SIMULATOR_FACTORY.select_simulator().name
```

kraigher:
It is probably good to have this on the public api. I avoided it before to not exclude running tests for multiple simulators in the same run. I do not think that will happen or is so important though so we can merge this.

LukasVik:
Thanks for the tip @eine. I really like you avatar. Monkey Island is great.


> It is probably good to have this on the public api.

I agree. The design pattern is probably common enough among VUnit users that it makes sense to add to the public api. While I realise we don't want to clutter the api, this makes it more convenient than the previously available solution.

Thanks for merging!

eine:
> I really like you avatar. Monkey Island is great.

Thanks! You are actually the first one that made a reference to it ;)

> this makes it more convenient than the previously available solution.

It is!

PR #425, opened 2018-12-13
Added verification components for SPI and AVMM

bjacobs1:
Pull request as requested in issue 424.(https://github.com/VUnit/vunit/issues/424#issuecomment-446905938)

PR #611, opened 2020-01-14
add SPI verification components

eine:
This PR is a subset of #425.

/cc @ttobsen


ttobsen:
Thanks for your fast reply. I will work on this asap. :+1:

ttobsen:
Even if the tests are passing, the SPI VC is pretty buggy. Seems it needs a bit time to invest to fix and test verything. Also the tests are not covering expecting functionality.

PR #614, opened 2020-01-21
WIP: Add sync point pkg

LarsAsplund:
Adds a synchronization point mechanism


vunit/vhdl/run/src/sync_point_db_pkg.vhd
```diff
@@ -0,0 +1,361 @@
...
+  impure function new_sync_point(
+    id : id_t := null_id;
+    logger : logger_t := null_logger;
+    checker : checker_t := null_checker
+  ) return sync_point_t;
+  impure function valid(signal sync_point : sync_point_t) return boolean;
+  impure function get_id(signal sync_point : sync_point_t) return id_t;
+  impure function has_members(signal sync_point : sync_point_t) return boolean;
+  impure function is_member(signal sync_point : sync_point_t; id : id_t) return boolean;
+  procedure join(signal sync_point : inout sync_point_t; id : in id_t);
```
LarsAsplund: I went for join/leave since that made sense when talking about sync points having groups that processes join and/or leave. Now that groups are not part of the public API I think we should switch to add/remove or similar. The reason is that join may be confusing for those used to fork/join construct where join has quite the opposite meaning

vunit/vhdl/run/src/id_pkg.vhd
```diff
@@ -0,0 +1,112 @@
...
+
+  type id_t is record
+    p_data : integer_vector_ptr_t;
+  end record id_t;
+
+  constant null_id : id_t := (p_data => null_integer_vector_ptr);
+
+  type id_vec_t is array (integer range <>) of id_t;
+
+  impure function new_id(name : string := "") return id_t;
```
LarsAsplund: Eventually this should be reused/merge with the name handling of the logger. One step in that direction is to only have a `get_id` which returns an existing ID or creates a new.

vunit/vhdl/run/src/sync_point_db_pkg.vhd
eine: What about enter/leave or attach/detach?

LarsAsplund: I like enter/leave. Feels "lighter" than attach/detatch

GlenNicholls: What about enter/exit?

PR #616, opened 2020-01-23
fix parsing adjacent hyphens in a literal

ktbarrett:
Addresses #529 by improving the comment removal regular expression. But this time it actually works!

Will rebase into a single commit tonight.

### DIFF
--- a/tests/unit/test_vhdl_parser.py
+++ b/tests/unit/test_vhdl_parser.py
@@ -538,6 +538,10 @@ record
     def test_remove_comments(self):
         self.assertEqual(remove_comments("a\n-- foo  \nb"), "a\n        \nb")
 
+    def test_issue_529(self):
+        stimulus = "signal a : std_logic_vector(3 downto 0) := \"----\";"
+        self.assertEqual(remove_comments(stimulus), stimulus)
+
     def parse_single_entity(self, code):
         """
         Helper function to parse a single entity
--- a/tests/unit/test_vhdl_parser.py
+++ b/tests/unit/test_vhdl_parser.py
@@ -538,8 +538,8 @@ record
     def test_remove_comments(self):
         self.assertEqual(remove_comments("a\n-- foo  \nb"), "a\n        \nb")
 
-    def test_issue_529(self):
-        stimulus = "signal a : std_logic_vector(3 downto 0) := \"----\";"
+    def test_two_adjacent_hyphens_in_a_literal(self):
+        stimulus = 'signal a : std_logic_vector(3 downto 0) := "----";'
         self.assertEqual(remove_comments(stimulus), stimulus)
 
     def parse_single_entity(self, code):
--- a/vunit/vhdl_parser.py
+++ b/vunit/vhdl_parser.py
@@ -1131,7 +1131,8 @@ class VHDLReference(object):
         return self.name_within == "all"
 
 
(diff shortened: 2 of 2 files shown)


kraigher:
I am ready to merge this if the checks pass. Do you need any help on how to run the checks locally?

kraigher:
I saw that you pushed a fix the same second I wrote the comment, good :-)

ktbarrett:
@kraigher I do not have the current master or black locally to check. I looked at the checks to see where it failed. If it is more work than that, I will have to wait until tonight.

kraigher:
You should be able to run black using just:

```console
> python -m pip install tox
> python -m tox -e py37-fmt
```
Where you can replace py37 which whatever version you have.

ktbarrett:
> `> python -m pip install tox`

I can't download anything where I am. If you want this in today, feel free to make any changes.

eine:
@ktbarrett, if this is just an issue with running black once, I'll pick it later. Otherwise, I'll push the format fix and leave it open.

eine:
I pushed linter fixes, but CI jobs seems to be stalled regardless. These are the last three attempts:

- https://github.com/VUnit/vunit/runs/405634137
- https://github.com/VUnit/vunit/runs/405693815
- https://github.com/VUnit/vunit/runs/405868716

ktbarrett:
@eine Thanks.

I'm home now. How do I recreate the environment the regressions run in? I'm assuming there is some assumed set of tools the regression work with. Just running tox locally leads to all the acceptance tests failing. And is there anyway to increase the verbosity of the tests while they are running? What I have now is not enough to go forward.

eine:
> I'm home now. How do I recreate the environment the regressions run in? I'm assuming there is some assumed set of tools the regression work with. Just running tox locally leads to all the acceptance tests failing.

The environment just requires a simulator (GHDL is used in the CI workflow) and `tox`. Then, execute `tox -e py38-acceptance-ghdl` in the root of the repo. If you have docker/podman, you can use the following command to run a container  (instead of installing GHDL on your host):

```sh
docker run --rm -tv $(pwd):/src -w /src ghdl/vunit:llvm bash -c "pip3 install to; tox -e py38-acceptance-ghdl"
```

>  And is there anyway to increase the verbosity of the tests while they are running? What I have now is not enough to go forward.

Tests are in subdir `tests`. However, I suggest to run all of them, and then copy the full name of the ones you want to further test. In `tox.ini` you will find the `pytest` command. For example:

```sh
pytest -v tests/acceptance/test_artificial.py::TestVunitArtificial::test_artificial_verilog
```

However, it seems to be failing when testing external run scripts. These are, indeed, the examples. Hence, you can first try to execute each example individually.

ktbarrett:
The old pattern was correct, but *really* slow, so I optimized. I'm running raw times master vs this PR to see the effect on performance.

eine:
All tests pass and there seems to be no time penalty any more. However, no specific test is added. That'd be interesting  to prevent future regressions. Anyway, @ktbarrett, I'm ok with merging if you confirm that the examples you were trying in #529 do work now.

ktbarrett:
@eine I have confirmed that these fix my issues locally. I can add a unit test to the regression as well.

eine:
Thanks. Please, go ahead on this same PR.

PR #620, opened 2020-02-06
Fix "ghdl.flags" error in documentation

LukasVik:
I believe this must be a copy-paste error. "ghdl.flags" is for compilation options, and is documented as such further up in the file. "ghdl.elab_flags" and "ghdl.sim_flags", which are for simulation options are documented below.

### DIFF
--- a/docs/py/opts.rst
+++ b/docs/py/opts.rst
@@ -86,10 +86,6 @@ The following simulation options are known.
 ``pli``
   A list of PLI file names.
 
-``ghdl.flags``
-   Extra arguments passed to ``ghdl --elab-run`` command *before* executable specific flags.
-   Must be a list of strings.
-
 ``incisive.irun_sim_flags``
    Extra arguments passed to the Incisive ``irun`` command when loading the design.
    Must be a list of strings.


eine:
Hi @LukasVik! I agree with fixing the location of that flag, however, I think that your interpretation might not be correct. I'd say that `ghdl.flags` is used both for compilation and simulation. Then, you can have extra elab flags and extra run flags. Hence, I think that the flag in section "Simulation options" should be moved down, together with other `ghdl.*` flags in the same section. But it should not be removed. Furthermore, it might be desirable to change the description so that "..passed to `ghdl -a` and `ghdl --elab-run` commands..." is said.

eine:
Upon inspection of the sources (https://github.com/VUnit/vunit/blob/master/vunit/sim_if/ghdl.py) I was wrong and your interpretation is correct. I must have been mislead by some wip I have somewhere... So, I'm ok with with this PR.

@LarsAsplund, @kraigher, should we consider to rename `ghdl.flags` to, say, `ghdl.analysis_flags` in order to avoid this ambiguity. This would be breaking a change, so we might keep `ghdl.flags` as an alias.

LukasVik:
> Upon inspection of the sources (https://github.com/VUnit/vunit/blob/master/vunit/sim_if/ghdl.py) I was wrong and your interpretation is correct. I must have been mislead by some wip I have somewhere... So, I'm ok with with this PR.

Yeah it threw me for a loop as well. Had to inspect the code to find what was what.

> should we consider to rename ghdl.flags to, say, ghdl.analysis_flags in order to avoid this ambiguity. This would be breaking a change, so we might keep ghdl.flags as an alias.

Makes sense. The current naming is confusing, as proven by this conversation.

kraigher:
If we change the name the old name must still work but have a deprecation message for many releases before we remove it.

PR #621, opened 2020-02-07
Fix for RUNTIME_0232 and RUNTIME_022 messages in Riviera-PRO

OyvindNNL:
When I run a simulation with VUnit in Riviera-PRO GUI using Avalon-MM BFMs from Altera I get these messages repeated every 32 ns (dependent upon clock rate):
`Warning: RUNTIME_0232 altera_avalon_mm_master_bfm.sv (607): Non-existent queue entry. Returning default value.`

These messages can be disabled by `set sv_container_non_existent_entry_verbose 0` in startup.tcl.
However, since VUnit loads the simulation from a procedure, this global variable needs to visible in the `vunit_load` function.

[https://www.aldec.com/resources/manuals/Riviera-PRO/tcl_predefvar.htm](https://www.aldec.com/resources/manuals/Riviera-PRO/tcl_predefvar.htm):

sv_container_non_existent_entry_verbose | Setting this variable to 0 disables the RUNTIME_0232 and RUNTIME_0222 messages   that occur when referring to a not existing entry in a SystemVerilog dynamic array,   associative array, or queue. The messages are printed by default.
-- | --


kraigher:
Why cannot this variable be declared global in the user side script? We want to avoid specific tool settings inside VUnit since it is not sustainable.

OyvindNNL:
The variable is declared in the startup.tcl located in the scripts folder of the Riviera-PRO installation as a global variable. I suspect the vsim command uses upvar to get the global variables instead of the global command. This means no global variables are available to the vsim command since it is called from a procedure in VUnit.

I get your point about not wanting to known and handle all the different tool specific settings. An alternative could be to use `uplevel vsim` instead of `eval vsim` in the `vunit_load` procedure, thereby allowing for the vsim command to run in the global variable context. This way you could also remove the `global aldec` and `global LICENSE_QUEUE` commands from `vunit_load`

I did a quick test and it worked fine with `uplevel vsim`, but I will run through the offical tests later when I have some time.

kraigher:
uplevel vsim sounds like a very good solution

OyvindNNL:
I ran a few more tests, including the official ones, and saw no problems for Riviera-PRO. I made it `uplevel #0 vsim` instead to make sure it runs in the global variable context regardless of the current procedure nesting.

vunit/sim_if/rivierapro.py
```diff
@@ -284,15 +284,13 @@ def _create_load_function(
...
         tcl = """
 proc vunit_load {{}} {{
-    # Make the variable 'aldec' visible; otherwise, the Matlab interface
-    # is broken because vsim does not find the library aldec_matlab_cosim.
-    global aldec
-    # Make the variable 'LICENSE_QUEUE' visible (if set); otherwise vsim
-    # will not wait for simulation licenses.
-    global LICENSE_QUEUE
-
+    # Run the 'vsim' command in the global variable context. This will make
```
eine: I feel this paragraph is now harder to understand than https://github.com/VUnit/vunit/blame/2bb4d331b629896d3941733454d533075831222a/vunit/sim_if/rivierapro.py#L290. I'd suggest:

```
# ... Otherwise:
# - The Matlab interface is broken because vsim does not find the
#   library aldec_matlab_cosim.
# - vsim will not wait for simulation licenses.
```

OyvindNNL: That is more readable. I have made the changes.

eine:
@OyvindNNL, it seems that you merged the changes into your master, instead of rebasing. As a result, this PR is difficult to read now. I pushed a rebased version to https://github.com/eine/vunit/commits/fix-rivierapro.

Apart from that, @LarsAsplund can this PR be merged?

dalex78:
Replaced by (cleaner) pull request #825 

@LarsAsplund, I suggest to close/refuse this merge request. I believe that there are no linked issue.

PR #622, opened 2020-02-08
Automatic Creation and Loading of wave.do files

nfrancque:
This pull request adds vunit_wave to vsim simulators.  vunit_wave can recursively add groups of waves to the wave window for the user and optionally store them off to a file in the testbench directory.

If a wave.do has previously been created there, vunit will automatically load it in new runs.

The generated wave.do is structured that if it is loaded in a vunit GUI signals will be added to wave, otherwise they will be logged.  This also enables wlf generation of the same signals of interest for batch runs.  Running that wave.do when loading up the wlf afterwards will also result in the groups being regenerated.


nfrancque:
I should point out that this will slow down batch runs, so maybe a command line argument should be added to enable / disable signal logging.

eine:
> I should point out that this will slow down batch runs, so maybe a command line argument should be added to enable / disable signal logging.

Yes, please.

As another quick comment, I don't feel comfortable with hardcoding such a long TCL file. Is there any reason not to save it as a separate file?

Note that I didn't look into this PR in-deep. In need to finish other tasks first.

nfrancque:
>As another quick comment, I don't feel comfortable with hardcoding such a long TCL file. Is there any reason not to save it as a separate file?

Had that thought as well, but wasn't sure where it would go.  It would seem odd to have only this tcl proc in a separate file and none of the others.  Could it be worth having a tcl folder under sim_if and storing each procedure as a file?

eine:
> Had that thought as well, but wasn't sure where it would go. It would seem odd to have only this tcl proc in a separate file and none of the others. Could it be worth having a tcl folder under sim_if and storing each procedure as a file?

Unless a relevant performance penalty is introduced (which it should not), I think that's a great idea. Precisely, subdir `sim_if` was created a few months ago, in order to move some of the sources from `vunit`. Hence, having a `vunit/sim_if/procedures` or `vunit/sim_if/tcl` subdir is coherent with that modularisation effort.

nfrancque:
That would also help avoid some of the weird escaping rules that crop up when writing tcl in Python strings.

I would propose have separate tcl files for each, starting with this one, and interpreting them as raw f strings in python.  Then you can write regular tcl with syntax highlighting and the escaping stuff can be handled automatically, and the f strings make reading the tcl file by itself make more sense.

eine:
> We'd need to use some custom thing for the f strings, i.e. {{{var_name}}} in the tcl file and pass that through a function that does the escaping of all of tcl's regular old {}'s.

Good point. I'm not familiar with using templating features with TCL sources. Instead, I normally write a plain simple file with constants the then use them in the actual TCL source. Anyway, I don't think this is acceptable for some of the smaller snippets we have. How about `{- -}` or `{{- -}}` instead of `{{{ }}}`. I think I've seem them used in both Jinja2 (Sphinx themes) and Hugo, although I believe that `-` is optional/ignored. In this case, we can use that as a requirement. Does `{{- ` have any meaning in TCL? Note the space.

nfrancque:
Can you check out newest commit?

Note that I am using <<>>.  It seems to grab the eye best as it it's the most different from what you see in tcl files usually.

Also, to get this to work it appears you have to wrap the tcl read in an eval every time.  Not seeing a workaround.  Thoughts?

nfrancque:
In hindsight the internet seems pretty adamant that using eval is a very bad idea.  I have refactored to explicitly take in the variables each needs and call format on it.  Seems like a good middle ground.

eine:
@nfrancque, can you please rebase to have the 'docs' job fixed? BTW, you can yun `tox -e py38-lint` locally.

@LarsAsplund, can you test this in Riviera?

nfrancque:
Should be good on checks now.  Unsure about Riviera, if someone can send example commands I can implement.  Research indicates what I have will probably not work.

smenzel:
unless I am missing something, does this implementation always assume that the wave file is named 'wave.do' in the testbench folder?
In my typical project folder structure I have multiple testbench files in one folder. I should be able to load and save a dedicated wave file for each of them (or even better separately per testcase if needed).

nfrancque:
Yes it does, I hadn't anticipated multiple testbenches in one folder. How do you think we could please all use cases here? By testbench would be simple, by testcase maybe could store that information in the same wave file. 

I did not implementing anything like editing, so you can specify multiple things at the same hierarchy. That is a big limitation that is already annoying me, so maybe when I write that can also have optional sublists for test cases

nfrancque:
Yes it does, I hadn't anticipated multiple testbenches in one folder. How do you think we could please all use cases here? By testbench would be simple, by testcase maybe could store that information in the same wave file. 

I did not implementing anything like editing, so you can specify multiple things at the same hierarchy. That is a big limitation that is already annoying me, so maybe when I write that can also have optional sublists for test cases

eine:
@nfrancque, I removed the duplicated and incomplete first comment of today. Hope it's ok.

It's hard to see the suggested changes in the current state, however, we already support compilation/simulation options to be defined for all the tbs, for a lib, for some specific tb, or for a test in a tb. Hence, I think that this option should be handled in a similar fashion.

smenzel:
My suggestion would be to save the wave files always with a filename equal to the testbench name. The automatic loading can then look for a tbname.do file (or tbname_wave.do to make it more clear what it is) and fall back to a wave.do file if the more specific filename does not exist.
It would also be nice to have a config option (in run.py, not in the tcl command) to specify the location of wave files (for people that like to keep simulator specific files seperate from hdl sources)

nfrancque:
@eine  brought up making it an init_file option - which is sort of already is anyways - to override the location.

I think we can do both, but unsure how to specify testbench vs. testcase level.  Something like a `-level {tb,tc}`?  

IMO it's important to keep the command as simple as possible since no one wants to learn another set of arguments inside the simulator.

GlenNicholls:
@nfrancque @eine Can't figure out how to attach these files, so I'll paste the code since this might be useful for the objective of this issue:

sourceSimulator.tcl
```
  
set shell_command "git rev-parse --show-toplevel"
set git_root [ exec sh -c $shell_command ]
set pathToScripts [ concat $git_root/fpga/scripts/sim/src ]

if { [info exists ::env(VUNIT_SIMULATOR)] } {
    set simulator $::env(VUNIT_SIMULATOR)

    if {"$simulator" == "ghdl"} {
        source [concat $pathToScripts/gtkAddWave.tcl]
    } elseif {"$simulator" == "modelsim"} {
        source [concat $pathToScripts/msimAddWave.tcl]
    } elseif {"$simulator" == "activehdl" || "$simulator" == "rivierapro"} {
        source [concat $pathToScripts/aldecAddWave.tcl]
    } else {
        puts "ERROR: Simulator in path is not supported!"
    }
} else {
    puts "ERROR: VUNIT_SIMULATOR env variable is not in path!"
}
```

msimAddWave.tcl
```
# File: <msimAddWave.tcl>
# Author: Glen.Nicholls
# Design: sim helpers
# Date: 15 Mar 2019
#
# Description:
#   This script is the top level that can be used to easily add
#   waveforms to ModelSim waveform viewer. To add all signals in
#   each entity, there are two methods:
#
#   depth:
#       This mode allows user to specify desired depth. The script
#       will parse all signals and add every signal inside each hierarchy
#       that is <= user specified depth
#
#   Name:
#       The user can specify path to the desired entity and the script
#       will add all signals inside that entity and group them inside the
#       name of the path

# procedure::_setListUnique
proc _setListUnique {list} {
  array set included_arr [list]
  set unique_list [list]
  foreach item $list {
    if { ![info exists included_arr($item)] } {
      set included_arr($item) ""
      lappend unique_list $item
    }
  }
  unset included_arr
  return $unique_list
}


# procedure::_sortList
proc _sortList { names } {
  return [ lsort -increasing [ _setListUnique $names ] ]
}


# procedure::_getAllWaves
#
proc _getAllWaves {} {
  return [ find signals * -r ]
}


# procedure::_displayWaves
#
proc _displayWaves { groups } {
  # organize list and remove duplicates
  foreach { group_name } [ _sortList $groups ] {
    set CMD "add wave -noupdate -group $group_name $group_name/*"
    echo $CMD
    eval $CMD
    eval "config wave -signalnamewidth 1"
  }
}


# procedure::addWavesByDepth
#
proc addWavesByDepth { depth } {
  # create dict key/val
  set group_names [ list ]

  # parse list to grab groups we want to add to waveform viewer
  foreach { path } [ _getAllWaves ] {
    # split path and determine how many elements there are
    set elements     [ split $path / ]

    # remove first element because redundant and last because this is the signal name, then join for matching
    set group_elements [ lreplace $elements end end ]
    set group_name     [ join $group_elements "/" ]

    # determine number of elements
    set num_elements [ llength $group_elements ]
    incr num_elements -1

    if {$num_elements <= $depth} {
      lappend group_names $group_name
    }
  }
  _displayWaves $group_names
}


# procedure::addWavesByNames
#
proc addWavesByName { names_list } {
  # create dict key/val
  set group_names [ list ]

  # parse list to grab groups we want to add to waveform viewer
  foreach { path } [ _getAllWaves ] {
    # split path and determine how many elements there are
    set elements     [ split $path / ]

    # remove first element because redundant and last because this is the signal name, then join for matching
    set group_elements [ lreplace [ lreplace $elements 0 0 ] end end ]
    set group_name     [ join $group_elements "/" ]

    # if we found match, append path to group key
    # search add_waves to see if we have group name match that is case insensitive
    set names_lower  [ string tolower $names_list ]
    set group_lower  [ string tolower $group_name ]
    set is_string_match [lsearch -exact $names_lower $group_lower ]

    if {$is_string_match >= 0} {
      lappend group_names $group_name
    }
  }
  _displayWaves $group_names
}
```

gtkAddWave.tcl
```
##########################################################################################
#
# Description:
#   This script is used to add signals to the GTKWave waveform
#   viewer recursively.
#


# procedure::_getAllWaves
#
proc _getAllWaves {} {
  # init internals
  set paths [list]

  # get number of waves
  set num_waves  [ gtkwave::getNumFacs ]

  # search entire waveform list and remove duplicates
  for {set idx 0} {$idx < $num_waves} {incr idx} {
    lappend paths [ gtkwave::getFacName $idx ]
  }
  return $paths
}


# procedure::_checkWaveMatch
# TODO: bug in below where it allows duplicates
# global old_path
# set old_path 1
# proc _isWaveDuplicate { path } {
#   set elements [ split $path \[ ]
#
#   if { [ llength $elements ] > 1 } {
#     set var [ lreplace $elements end end ]
#     set path [ join $var \[ ]
#   }
#
#   # make sure we aren't adding duplicate wave
#   # gtk will add vector multiple times if numerous
#   # single vector elements added.
#   if { [ global old_path ] ne $path} {
#     set old_path $path
#     return 0
#   } else {
#     set old_path $path
#     return 1
#   }
# }


# procedure::_displayWaves
#
proc _displayWaves { dictWaveGroups } {
  set first 0

  foreach {k v} $dictWaveGroups {
    set add_wave [ gtkwave::addSignalsFromList $v ]

    # GTK doesn't highlight the first set of signals added so nothing gets grouped
    if {$first == 0} {
      gtkwave::/Edit/Highlight_All
    }
    gtkwave::/Edit/Create_Group "$k"
    gtkwave::setLeftJustifySigs on
    gtkwave::/Edit/Toggle_Group_Open|Close
    gtkwave::/Edit/UnHighlight_All

    # TODO: figure out how to determine signal type so we can set data format
    #set flags [ gtkwave::getTraceFlagsFromName {$v} ]
    #puts "$flags"
    incr first
  }
}


# procedure::addWavesByDepth
#
proc addWavesByDepth { depth } {
  set dictGroup [ dict create ]
  set old_path 1

  foreach { path } [ _getAllWaves ] {
    # split path and determine how many elements there are
    set elements     [ split $path . ]
    set num_elements [ llength $elements ]
    incr num_elements -2

    # remove first element because redundant and last because this is the signal name, then join for matching
    set group_elements [ lreplace [ lreplace $elements 0 0 ] end end ]
    set group_name     [ join $group_elements "." ]

    # if we found match, append path to group key
    if {$num_elements <= $depth} {
      set elements [ split $path \[ ]

      if { [ llength $elements ] > 1 } {
        set var [ lreplace $elements end end ]
        set path [ join $var \[ ]
      }

      if {$old_path ne $path} {
        set key $group_name
        set val $path
        dict lappend dictGroup $key $val
        set old_path $path
      }
    }
  }
  _displayWaves $dictGroup
}


# procedure::addWavesByName
#
proc addWavesByName { names_list } {
  set dictGroup [ dict create ]
  set old_path 1

  foreach { path } [ _getAllWaves ] {
    # split path and determine how many elements there are
    set elements     [ split $path . ]
    set num_elements [ llength $elements ]

    # remove first element because redundant and last because this is the signal name, then join for matching
    set group_elements [ lreplace [ lreplace $elements 0 0 ] end end ]
    set group_name     [ join $group_elements "." ]

    # if we found match, append path to group key
    # search add_waves to see if we have group name match that is case insensitive
    set is_string_match [lsearch -exact [ string tolower $names_list ] $group_name]

    if {$is_string_match >= 0} {
      set elements [ split $path \[ ]

      if { [ llength $elements ] > 1 } {
        set var [ lreplace $elements end end ]
        set path [ join $var \[ ]
      }

      if {$old_path ne $path} {
        set key $group_name
        set val $path
        dict lappend dictGroup $key $val
        set old_path $path
      }
    }
  }
  _displayWaves $dictGroup
}
```

Now, in my `run.py` file, I would do this:

```
vu.set_sim_option('modelsim.init_file.gui', 'addWave.tcl')
```

This sources `addWave.tcl` that is located in the same directory as `run.py` and this file simply sources the simulator specific file to source the desired procedures. Then, this file simply calls those procedures to either add all waves by hierarchy depth or by signal path-name. This prevents the user from having to write any vendor specific code:

addWave.tcl
```
set shell_command "git rev-parse --show-toplevel"
set git_root [ exec sh -c $shell_command ]

source [ concat $git_root/fpga/scripts/sim/src/sourceSimulator.tcl ]

# add all signals in each entity down to user specified hierarchy depth
set depth 3
addWavesByDepth $depth

# add all signals in each entity matching user specified path
#set add_waves_list [list]
#lappend add_waves_list tb_nameHere
#lappend add_waves_list tb_nameHere.entityHere
#addWavesByName $add_waves_list
```

`addWave.tcl` finds the repository root and uses that to find the location of the scripts. However, if these were built into VUnit, then all the user would have to do is pass some function the signal names or hierarchy depth and the framework would take care of the rest

I am working with Aldec for the next few weeks, so I can write that script as well. That should work for Riviera PRO, but I'm not positive on that.

PR #624, opened 2020-02-11
rename compile option 'ghdl.flags' to 'ghdl.a_flags'

eine:
As commented in #620, apart from renaming the option, the deprecated `ghdl.flags` is kept as an *alias* and a warning is shown. This should be kept for many releases, because it is a very used feature and the warning might be overlooked easily.


LukasVik:
Looks good!

Issue #437, opened 2019-01-17
ghdl code coverage in multi tests


qarlosalberto:
Hello,

I am trying to do code coverage with multiple generated tests with ghdl (gcc backend). I have a problem with gcov outputs. One test overwrites the gcov outputs of another.

I have found a solution, change the environment variable for each test. It is possible?

https://gcc.gnu.org/onlinedocs/gcc/Cross-profiling.html

kraigher:
Each test run is assigned its own output path even if it is a generated test. In the cases you mention GHDL might need an extra flag to make the coverage and psl artifacts end up in the output path. Alternatively running the GHDL command with the output path as the working directory. The relevant file is ghdl_interface. py if you want to make a change.

qarlosalberto:
Thanks for answering.

For psl report I have ```ui.set_sim_option("ghdl.sim_flags"      , ["--psl-report=psl_coverage.json"])```
How can I pass the output path?

kraigher:
You cannot access the output path and use it in a sim option currently.

PR #627, opened 2020-02-16
add code coverage support for ghdl

ludli505:
This adds:
* Handle for letting the user know if the version of ghdl supports coverage (i.e. if it uses the gcc back-end)
* Puts the compile-time gcno with the compiled .o-files, and run-time gcda folders within the output folder for each test

Each test case produces a separate .gcda file, and it is then left to the user to merge them as desired. This is in line with what is desired in issue #437.

This change will break the workflow of those who expect the output to just appear in the default location. For compilation these users would currently have problems with the gcno files not being updated at recompile (which is the main reason for this change). For simulation, maybe I should change the behaviour so that vunit only controls the run-time .gcda if the "enable_coverage" flag is set?

Feel free to bash on the code, I'm not a Python wizard and think it needs some feedback :-)

### DIFF
--- a/examples/vhdl/coverage/run.py
+++ b/examples/vhdl/coverage/run.py
@@ -26,9 +26,6 @@ LIB.set_compile_option("rivierapro.vcom_flags", ["-coverage", "bs"])
 LIB.set_compile_option("rivierapro.vlog_flags", ["-coverage", "bs"])
 LIB.set_compile_option("modelsim.vcom_flags", ["+cover=bs"])
 LIB.set_compile_option("modelsim.vlog_flags", ["+cover=bs"])
-LIB.set_compile_option(
-    "ghdl.a_flags", ["-fprofile-arcs", "-ftest-coverage"]
-)
-LIB.set_sim_option("ghdl.elab_flags", ["-Wl,-lgcov"])
+LIB.set_compile_option("enable_coverage", True)
 
 VU.main(post_run=post_run)
--- a/examples/vhdl/coverage/run.py
+++ b/examples/vhdl/coverage/run.py
@@ -6,10 +6,13 @@
 
 from pathlib import Path
 from vunit import VUnit
+from subprocess import call
 
 
 def post_run(results):
     results.merge_coverage(file_name="coverage_data")
+    if VU.get_simulator_name() == "ghdl":
+        call(["gcovr", "coverage_data"])
 
 
 VU = VUnit.from_argv()
@@ -17,10 +20,15 @@ VU = VUnit.from_argv()
 LIB = VU.add_library("lib")
 LIB.add_source_files(Path(__file__).parent / "*.vhd")
 
+LIB.set_sim_option("enable_coverage", True)
+
 LIB.set_compile_option("rivierapro.vcom_flags", ["-coverage", "bs"])
 LIB.set_compile_option("rivierapro.vlog_flags", ["-coverage", "bs"])
 LIB.set_compile_option("modelsim.vcom_flags", ["+cover=bs"])
 LIB.set_compile_option("modelsim.vlog_flags", ["+cover=bs"])
-LIB.set_sim_option("enable_coverage", True)
+LIB.set_compile_option(
+    "ghdl.a_flags", ["-g", "-O2", "-fprofile-arcs", "-ftest-coverage"]
+)
+LIB.set_sim_option("ghdl.elab_flags", ["-Wl,-lgcov"])
 
 VU.main(post_run=post_run)
--- a/examples/vhdl/coverage/run.py
+++ b/examples/vhdl/coverage/run.py
@@ -27,7 +27,7 @@ LIB.set_compile_option("rivierapro.vlog_flags", ["-coverage", "bs"])
 LIB.set_compile_option("modelsim.vcom_flags", ["+cover=bs"])
 LIB.set_compile_option("modelsim.vlog_flags", ["+cover=bs"])
 LIB.set_compile_option(
-    "ghdl.a_flags", ["-g", "-O2", "-fprofile-arcs", "-ftest-coverage"]
+    "ghdl.a_flags", ["-fprofile-arcs", "-ftest-coverage"]
 )
 LIB.set_sim_option("ghdl.elab_flags", ["-Wl,-lgcov"])
 
--- a/vunit/sim_if/ghdl.py
+++ b/vunit/sim_if/ghdl.py
@@ -438,10 +438,7 @@ class GHDLInterface(SimulatorInterface):  # pylint: disable=too-many-instance-at
         # Add compile-time .gcno files as well, they are needed or the report
         if file_exists(dir_name):
             for library in self._project.get_libraries():
-                lib_dir_path = Path(library.directory)
-                if lib_dir_path.exists():
-                    gcno_files = lib_dir_path.glob("*.gcno")
-                    for gcno_file in gcno_files:
+                for gcno_file in Path(library.directory).glob("*.gcno"):
                         # gcda files are output to the the coverage directory, but within subdirectories
                         # taken directly from where the command was run, so appdn cwd to output here
                         shutil.copy(gcno_file, gcda_dir)
--- a/vunit/sim_if/ghdl.py
+++ b/vunit/sim_if/ghdl.py
@@ -164,11 +164,18 @@ class GHDLInterface(SimulatorInterface):
     @classmethod
     def supports_vhpi(cls):
         """
-        Return if the simulator supports VHPI
+        Returns True when the simulator supports VHPI
         """
         return (cls.determine_backend(cls.find_prefix_from_path()) != "mcode") or (
             cls.determine_version(cls.find_prefix_from_path()) > 0.36
         )
+
+    @classmethod
+    def supports_coverage(cls):
+        """
+        Returns True when the simulator supports coverage
+        """
+        return cls.determine_backend(cls.find_prefix_from_path()) == "gcc"
 
     def _has_output_flag(self):
         """
--- a/vunit/sim_if/ghdl.py
+++ b/vunit/sim_if/ghdl.py
@@ -263,6 +263,12 @@ class GHDLInterface(SimulatorInterface):  # pylint: disable=too-many-instance-at
             a_flags += flags
 
         cmd += a_flags
+
+        if source_file.compile_options.get("enable_coverage", False):
+            # Add gcc compilation flags for coverage
+            #   -ftest-coverages creates .gcno notes files needed by gcov
+            #   -fprofile-arcs creates branch profiling in .gcda database files
+            cmd += ["-fprofile-arcs", "-ftest-coverage"]
         cmd += [source_file.name]
         return cmd
 
@@ -291,6 +297,9 @@ class GHDLInterface(SimulatorInterface):  # pylint: disable=too-many-instance-at
         if self._has_output_flag():
             cmd += ["-o", bin_path]
         cmd += config.sim_options.get("ghdl.elab_flags", [])
+        if config.sim_options.get("enable_coverage", False):
+            # Enable coverage in linker
+            cmd += ["-Wl,-lgcov"]
         cmd += [config.entity_name, config.architecture_name]
 
         sim = config.sim_options.get("ghdl.sim_flags", [])
--- a/vunit/sim_if/ghdl.py
+++ b/vunit/sim_if/ghdl.py
@@ -272,7 +272,9 @@ class GHDLInterface(SimulatorInterface):  # pylint: disable=too-many-instance-at
         cmd += [source_file.name]
         return cmd
 
-    def _get_command(self, config, output_path, elaborate_only, ghdl_e, wave_file):
+    def _get_command(  # pylint: disable=too-many-branches
+        self, config, output_path, elaborate_only, ghdl_e, wave_file
+    ):
         """
         Return GHDL simulation command
         """
--- a/vunit/sim_if/ghdl.py
+++ b/vunit/sim_if/ghdl.py
@@ -14,18 +14,19 @@ import logging
 import subprocess
 import shlex
 import re
+import shutil
 from json import dump
 from sys import stdout  # To avoid output catched in non-verbose mode
 from warnings import warn
 from ..exceptions import CompileError
-from ..ostools import Process
+from ..ostools import Process, file_exists
 from . import SimulatorInterface, ListOfStringOption, StringOption, BooleanOption
 from ..vhdl_standard import VHDL
 
 LOGGER = logging.getLogger(__name__)
 
 
-class GHDLInterface(SimulatorInterface):
+class GHDLInterface(SimulatorInterface):  # pylint: disable=too-many-instance-attributes
     """
     Interface for GHDL simulator
     """
@@ -108,6 +109,7 @@ class GHDLInterface(SimulatorInterface):
         self._gtkwave_args = gtkwave_args
         self._backend = backend
         self._vhdl_standard = None
+        self._coverage_dirs = set()
 
     def has_valid_exit_code(self):
         """
@@ -355,10 +357,14 @@ class GHDLInterface(SimulatorInterface):
 
         status = True
 
-        try:
+        gcov_env = os.environ.copy()
+        if config.sim_options.get("enable_coverage", False):
             # Set environment variable to put the coverage output in the test_output folder
-            gcov_env = os.environ.copy()
-            gcov_env["GCOV_PREFIX"] = os.path.join(output_path, "coverage")
+            coverage_dir = join(output_path, "coverage")
+            gcov_env["GCOV_PREFIX"] = coverage_dir
+            self._coverage_dirs.add(coverage_dir)
+
+        try:
             proc = Process(cmd, env=gcov_env)
             proc.consume_output()
         except Process.NonZeroExitCode:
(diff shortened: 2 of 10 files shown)


ludli505:
The "36-unit" tests fails, and I don't understand why.

The failures occurred due to me putting the check_output function within the `SimulatorInterface` class, to make it accessible when running its `_compile_source_file` from within `GHDLInterface`. It would probably be better if that change could be avoided somehow.

eine:
Hi @ludli505! Incidentally, some weeks ago I tried to extend the existing coverage example, which is currently supported for non-GHDL simulators only. My motivation and work-in-progress have some subtle differences with yours, hence please bear with me:

- As discussed in ghdl/docker#11, a user requested some coverage related packages to be added to `ghdl/vunit:gcc*` images. Precisely, `lcov` and `gcovr`.
- Since this repo contains an example about how to use code coverage with other simulators, I modified it to use GHDL (see https://github.com/ghdl/docker/issues/11#issuecomment-576406937), so I could test docker images.
- Then, I tried to improve it, in order to have the test executed as part of the regular test suite. The work in progress is in branch [ghdl-coverage](https://github.com/eine/vunit/commits/ghdl-coverage) of my fork: https://github.com/VUnit/vunit/compare/master...eine:ghdl-coverage

Overall, I'd be glad if you could have a look at that WIP. Nonetheless, it would be good to split the modifications in this PR into multiple smaller commits that are easier to review. One of the targets (to detect if GHDL's backend supports coverage) is common to both implementations; while moving the results is considered here only.

Regarding the issue with modifying the `mock`, I think a better approch would be to handle moving the results in: https://github.com/VUnit/vunit/blob/a45f4d4791c35a03f722b07a7a588ce2e82fd685/vunit/sim_if/ghdl.py#L275-L325

as this seems to be specific to GHDL. Is it expected to be used with other simulators too?

ludli505:
Thanks for your feedback! I'll look through your WIP later today, and split into multiple commits.

I've found two issues with my changes:

__Problem 1__
Creating one output .gcda per test and source-file means that merging taking insanely long time, in my case ~20 minutes for less than 100 tests. The motivation for creating one output per test case was to give the user as much control as possible, but perhaps it is not worth it. 

The alternative is to have gcov merge all the results while running the tests. If the user has specific requirements on which coverage to merge they'd have to make sure to clean any previous test outputs and run the specific tests desired. For me personally it seems better this way than to have the merging be super-slow. Do you agree that I should change it?

__Problem 2__
If we have a source file `axi_pkg.vhd`, the resulting .gcda file is `axi_pkg.gcda`. This causes a conflict when there are multiple files named `axi_pkg.vhd` (which is the case if using vunit and tsfpga). The solution is to use the `-p` or `-pg` flag to get full paths in the filename, but unfortunately it is not accepted by GHDL. It seems like the option is commented out in a Makefile when building GHDL with gcc. I'll check if this can be changed.

qarlosalberto:
@ludli505  it's really interesting, I use code coverage in GHDL regularly. And we support it in @TerosTechnology

https://youtu.be/tgr1KGIitIQ?t=170

I have the same problem that you with the output folders. If I could choose (regardless of performance):

1. One output .gcda per test.
2. Merged .gcda
3. Code coverage % per test in output result file.
4. Code coverage % total with merged results in output result file.

Actually I do --clean when I want to regenerate the code coverage. But it's really slow when I have to compile a lot of libraries...

ludli505:
@eine:
* I've copied your commit regarding the `supports_coverage` function.
* Regarding the `mock`: the results are already moved in the chunk you refer to. The compile-time .gcno are moved in an overloaded version of `compile_output`. I've cleaned this up a bit now.
* Messing around with environment variable seems dangerous. I'd be grateful if you know whether this
[this](https://github.com/VUnit/vunit/pull/627/commits/2b9263ae708d134ccbab8d773cdeeea3e1922838#diff-2d4a56f7b7c79d9df2c628aabbe64e43R317-R321) is safe or not when running multiple simulation threads.
* It seems like I should add a merge_coverage procedure as well, and expand the example/test as you have done on your branch.

@qarlosalberto:
* The slow merging was due to my stupid. Merging with `gcov_tool` seems fast.
* I'd be happy to implement a merge_coverage function, when I know these changes are on the right track :-)

ludli505:
I've now merge_coverage for ghdl. It puts merges the gcda files from all tests and adds them together with the gcno files in the correct folder, so that they can be easily handled with `gcovr`

examples/vhdl/coverage/run.py
```diff
@@ -6,21 +6,29 @@
...
+
 LIB.set_compile_option("rivierapro.vcom_flags", ["-coverage", "bs"])
 LIB.set_compile_option("rivierapro.vlog_flags", ["-coverage", "bs"])
 LIB.set_compile_option("modelsim.vcom_flags", ["+cover=bs"])
 LIB.set_compile_option("modelsim.vlog_flags", ["+cover=bs"])
-LIB.set_sim_option("enable_coverage", True)
+LIB.set_compile_option(
+    "ghdl.a_flags", ["-g", "-O2", "-fprofile-arcs", "-ftest-coverage"]
+)
+LIB.set_sim_option("ghdl.elab_flags", ["-Wl,-lgcov"])
```
LukasVik: To follow the convention  in e.g. modelsim.py:258 this option should be moved to the ghdl interface class and be enabled based on the `enable_coverage` sim option.

Maybe `ghdl.a_flags` should be handled the same way. For riviera and modelsim the user has to manually set compile option since they have to make a decision on what coverage to enable. For ghdl that choice is not there, so `ghdl.a_flags` could be set in the class solely based on `enable_coverage` sim option.

vunit/sim_if/ghdl.py
```diff
@@ -108,6 +109,7 @@ def __init__(  # pylint: disable=too-many-arguments
         self._gtkwave_args = gtkwave_args
         self._backend = backend
         self._vhdl_standard = None
+        self._coverage_dirs = set()
```
LukasVik: To avoid confusion in `merge_coverage`, where a lot of coverage dirs are thrown around, maybe rename this to `_coverage_test_dirs`

vunit/sim_if/ghdl.py
```diff
@@ -364,3 +381,67 @@ def simulate(  # pylint: disable=too-many-locals
             subprocess.call(cmd)
 
         return status
+
+    def _compile_source_file(self, source_file, printer):
+        """
+        Runs parent command for compilation, and moves the .gcno file (if any) to the compilation output folder
```
LukasVik: Very long line. Not sure what vunit coding guidelines say but it is too long for a side-by-side diff on my screen.

vunit/sim_if/ghdl.py
```diff
@@ -364,3 +381,67 @@ def simulate(  # pylint: disable=too-many-locals
...
+            gcno_file.rename(new_path)
+
+        return compilation_ok
+
+    def merge_coverage(self, file_name, args=None):
+        """
+        Merge coverage from all test cases
+        """
+
+        dir_name = file_name
```
LukasVik: Maybe rename to `output_dir` for clarity

vunit/sim_if/ghdl.py
```diff
@@ -364,3 +381,67 @@ def simulate(  # pylint: disable=too-many-locals
...
+            new_path = Path(source_file.library.directory) / gcno_file
+            gcno_file.rename(new_path)
+
+        return compilation_ok
+
+    def merge_coverage(self, file_name, args=None):
+        """
+        Merge coverage from all test cases
+        """
+
```
LukasVik: empty line

vunit/sim_if/ghdl.py
```diff
@@ -364,3 +381,67 @@ def simulate(  # pylint: disable=too-many-locals
...
+            else:
+                LOGGER.warning("Missing coverage directory: %s", coverage_dir)
+
+        # Find actual output path of the .gcda files (they are deep in hierarchy)
+        dir_path = Path(dir_name)
+        gcda_files = dir_path.glob("**/*.gcda")
+        gcda_dirs = [x.parent for x in gcda_files]
+        gcda_dirs = set(gcda_dirs)
+        assert len(gcda_dirs) == 1, "Expected exactly one folder with gcda files"
+        gcda_dir = str(gcda_dirs.pop())
```
LukasVik: Do not need to go back to string representation here

vunit/sim_if/ghdl.py
```diff
@@ -364,3 +381,67 @@ def simulate(  # pylint: disable=too-many-locals
...
+
+        # Add compile-time .gcno files as well, they are needed or the report
```
LukasVik:
```suggestion
        # Add compile-time .gcno files as well, they are needed for the report
```

vunit/sim_if/ghdl.py
```diff
@@ -364,3 +381,67 @@ def simulate(  # pylint: disable=too-many-locals
...
+        if file_exists(dir_name):
+            for library in self._project.get_libraries():
+                lib_dir_path = Path(library.directory)
+                if lib_dir_path.exists():
+                    gcno_files = lib_dir_path.glob("*.gcno")
+                    for gcno_file in gcno_files:
+                        # gcda files are output to the the coverage directory, but within subdirectories
+                        # taken directly from where the command was run, so appdn cwd to output here
```
LukasVik: This comment seems out of place

vunit/sim_if/ghdl.py
```diff
@@ -364,3 +381,67 @@ def simulate(  # pylint: disable=too-many-locals
...
+        # Find actual output path of the .gcda files (they are deep in hierarchy)
+        dir_path = Path(dir_name)
+        gcda_files = dir_path.glob("**/*.gcda")
+        gcda_dirs = [x.parent for x in gcda_files]
+        gcda_dirs = set(gcda_dirs)
+        assert len(gcda_dirs) == 1, "Expected exactly one folder with gcda files"
+        gcda_dir = str(gcda_dirs.pop())
+
+        # Add compile-time .gcno files as well, they are needed or the report
+        if file_exists(dir_name):
```
LukasVik: We know that the folder exists based on the assert on line 435. Can remove this.

vunit/sim_if/ghdl.py
```diff
@@ -364,3 +381,67 @@ def simulate(  # pylint: disable=too-many-locals
...
+            for library in self._project.get_libraries():
+                lib_dir_path = Path(library.directory)
+                if lib_dir_path.exists():
+                    gcno_files = lib_dir_path.glob("*.gcno")
+                    for gcno_file in gcno_files:
```
LukasVik:
```suggestion
                for gcno_file in Path(library.directory).glob("*.gcno"):
```
The glob will return empty if `lib_dir_path` does not exist

ludli505: Nice

vunit/sim_if/ghdl.py
ludli505: Will fix

vunit/sim_if/ghdl.py
ludli505: Will fix

vunit/sim_if/ghdl.py
ludli505: Thanks!

vunit/sim_if/ghdl.py
ludli505: Will fix

vunit/sim_if/ghdl.py
ludli505: Vunit is ok, but I'll change it anyway

vunit/sim_if/ghdl.py
ludli505: Agreed

examples/vhdl/coverage/run.py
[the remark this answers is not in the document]
ludli505: Yeah, it's a bit obscure. These are gcc-flags and are directly forwarded by ghdl to gcc. I'll add some comments on what they do.

examples/vhdl/coverage/run.py
```diff
@@ -6,21 +6,29 @@
...
 
+LIB.set_sim_option("enable_coverage", True)
+
 LIB.set_compile_option("rivierapro.vcom_flags", ["-coverage", "bs"])
 LIB.set_compile_option("rivierapro.vlog_flags", ["-coverage", "bs"])
 LIB.set_compile_option("modelsim.vcom_flags", ["+cover=bs"])
 LIB.set_compile_option("modelsim.vlog_flags", ["+cover=bs"])
-LIB.set_sim_option("enable_coverage", True)
+LIB.set_compile_option(
+    "ghdl.a_flags", ["-g", "-O2", "-fprofile-arcs", "-ftest-coverage"]
```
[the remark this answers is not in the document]
ludli505: I agree on `-g`, and probably on `-O` as well. Optimizations may mess with coverage so I'll test this first.

examples/vhdl/coverage/run.py
ludli505: This was my original idea, but I opted not to do it as it seems like the `enable_coverage` sim option should not affect compilation.

But it would be more convenient, I'll change it.

examples/vhdl/coverage/run.py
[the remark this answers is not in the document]
ludli505: The flags have been removed. 

Note: According to [the documentation on -ftest-coverage](https://gcc.gnu.org/onlinedocs/gcc-6.2.0/gcc/Instrumentation-Options.html#Instrumentation-Options), coverage will be better when no optimizations are used.

examples/vhdl/coverage/run.py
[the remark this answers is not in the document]
ludli505: Now commented

ludli505: The sim option `-Wl,-lgcov`, has now been moved into `ghdl.py`

The compilation function only works on the `SourceFile` object, which is neat but means that the `enable_coverage` _simulation_ option is not conveniently available. I've added a new `enable_coverage` _compilation_ option. This also means that coverage may be enabled only for the desired files. Proof of concept on my [tsfpga branch](https://gitlab.com/truestream/tsfpga/-/tree/coverage_in_simulate_py)

vunit/sim_if/ghdl.py
ludli505: Done

vunit/sim_if/ghdl.py
ludli505: Done

vunit/sim_if/ghdl.py
ludli505: Done

vunit/sim_if/ghdl.py
ludli505: Fixed

vunit/sim_if/ghdl.py
ludli505: Fixed

vunit/sim_if/ghdl.py
ludli505: Removed comment

vunit/sim_if/ghdl.py
ludli505: Fixed

LukasVik:
I have reviewed it again and think it looks good to merge. It works well, both in the `coverage` example as well as in the tsfpga code base.

ludli505:
Thanks @LukasVik .

@eine, could you have a look at the changes?

vunit/sim_if/ghdl.py
```diff
@@ -347,8 +367,16 @@ def simulate(  # pylint: disable=too-many-locals
         )
 
         status = True
+
+        gcov_env = environ.copy()
```
eine: What's the motivation to copy the environment here instead of setting `environ["GCOV_PREFIX"]`? Is it because of parallel executions?

vunit/sim_if/ghdl.py
```diff
@@ -364,3 +392,60 @@ def simulate(  # pylint: disable=too-many-locals
...
+    def merge_coverage(self, file_name, args=None):
+        """
+        Merge coverage from all test cases
+        """
+        output_dir = file_name
+
+        # Loop over each .gcda output folder and merge them two at a time
+        first_input = True
+        for coverage_dir in self._coverage_test_dirs:
+            if file_exists(coverage_dir):
```
eine: Is it possible to use `Path(coverage_dir).exists()` instead?

vunit/sim_if/ghdl.py
```diff
@@ -364,3 +392,60 @@ def simulate(  # pylint: disable=too-many-locals
             subprocess.call(cmd)
 
         return status
+
+    def _compile_source_file(self, source_file, printer):
+        """
+        Runs parent command for compilation, and moves any .gcno files to the compilation output
```
eine: Is this specific to GHDL? Do other simulators place the files when desired?

vunit/sim_if/ghdl.py
```diff
@@ -364,3 +392,60 @@ def simulate(  # pylint: disable=too-many-locals
...
+        """
+        Merge coverage from all test cases
+        """
+        output_dir = file_name
+
+        # Loop over each .gcda output folder and merge them two at a time
+        first_input = True
+        for coverage_dir in self._coverage_test_dirs:
+            if file_exists(coverage_dir):
+                if first_input:
```
eine: Maybe `first_dir = coverage_dir if first_input else output_dir`? Moreover, it seems that declaring `first_dir` can be avoided.

vunit/sim_if/ghdl.py
```diff
@@ -364,3 +392,60 @@ def simulate(  # pylint: disable=too-many-locals
...
+                ]
+                subprocess.call(merge_command)
+                first_input = False
+            else:
+                LOGGER.warning("Missing coverage directory: %s", coverage_dir)
+
+        # Find actual output path of the .gcda files (they are deep in hierarchy)
+        dir_path = Path(output_dir)
+        gcda_files = dir_path.glob("**/*.gcda")
+        gcda_dirs = [x.parent for x in gcda_files]
```
eine: Maybe replace 442-444 with `gcda_dirs = set([x.parent for x in dir_path.glob("**/*.gcda")])`?

vunit/ui/__init__.py
```diff
@@ -1099,3 +1099,13 @@ def get_simulator_name(self):
         if self._simulator_class is None:
             return None
         return self._simulator_class.name
+
+    def simulator_supports_coverage(self):
+        """
+        Returns True if the simulator supports coverage
```
eine: Please, replace `if` with `when`, for consistency with other sources.

vunit/sim_if/ghdl.py
ludli505: Yes, it's because of parallel executions. GCOV_PREFIX needs different values for different tests, which is why it is set per subprocess spawned.

vunit/sim_if/ghdl.py
ludli505: Yes, I've changed this now

vunit/sim_if/ghdl.py
ludli505: I'm not sure, but think that at least Modelsim doesn't have any extra compile-time files generated.

vunit/sim_if/ghdl.py
ludli505: Thanks! I've simplified as you suggested.

vunit/sim_if/ghdl.py
ludli505: Done, thanks. Actually, replaced with set comprehension after recommendation from linter: `gcda_dirs = {[x.parent for x in dir_path.glob("**/*.gcda")}`

vunit/ui/__init__.py
ludli505: Done, thanks

ludli505:
@eine, thanks for your review. I've updated according to your comments.

> Currently example coverage is not added to acceptance tests. Would you mind adding eine@3e16fc5#diff-ec10270706c0b42d3ef658d092217a3a to this PR?

Done as well

> It seems that Problem 1 from #627 (comment) is not an issue anymore, is it? What about Problem 2?

That problem still exists. I've not looked into if this can be solved in GHDL. I consider it a separate problem.

> Overall, I think it is good to merge (after these minor issues are fixed). However, I am unsure about how to handle backwards compatibility. @LarsAsplund, @kraigher, as commented above, "this change will break the workflow of those who expect the output to just appear in the default location". Should we wait to merge this when we bump to v5.0.0 or is it ok to include it in v4.4.0?

After reading your comment, I realized that only one additional check for `enable_coverage` flag in `_compile_source_file` of `ghdl.py` was needed for backwards compatibility. I've added this now.

eine:
@ludli505, so quick! Thanks!

> That problem still exists. I've not looked into if this can be solved in GHDL. I consider it a separate problem.

Fair enough!

> After reading your comment, I realized that only one additional check for `enable_coverage` flag in `_compile_source_file` of `ghdl.py` was needed for backwards compatibility. I've added this now.

Can you please add a very brief description about this? Ideally, something should be added to the docs (in http://vunit.github.io/py/opts.html#simulation-options and/or vunit.github.io/py/vunit.html#vunit.ui.results.Results.merge_coverage). However, I think it is ok with adding a description to the commit message, at least.

ludli505:
My last commit adds description of the new `enable_coverage` flags, and updated the description for the old one in  py/opts.html.

Is that the information you're looking for or did I miss something?

Cheers!

eine:
> Is that the information you're looking for or did I miss something?

That's exactly what I was looking for. Thanks!

PR #631, opened 2020-03-04
Fix GHDL "prefix of array attribute must be an object name"

LukasVik:
Since a couple of days ago, new VUnit version and new GHDL version gives errors like

`/usr/local/lib/python3.8/site-packages/vunit_hdl-4.3.1rc0-py3.8.egg/vunit/vhdl/data_types/src/codec.vhd:520:47: prefix of array attribute must be an object name`

during compilation. This PR fixes these so that compile passes.

### DIFF
--- a/vunit/vhdl/data_types/src/codec-2008p.vhd
+++ b/vunit/vhdl/data_types/src/codec-2008p.vhd
@@ -110,7 +110,8 @@ package body codec_2008p_pkg is
   function decode (
     constant code : string)
     return boolean_vector is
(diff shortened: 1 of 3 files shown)


eine:
I'm finding this in other places: https://github.com/eine/vunit/runs/489366945?check_suite_focus=true#step:4:357 According to ghdl/ghdl#1152, we should fix it on our side. However, it is surprising for me that all checks passed when this PR was merged and only GNU/Linux tests are failing now.

PR #638, opened 2020-03-30
call supports_coverage() rather than returning method object

LukasVik:
Previously the method was returned rather than a boolean. So doing 

```python
if vunit_proj.simulator_supports_coverage():
    ...
```

in run.py would always evaluate as True.

`supports_coverage` is a static method or a class method in all the simulator interfaces. Never a property.


eine:
Thanks!

PR #660, opened 2020-06-23
RFC: Change log message time unit

ttobsen:
When using GHDL its pretty annoying that the log message time is in fs resolution. With VHDL 2008, there is a easy way to print time with a given unit (see this PR as an example).

I think it would be very nice if the time unit can be set globally, but I currently have no idea where is the right place for such a setting.

Discussions an ideas are very welcome. :-)


eine:
Overall, I think it is an interesting request/proposal. However, I'm neither familiar with the logging codebase...

Meanwhile, an alternative can be to use GHDL's own features to set the simulation time resolution. It works with `mcode` backend only, and it might not be useful if you want to log above the "fundamental" resolution. However, for this case where I assume that you want to both simulate and log using `ns`, it would work.

ttobsen:
Yes, this workaround is in general great, but in my case I'm using the GCC backend to fasten simulation runtime.

Maybe to solve this, we could think about creating something like a default vunit_config_pkg.vhd where settings (like the print time unit) are written on VHDL level. Then the user can decide if the VHDL package of the VUnit installation or the local on in the verification project should be used. It also has the advantage that the settings are part of the SCM.

And maybe it would also be better I had opened an issue instead of a PR to discuss this feature. :-(

eine:
Regarding having a default VHDL package, or handling it as a generic/parameter (so that users can define it as a constant in their testbench and use whatever mechanism to set it), let's wait for @LarsAsplund to have a thought at it. I believe he is the architect behind the logging infrastructure. The concept of providing some default VHDL sources and allowing users to "overwrite" them is already used for the "external" API of internal vectors. However, If an easier solution is possible, I think it'd be desirable.

> And maybe it would also be better I had opened an issue instead of a PR to discuss this feature. :-(

No worries at all :P. PRs are, in fact, just issues with some additional context. And in this case, the additional context is pertinent.

ttobsen:
Alright, easy is always a good option. The good thing is that VHDL2008 helps a lot to print time values easily with every unit wanted.

LarsAsplund:
This was added by https://github.com/VUnit/vunit/issues/1044 and was released in  https://vunit.github.io/release_notes.html#dev3-pre-release-2024-08-15

Issue #615, opened 2020-01-22
GUI Wave Tcl


nfrancque:
I would like to improve the wave.do loading for vunit.  This is all from a modelsim background, please let me know if there's something that would be different in another simulator.

1)

Right now I have one run.py with many testbenches, some of which are not at the same hierarchy and therefore need different wave files.  I would prefer to keep this master run.py from being edited by individual block designers 

So it would be nice to automatically search the folder containing the source file of the next testbench to run and load up wave.do if it exists - this is what modelsim does anyways, except only in current directory rather than directory of testbench.  This also makes vunit more backwards compatible with standard vsim interface.

I would expect this to be a default overridden by the init_file of the python interface, but hopefully its usage would decrease quite a bit with this default.

2)

In combination with #455, it could be nice to generate a tcl proc like
 `vunit_wave [path/to/hier/level] [depth] [-save]`

Which automatically add and group all modules at the specified hierarchy, recursively with the specified depth.  Likely default to top level with depth 1, so it adds a group for each block in the top level. Depth of 0 would recurse completely.

 -save should save the wave.do to the folder containing the source file of the currently option testbench.

I could see these both being combined so that the first time we load up the gui we run vunit_wave -save to save wave.do in the corresponding folder.

It would also be much better if the generated .do files had wildcards rather than explicitly referencing all.  This is how modelsim generates them by default so if a signal changes you have to manually modify.

kraigher:
Sure sounds good. We try to keep the same feature set for both modelsim and riviera so it would be good if it can share implementation for both.

smenzel:
I wrote a wave_init.do script that does something similar to your request:
It doesn't load a wave from the location of the testbench but from a subfolder _waves_ in the vunit folder. Here I keep the waves of all my testbenches and name them {testbench_name}_wave.do.

```
# get a lot of info about the loaded design
set data [ write report -tcl ]

# extract the name of the testbench
regexp {([A-Z_a-z0-9]*_tb).vhd} $data -> design_unit
puts "Found Design Unit: $design_unit"

# try to source wave file with appropriate name from waves folder
set wavename [pwd]/../../waves/${design_unit}
if { [file exists ${wavename}.do] } {
    source ${wavename}.do
} elseif { [file exists ${wavename}_wave.do] } {
    source ${wavename}_wave.do
} elseif { [file exists [pwd]/../../wave.do] } {
    # fallback: generic wave.do
    source [pwd]/../../wave.do
} else {
[...]

PR #665, opened 2020-07-07
Add add_waves_by_depth to UI

GlenNicholls:
added add_waves_by_depth method and TCL scripts for modelsim, aldec, and ghdl

This is in primarily in relation to #455, but aligns with my comments in #615 . I avoided the #622 PR because it is pretty far behind master and @nfrancque appeared to mostly have stuff for the ModelSim save/load of wave files which is different than what I'm looking to add. I have some outstanding questions:

* Does it make sense to add waves as I have done with `add_waves_by_depth()` which calls `set_sim_option()` under the covers?
* do I need a unit test for `add_waves_by_depth()`? If so, how would I do this since the entire interface is creating a file and adding waves to the simulator?

This PR would close out #455, but the below feature requests are my vision for #615:

* JSON interface such that user can add hierarchy path to signals/variables/constants and some options and VUnit would take care of simulator specific TCL commands
* Add bools to `add_waves_by_depth()` so user can include constants and/or variables as well.
* Add command-line option to add by depth in cases where adding `add_waves_by_depth` to `run.py` doesn't always make sense
* Add simulator GUI option for adding waves by depth (gated by #622 being accepted and merged with master)

### DIFF
--- a/vunit/sim_if/tcl/aldec_if.tcl
+++ b/vunit/sim_if/tcl/aldec_if.tcl
@@ -1,0 +1,124 @@
+# Description:
+#   This script is the top level that can be used to easily add
+#   waveforms to activeHDL and RivieraPRO waveform viewer.
+#
+#   depth:
+#       This mode allows user to specify desired depth. The script
+#       will parse all signals and add every signal inside each hierarchy
+#       that is <= user specified depth
+
+# globals because unable to call each of these commands
+# within a procedure
+find -type signal -rec * {$>} ALL_SIGNALS
+find -type instance -rec * {$>} ALL_GROUPS
+find -type const -rec * {$>} ALL_CONSTANTS
+find -type variable -rec * {$>} ALL_VARIABLES
+
+
+# procedure::_filterList
+#    filters list passed to it for junk Aldec formats it with
+proc _filterList {list} {
+  # list of items to filter out
+  set list_to_filter \
+    {"Signals:" "Constants:" "Variables:" "Instances:" "Generics and parameters:"}
+  set filt_list {}
+  # re-format list created by Aldec
+  foreach item $list {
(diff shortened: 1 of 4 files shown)


GlenNicholls:
Also, I'm unable to run tox locally so I can't figure out what's causing the fmt failure. The logs aren't giving me anything to work with, can someone let me know what I'm doing wrong style-wise?

eine:
@GlenNicholls, you can run `black` (the formatter, which is a pip package) on the single file which is complaining: https://github.com/VUnit/vunit/pull/665/checks?check_run_id=844062265#step:5:15

`black vunit/vunit/ui/__init__.py`

Calling `fmt` through `tox` does that recursively: https://github.com/VUnit/vunit/blob/master/tox.ini#L31

GlenNicholls:
@eine So why does CI fail instead of running black and pushing the change?

eine:
> @eine So why does the `fmt` CI fail instead of running black and pushing the change?

In fact, that's explicit (see arg `--check` in https://github.com/VUnit/vunit/blob/master/.github/workflows/push.yml#L28), because we don't want an automated service pushing code modifications into the main branch. We do overwrite some branches from GitHub Actions, such as `gh-pages`, but not branches with sources.

That's why `tox -e py37-fmt` exists, to allow contributors to have sources formate automatically before they push. It is unfortunate that tox does not work on your host... Maybe you can open an issue and I can have a look at it?

GlenNicholls:
>Maybe you can open an issue and I can have a look at it?

Don't worry about it, tox fails locally for other repos, I just haven't run it enough times to check since my work CI tox catches stuff on my branches before the PR is accepted. I'm assuming it's a fairly simple fix once I start digging around.

GlenNicholls:
Moved to https://github.com/VUnit/vunit/pull/690

PR #666, opened 2020-07-07
Avalon stream update

slaweksiluk:
also deleted gtkw files


vunit/vhdl/verification_components/src/avalon_stream_pkg.vhd
```diff
@@ -123,6 +129,29 @@ package body avalon_stream_pkg is
     return source.p_data_length;
   end;
 
+  function calc_empty_length(data_width : positive) return natural is
```
eine, line 132: Overall, the PR looks good. However, I'm in doubt with this function. Isn't this equivalent to log2? Otherwise, should it be placed in a different location, to be reused by other parts of VUnit?

eine:
@LarsAsplund is it ok to merge this before #520?

vunit/vhdl/verification_components/src/avalon_stream_pkg.vhd
slaweksiluk: Yes it is equivalent. However as I remember I had unexpected results due to rounding issues when used log2 and ceil/floor from ieee math library. I might applied them in wrong order or made other mistake, so will be grateful for some example.

Placing function for calculating vector width on in vunit lib is good idea overall. Its often used in test benches.

eine: Actually, I faced those issues myself too... I "fixed" them for `n-1` or for `0`, but I always have issues with some of them. Apart from that, I've seen variants of your function multiple times, but I cannot remember if it exists in VUnit's codebase already.

Let me ask @Paebbels, what is his suggested approach? Do you use `log2 + ceil/floor` or a custom function? Is there anything in the latest IEEE libs?

Paebbels: The PoC Library provides:
* `log2ceil` - 2s logarithm that rounds up
* `log2ceilnz`  - 2s logarithm that rounds up and is not zero (nz)

The latter function is sometimes called `to_bits`.

eine: Thanks @Paebbels!

@slaweksiluk, I'm thinking that we might move `calc_empty_length` to a new package in the core `vunit/vunit/vhdl/core/src/utils_pkg.vhd`. If your implementation is equivalent to either `log2ceil` or `log2ceilnz` from PoC, I think we should go with that (not adding PoC as a ref, but using the same name).

LarsAsplund: There is also a similar function [here](https://github.com/VUnit/vunit/blob/master/vunit/vhdl/random/src/random_pkg.vhd#L118). I'm for a "utility package" but I would avoid using that word since that is used to denote the set of optional VUnit packages, i.e. the packages that helps you write testbenches but aren't essential for running VUnit.

eine: @LarsAsplund what about `vunit/vunit/vhdl/helpers/src/helpers_pkg.vhd`?

slaweksiluk: How to tell vunit to compile this new package?  https://github.com/VUnit/vunit/pull/666/commits/215c64d6e43de6915522a61276c50e658b8be0f6

eine: @slaweksiluk, add 'helpers' to https://github.com/VUnit/vunit/blob/master/vunit/builtins.py#L232-L240.

slaweksiluk:
please review. Mind the commit https://github.com/VUnit/vunit/pull/666/commits/f14fcae64199cada679f3c71b5c0061f199c25de where I modified ready signal stimulation - to be truly random and not dependent on calling VC API functions in user main tb.

Issue #573, opened 2019-10-23
Regression Bug in AXI-Stream slave when upgrading from v4.0.8 


FranzForstmayr:
Hi,
I triggered a regression bug when upgrading VUnit to v4.2.0 from v4.0.8.
With git bisect I found the offinding commit 3e098f567af0a442ae9f3a41b0fcb90f2092daed, merged in #429  .

When using the axi-stream slave, the ready signal starts toggling. 
![image](https://user-images.githubusercontent.com/9936788/67392591-1c02dc00-f5a1-11e9-97a2-e3dd542be956.png)

Same issue is in the axi_stream_slave, although the tests are passing, just need approximately twice as long. The offending line is https://github.com/VUnit/vunit/blob/master/vunit/vhdl/verification_components/src/axi_stream_slave.vhd#L95, however I haven't digged deep enough to find out why this change was made.

umarcor:
FTR, other changes to AXI-Stream a few days before #429, introduced another partial regression as explained in #494. Does GitHub allow to easily compare v4.0.8 and v4.2.0 of subdir https://github.com/VUnit/vunit/tree/master/vunit/vhdl/verification_components/src only?

eschmidscs:
We've hit something similar some time ago.
You probably use the pop function with direct response variables or the check function with blocking=>True.
In this case, the slave takes one message, handles it and goes back to idle. Going back to idle causes ready to go low for one cycle.
You either have to use pop with references or check with blocking=>False.

That was different, probably before the merge you mentioned. But the process structure in the slave was changed.

eine:
I created a synthetic example in https://github.com/eine/vunit/tree/axis/synthetic/examples/vhdl/array_axis_vcs/src/test, in an attempt to better understand this issue. According to previous comments here and in #673, the blocking behaviour is produced by `pop_axi_stream` in https://github.com/eine/vunit/blob/axis/synthetic/examples/vhdl/array_axis_vcs/src/test/tb_axis_loop.vhd#L113 (that is, by the API of the AXI Stream Slave verification component).

[...]

PR #673, opened 2020-08-15
Examples/axis tb

FranzForstmayr:
This pull request changes the tb_axis_loop example to show a non-blocking axi-stream pop without any stall cycle, becauseI think the example should work without any stalls. (#573)
A second example shows the random stall feature.

### DIFF
--- a/examples/vhdl/array_axis_vcs/src/test/tb_axis_loop.vhd
+++ b/examples/vhdl/array_axis_vcs/src/test/tb_axis_loop.vhd
@@ -44,7 +44,7 @@ architecture tb of tb_axis_loop is
   signal clk, rst, rstn : std_logic := '0';
   constant m_I : integer_array_t := load_csv(tb_path & csv_i);
   constant m_O : integer_array_t := new_2d(width(m_I), height(m_I), data_width, true);
-  signal start, done, saved : boolean := false;
+  signal start, start_stall, done, saved : boolean := false;
 
 begin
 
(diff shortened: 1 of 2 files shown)


eine:
Overall, I'm good with the proposal, but I'm not sure to understand the motivation for some changes:

On the one hand, the only difference between both testbench files seems to be that the new one defines `master_stall_config` and `slave_stall_config`. Instead of duplicating the files, I think that it should be handled through generics, or functions.

On the other hand, I don't understand the motivation to add an intermediate queue. Is it because you want to show both behaviours at the same time? One transference with blocking and another one with non-blocking? If that is the case, I would avoid it. Instead, I would just add a comment explaining that `await_pop_axi_stream_reply` can be used instead of `pop_axi_stream`. Or, if you prefer, just change it directly and show the non-blocking case only.

My motivation is that this example is meant as a very basic introduction to users willing to connect an AXI IP they have with a couple of Verification Components. Those users typically have no previous experience with VUnit's libs, and this is their first contact. The current example is relatively easy to understand, because there is a process that sends data to the core and there is another process that reads it. However, using queues adds a layer of complexity that is out of the scope of this example.

Anyway, I think there is currently no example that showcases `queue`. Hence, it might be interesting to add one, which might be as simple as 2-3 loops passing data to each other. So, you might take your "new" testbench, and remove the AXI stream from there. Then, we could present the "queues" example as the previous step to understanding stream VCs. What do you think?

FranzForstmayr:
The main reason why I made this PR is the following issue, also described in #573 
Without the patch, a VUnit AXI Stream slave deasserts the ready signal, so there's one clock cycle without data transmission.

bad

The patch enables non-blocking communication between the DUT and the AXI Strream slave, no dead-cycles between the transmission.
good

To show the feature with random stalls, I thought it would fit nice into an example. 
I'm with you, that copying the files is not the best solution, however otherwise the testbench would get much more complicated.

Probably just drop the second commit with the stall example.

Issue #672, opened 2020-08-10
Get Simulator Informations in Python API


jonasexpleo:
For testing with netlists I need some vendor libraries. These are shipped with the simulator ActiveHDL and I use this to include them
```
vu.add_external_library('pmi_work', activehdl_lattice_path / "verilog/pmi_work")
````
 In this case I need to specify the absolute path to ActiveHDL. This path is should be known by vunit, but has no stable interface to use it. My workaround is to use a private interface...
```
vu._simulator_class.find_prefix()
```

A second problem is that I need to change the path of the external library depending on the version of ActiveHDL (10.3 or 11.1).
At least [ActiveHDL](https://github.com/VUnit/vunit/blob/c418e8a18ba38313da013091495e841659ab6c05/vunit/sim_if/activehdl.py#L57) and [GHDL](https://github.com/VUnit/vunit/blob/c418e8a18ba38313da013091495e841659ab6c05/vunit/sim_if/ghdl.py#L121) has a functionality to receive the version of the program.

My idea is to define a function in the `SimulatorInterface` which would return a dict with the path and version of the simulator. Or is there a functionality which expose these information already? Any other suggestions?

eine:
Might the following sequence work for you?

`from vunit.sim_if.factory import SIMULATOR_FACTORY` (https://github.com/VHDL/Compliance-Tests/blob/b5a8580af6a9cd3b0ee40bc74a7fa02b89d0500f/run.py#L3)

https://github.com/VUnit/vunit/blob/f0d861e466eaff965357e9cbb45858e7e9a5c047/vunit/sim_if/factory.py#L106

`select_simulator` should return a class corresponding to the active/selected simulator, and from there you should be able to get the path and the version. If some of them are private, maybe we can just expose them.

@LarsAsplund, what do you think?

jonasexpleo:
Until till now I didn't though about to talk to the interface directly. It seams to be an easy solution.

I did a small PoC in my branch: [jonasexpleo/vunit](https://github.com/jonasexpleo/vunit/tree/fix672) It is only tested for GHDL and ActiveHDL, but Rivera should also work. For ModelSim I didn't find a function to receive the version information.

Probably a better unification of the version number parsing and returning would be better, but for a PoC it should be ok.

eine:
You might want to open a PR with those changes. Not meaning we should merge them (yet). But a (draft) PR will allow us to better discuss it, as we can comment on the code and keep a track of resolved issues.

[...]

PR #674, opened 2020-08-17
Add simulator information interface

jonasexpleo:
A PoC for #672 

- [x] Implemented for GHDL, ActiveHDL and Rivera
- [x] Tested with GHDL and ActiveHDL
- [ ] Implementation for ModelSim
- [ ] Unification of the version number parsing and returning
- [ ] Documentation


eine:
@jonasexpleo, I'm sorry this review was delayed. After discussion with Lars:

- activehdl and rivierapro interfaces have `__str__` and it returns the version. If anything, it should return the name (and be added to ghdl too).
- The versions are already parsed for getting the components. It doesn't seen sensible to undo that work and return the raw string for the user to parse it again. Instead, we should return the components.
- We'd rather have `get_path` and `get_version`, instead of a single `get_info`.

jonasexpleo:
@eine no problem

> activehdl and rivierapro interfaces have __str__ and it returns the version.

1. I only define `__str__` for the version classes of ActiveHDL and RiveraPro. In this class I don't have any information about the name of the simulator.

> It doesn't seen sensible to undo that work and return the raw string for the user to parse it again. Instead, we should return the components.

2. I had the same concerns about to returning the raw string, but it was for me a cleaner way than returning various formats. The cleanest way for me would be to introduce a new version class for all components and returning that (the classes of ActiveHDL and RiveraPro are different implementations). Do I understand your comment correctly, that you would return a dict with the components?

> We'd rather have get_path and get_version, instead of a single get_info.

3. I thought about to be extendable for different information depending on the simulator, but it's OK for me.

PR #675, opened 2020-08-19
fix riviera pro coverage merge error

noasic:
Added braces around the acdb_merge.tcl file path for the case that the path contains spaces (e.g. `C:/Program Files (x86)/Jenkins/workspace/crown-fpga/fpga/projects/crown/sim/vunit_out/rivierapro/acdb_merge.tcl`).


eine:
Thanks!

PR #681, opened 2020-09-22
Workaround for empty macro issue in Riviera-PRO

pib88:
When you add sources and create empty macro like following:
`VU.add_library("tb_uart_lib").add_source_files(SRC_PATH / "test" / "*.sv", defines={"EMPTY_MACRO" : ""})`

It adds to vlog command a following switch:
`vlog +define+EMPTY_MACRO=`

However in Riviera-PRO it ends with following error:
`ERROR VCP0501 "Missing option argument in command line. Option -dv requires an argument."`

To workaround this you need to remove '=' operator when macro has no value.

### DIFF
--- a/vunit/sim_if/rivierapro.py
+++ b/vunit/sim_if/rivierapro.py
@@ -210,7 +210,10 @@ class RivieraProInterface(VsimSimulatorMixin, SimulatorInterface):
         for include_dir in source_file.include_dirs:
             args += ["+incdir+%s" % include_dir]
         for key, value in source_file.defines.items():
-            args += ["+define+%s=%s" % (key, value)]
+            if value != "":
+               args += ["+define+%s=%s" % (key, value)]
+            else:
+               args += ["+define+%s" % (key)]
         return args
 
     def create_library(self, library_name, path, mapped_libraries=None):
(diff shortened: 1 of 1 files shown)


vunit/sim_if/rivierapro.py
```diff
@@ -210,7 +210,10 @@ def compile_verilog_file_command(self, source_file):
         for include_dir in source_file.include_dirs:
             args += ["+incdir+%s" % include_dir]
         for key, value in source_file.defines.items():
-            args += ["+define+%s=%s" % (key, value)]
+            if value != "":
```
eine, line 213: What do you think about the following?

```py
args += ["+define+%s" % key]
if value:
    args[-1] += "=%s" % value
```

Also, please, run `tox -e py38-fmt` for getting rid of CI errors.

pib88: I've updated code with your solution. Seems like it passes in tox tests now

eine: Yes. Errors were produced because of an indentation issue. Your code was otherwise correct too. The change was just a style preference, since it had to be edited.

Thanks!

PR #683, opened 2020-09-29
Special verilog parameters tb_path and output_path don't work in Riviera-PRO

pib88:
When you use tb_path or output_path in verilog they are specified in command line like this:
`-gtb_path=<TB_PATH>`

That does not work in Riviera because you need to put string in double quotes to have it interpreted correctly:
`-gtb_path="<TB_PATH>"`

Since it's impossible to determine parameter type by its value, solution should be limited to parameters that we know are strings: tb_path and output_path.


dalex78:
It look good to me, however, I do not have Riviera and cannot confirm the issue or the proposed fix. But I do not see any issue with the code.
@LarsAsplund, I suggest to merge.

PR #684, opened 2020-09-29
Xcelium support

powARman:
Added support for Cadence Xcelium simulator which is the successor of incisive. 

- Extended incisive.py to support xcelium simulator and maintain backwards compatibility with incisive
- Support vunit options disable_ieee_warnings and vhdl_assert_stop_level
- Fix compile issues in log_deprecated_pkg.vhd
- Fix problems running the test suite (spaces in path)


LarsAsplund:
@powARman Question about Incisive and Xcelium. Is it a use case where you have both an old Incisive license and a new Xcelium licence. In that case you might want to run parallel simulations using both licenses. You could do that by setting `VUNIT_SIMULATOR` to two different simulators for the two test (suites). Now they are handled as one and the same

rodrigomelo9:
Hi @powARman news about this PR? I am using Xcelium at work and I want to introduce Vunit to perform TDD. I am new at Vunit, and I will try with specifying incisive as the simulator, but may be useful to try this PR? Please, let me know

PR #690, opened 2020-10-31
Issue 455

GlenNicholls:
Identical to https://github.com/VUnit/vunit/pull/665 as I needed to fork for other development.

added add_waves_by_depth method and TCL scripts for modelsim, aldec, and ghdl

This is in primarily in relation to #455, but aligns with my comments in #615 . I avoided the #622 PR because it is pretty far behind master and @nfrancque appeared to mostly have stuff for the ModelSim save/load of wave files which is different than what I'm looking to add. I have some outstanding questions:

* Does it make sense to add waves as I have done with `add_waves_by_depth()` which calls `set_sim_option()` under the covers?
* do I need a unit test for `add_waves_by_depth()`? If so, how would I do this since the entire interface is creating a file and adding waves to the simulator?

This PR would close out #455, but the below feature requests are my vision for #615:

* JSON interface such that user can add hierarchy path to signals/variables/constants and some options and VUnit would take care of simulator specific TCL commands
* Add bools to `add_waves_by_depth()` so user can include constants and/or variables as well.
* Add command-line option to add by depth in cases where adding `add_waves_by_depth` to `run.py` doesn't always make sense
* Add simulator GUI option for adding waves by depth (gated by #622 being accepted and merged with master)

### DIFF
--- a/vunit/ui/__init__.py
+++ b/vunit/ui/__init__.py
@@ -735,6 +735,7 @@ avoid location preprocessing of other functions sharing name with a VUnit log or
         .. note::
            Only affects test benches added *before* the option is set.
         """
+        parent_path = Path(__file__).parent.parent.resolve()
         simulator = self._simulator_class.name
         output_path = str(Path(self._output_path) / simulator / "waves.tcl")
         tcl_sources = {
@@ -746,10 +747,7 @@ avoid location preprocessing of other functions sharing name with a VUnit log or
         # error check selected simulator and create path for TCL files
         # and simulator option
         if simulator in tcl_sources:
-            tcl_src_path = str(
-                Path(__file__).parent.parent.resolve() / "sim_if" / "tcl" / \
-                tcl_sources[simulator]
-            )
+            tcl_src_path = str(parent_path / "sim_if" / "tcl" / tcl_sources[simulator])
             sim_opt = f"{simulator}" + \
                 (".gtkwave_script.gui" if simulator == "ghdl" else ".init_file.gui")
         else:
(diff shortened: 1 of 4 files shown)


vunit/ui/__init__.py
```diff
@@ -708,6 +708,61 @@ def enable_location_preprocessing(
...
+
+        Which would, for example, add waveforms to the waveform viewer
+        like so:
+        tb/*
+        tb/uut/*
+        tb/uut/ent1/*
+        tb/uut/ent2/*
+
+        .. note::
+           Only affects test benches added *before* the option is set.
```
eine, line 736: I think it would be desirable to let the option affect different scopes: project, library, specific test... Having it affect "all previously defined tests" feels slightly unidiomatic in the context of VUnit.

eine:
> do I need a unit test for `add_waves_by_depth()`? If so, how would I do this since the entire interface is creating a file and adding waves to the simulator?

I'm not sure about how to test this in CI, but it would be helpful to add some example, or to add this feature to an existing example. That would allow to easily test it locally.

> JSON interface such that user can add hierarchy path to signals/variables/constants and some options and VUnit would take care of simulator specific TCL commands

Are you aware of https://ghdl.github.io/ghdl/using/Simulation.html#cmdoption-ghdl-read-wave-opt? It's not JSON and it's not exactly the same feature. But I guess it might some optional feature for GHDL. Furthermore, it might be feasible to enhance GHDL so that the JSON format is accepted too.

GlenNicholls:
>Are you aware of https://ghdl.github.io/ghdl/using/Simulation.html#cmdoption-ghdl-read-wave-opt? It's not JSON and it's not exactly the same feature. But I guess it might some optional feature for GHDL. Furthermore, it might be feasible to enhance GHDL so that the JSON format is accepted too.

I was not aware of that, but it seems useful to simplify the TCL script provided for GHDL. I'll look into it to see if it would be helpful!

a-panella:
One thing I love about VUnit + Gtkwave is that I can simulate, open the waveform, layout my signals how I like, and then make changes to the hdl as required, resimulate, then all I have to do is hit CTRL+SHIFT+R in gtkwave and the waveform updates. It is so much faster than the way I used to do things. The current sim time cursor doesn't even move either, so you can change logic, rerun vunit test, hit CTRL+SHIFT+R, and see the signals adjust in real time.
**Somewhat related feature request**: provide the feature mentioned above with the Vunit -g switch

PR #693, opened 2020-11-12
fix broken msg passing in wishbone master for (fix #692)

### LINKED ISSUE #692: Wishbone master gives `pop from empty queue error` in Modelsim
I have a problem driving simple Wishbone slave from VUnit WB master. It fails already on the first transaction with message
`Error: Pop from empty queue`. Waveform of faulty behaviour looks like this:
image

After a bit of investigating I've found out that error comes this process:
https://github.com/VUnit/vunit/blob/master/vunit/vhdl/verification_components/src/wishbone_master.vhd#L146
which is receiving message sent here:
https://github.com/VUnit/vunit/blob/master/vunit/vhdl/verification_components/src/wishbone_master.vhd#L97

I think that the problem is connected with delta cycles. My WB slave logic looks like this:
```vhdl
ack_int <= rd_ack_int or wr_ack_int;
wb_o.ack <= ack_int;
wb_o.stall <= not ack_int and wb_en;
```
The `wb_o.ack` and `wb_o.stall` signals are set in the same delta cycle. But it looks like WB master expects that `ack` should be set at least one delta cycle later (assuming that I correctly understand how VUnit message passing works).

@slaweksiluk is my reasoning correct?

VUnit 4.4.0
Modelsim SE (Linux) 10.7a

slaweksiluk:


vunit/vhdl/verification_components/run.py
```diff
@@ -22,7 +22,7 @@ def encode(tb_cfg):
 
 
 def gen_wb_tests(obj, *args):
-    for dat_width, num_cycles, strobe_prob, ack_prob, stall_prob in product(*args):
+    for dat_width, num_cycles, strobe_prob, ack_prob, stall_prob, slave_inst in product(*args):
```
eine, line 25: I would use `slave_comb` instead of `slave_inst`, since my understanding is that the difference between the two cases is whether the slave is combinatorial or registered.

vunit/vhdl/verification_components/run.py
```diff
@@ -86,13 +87,17 @@ def gen_avalon_master_tests(obj, *args):
...
     #  TODO strobe_prob not implemented in slave tb
-    gen_wb_tests(test, [8, 32], [1, 64], [1.0], [0.3, 1.0], [0.4, 0.0])
+    gen_wb_tests(test, [8, 32], [1, 64], [1.0], [0.3, 1.0], [0.4, 0.0], [True, ])
 
 
 TB_WISHBONE_MASTER = LIB.test_bench("tb_wishbone_master")
 
 for test in TB_WISHBONE_MASTER.get_tests():
-    gen_wb_tests(test, [8, 32], [1, 64], [0.3, 1.0], [0.3, 1.0], [0.4, 0.0])
+    if test.name == "slave comb ack":
```
eine, line 96: I'm not sure to understand this. Why is the combinatorial version tested with one set of params but the other one is tested with two?

vunit/vhdl/verification_components/test/tb_wishbone_master.vhd
```diff
@@ -182,22 +189,34 @@ begin
...
-      dat_i => dat_o,
-      dat_o => dat_i,
-      sel   => sel,
-      cyc   => cyc,
-      stb   => stb,
-      we    => we,
-      stall => stall,
-      ack   => ack
-    );
+  slave_gen : if tb_cfg.slave_inst generate
```
eine, line 192: Would it make sense for `comb` to be an option (generic) of the wishbone slave, instead of using a generate in the testbench?

vunit/vhdl/verification_components/run.py
slaweksiluk: The idea is to have possibility to skip wb slave VC instantiation and drive wishbone slave signals directly. It makes possible to test various corner cases - like here when specific set of acknowledge logic lead to crash on modelsim, but not ghdl. I thinks it's not feasible to add this corner cases reproducers to wishbone slave as it will obfuscate code. This is all about wb master VC bug.

vunit/vhdl/verification_components/run.py
slaweksiluk: slave comb ack test case doesn't need/cannot have random stimulus

vunit/vhdl/verification_components/test/tb_wishbone_master.vhd
slaweksiluk: as described in first comment IMHO: no, it would not

PR #703, opened 2021-01-06
Simple signal_driver

eschmidscs:
This is a simple counter-part of the signal_checker / std_logic_checker.

### DIFF
--- a/vunit/vhdl/verification_components/src/signal_driver_pkg.vhd
+++ b/vunit/vhdl/verification_components/src/signal_driver_pkg.vhd
@@ -1,0 +1,69 @@
+
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this file,
+-- You can obtain one at http://mozilla.org/MPL/2.0/.
+--
+-- Copyright (c) 2014-2020, Lars Asplund lars.anders.asplund@gmail.com
+
+library ieee;
+use ieee.std_logic_1164.all;
+
+context work.vunit_context;
+context work.com_context;
+
+use work.sync_pkg.all;
+
+package signal_driver_pkg is
+  type signal_driver_t is record
+    p_actor  : actor_t;
+    p_logger : logger_t;
+    initial  : std_logic_vector;
+  end record;
+
+  impure function new_signal_driver(
+    initial : std_logic_vector;
+    logger  : logger_t := null_logger)
+    return signal_driver_t;
+
+  procedure drive(signal net    : inout network_t;
+                  signal_driver :       signal_driver_t;
+                  value         :       std_logic_vector
+                  );
+
+  procedure wait_until_idle(signal net    : inout network_t;
+                            signal_driver :       signal_driver_t);
+
+  constant drive_msg : msg_type_t := new_msg_type("drive");
+
+end package signal_driver_pkg;
+
+package body signal_driver_pkg is
+  impure function new_signal_driver(initial : std_logic_vector; logger : logger_t := null_logger) return signal_driver_t is
+    variable result : signal_driver_t(initial (initial'range));
+  begin
+    result := (p_actor  => new_actor,
+               p_logger => logger,
+               initial  => initial);
+    if logger = null_logger then
+      result.p_logger := default_logger;
+    end if;
+    return result;
+  end;
+
+  procedure drive(signal net    : inout network_t;
+                  signal_driver :       signal_driver_t;
+                  value         :       std_logic_vector
+                  ) is
+    variable msg : msg_t := new_msg(drive_msg);
+  begin
+    push_std_ulogic_vector(msg, value);
+    send(net, signal_driver.p_actor, msg);
+  end;
+
+  procedure wait_until_idle(signal net    : inout network_t;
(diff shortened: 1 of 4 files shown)


eine:
Ref #704.

smenzel:
I like this concept, using the message queue as a way to encode a sequence of value changes to a signal without the need to wait for clock edges inside the calling process. To make it more useful with sequences that don't change on every clock, is it possible to extend the drive() function with a parameter 'delay_clocks' (or a better name) with default of 0, that waits for the specified amount of clock cycles?

eschmidscs:
@smenzel That is certainly possible, probably not even difficult - but I'm very time-limited right now.
The PR won't be merged until #520 is done, and then I have to modify this PR anyway. So for the "final" version, I'll include this.

eschmidscs:
@smenzel The waiting for some time is now included as by VC requirements. Instead of clock cycles, you have to specify a time. I've added two short testcases to check if it does not fall off by one cycle this way. It seems to work.
However, the tests rarely cover all applications, so please expect some off-by-one-cycle effects...

vunit/vhdl/verification_components/src/signal_driver_pkg.vhd
```diff
@@ -10,60 +9,76 @@ use ieee.std_logic_1164.all;
...
+    logger                     : logger_t                     := null_logger;
+    actor                      : actor_t                      := null_actor;
+    checker                    : checker_t                    := null_checker;
+    unexpected_msg_type_policy : unexpected_msg_type_policy_t := fail)
     return signal_driver_t;
 
+  function get_std_cfg(signal_driver : signal_driver_t) return std_cfg_t;
+
+  impure function as_sync(signal_driver : signal_driver_t) return sync_handle_t;
+
```
LarsAsplund, line 37: Consider supporting the stream master interface. Have a look at
https://github.com/VUnit/vunit/blob/14b5ad16dc80a5e5b6c165f09f010c60ade94808/vunit/vhdl/verification_components/src/avalon_stream_pkg.vhd#L44 and https://github.com/VUnit/vunit/blob/14b5ad16dc80a5e5b6c165f09f010c60ade94808/vunit/vhdl/verification_components/src/avalon_source.vhd#L47

vunit/vhdl/verification_components/src/std_logic_driver.vhd
```diff
@@ -32,14 +33,17 @@ begin
...
+    receive(net, get_actor(signal_driver.p_std_cfg), request_msg);
     msg_type := message_type(request_msg);
 
     if msg_type = drive_msg then
       internal_msg := new_msg;
       push_std_ulogic_vector(internal_msg, pop_std_ulogic_vector(request_msg));
       send(net, internal_actor, internal_msg);
 
+    elsif msg_type = wait_for_time_msg then
+      handle_wait_for_time(net, msg_type, request_msg);
```
LarsAsplund, line 45: You can actually remove this elsif branch and move the `handle_wait_for_time` before the if statement

vunit/vhdl/verification_components/test/tb_std_logic_driver.vhd
```diff
@@ -0,0 +1,92 @@
...
+  signal value : std_logic_vector(3 downto 0);
+begin
+  main : process is
+    variable t : time;
+  begin
+    test_runner_setup(runner, runner_cfg);
+    show(logger, display_handler, pass);
+
+    if run("Test default") then
+      check_equal(checker, value, std_logic_vector'("0010"), "Incorrect initial value");
```
LarsAsplund, line 36: You should probably make sure that the initial value also survives the first clock edges.

eschmidscs:
@LarsAsplund Added your inputs with a separate commit. Do you prefer that squeezed?

eschmidscs:
@LarsAsplund Actually, your comment about the stream master interface made me think about the use of this block. I guess one point where it differs from a stream master is that there is no valid signal and the output is always driven.
But maybe I should add a generic option to not have the output clocked? That would mirror the signal checker behaviour directly and could have some use.

LarsAsplund:
@eschmidscs I rather have that squeezed. Makes the history cleaner. 

The stream master interface doesn't dictate that you need a valid signal. The push procedure has an optional `last` parameter and it defaults to false. As long as we're pushing data it's a valid interface to reuse. The details about how things are pushed is for the VC to decide so that may differ between different components.

eschmidscs:
@LarsAsplund Ok, I just squeezed all three commits.

Sorry, I was somewhat unclear: It is not the stream master interface that requires the valid, but the stream master VC (at least the axis one). I was trying to find a reason to have the signal_driver at all. If the axi_stream_master would keep the output stable after valid, it would be some kind of signal_driver, right?
Thus the signal_driver could distinguish a bit more by allowing signals to be driven unclocked (i.e. in the moment the push is issued).

LarsAsplund:
@eschmidscs Both `axi_stream_master` and `avalon_source` could be reduced to just outputting data if you do not connect the other signals and keep some inputs static. I think the signal driver has a place though. It is more intuitive to use that rather than a bus standard which many would feel is unrelated to what they are trying to achieve.

I think you're right about the clock. Default should be as simple as possible and that means no wait for the clock. The port we're driving might be asynchronous.

PR #704, opened 2021-01-06
signal checker extensions

eschmidscs:
This PR contains a bunch of changes that I missed.
- Have it in the context, just for usability
- Allow to disable the monitoring. We have lots of error signals monitored by this block. During reset, these signals do about anything. It becomes quite tedious to add all steps to simulation start...
- Function to actually read back the monitored value. A TB might just read the value for further actions.
- Add a checker_t as well. The logger_t does not add to the checker statistics.


eschmidscs:
Sorry, too fast here: I have to update the TB.

vunit/vhdl/verification_components/src/signal_checker_pkg.vhd
```diff
@@ -17,11 +17,13 @@ package signal_checker_pkg is
     -- Private
     p_actor : actor_t;
     p_logger : logger_t;
+    p_checker : checker_t;
     initial_monitor_enable : boolean;
   end record;
 
   impure function new_signal_checker(
     logger : logger_t := null_logger;
+    checker: checker_t := null_checker;
```
LarsAsplund, line 26: I think this is a good time to start having a look at the new VC standard we will implement going forward. Not all of it but have a look at how logger and checkers are related in this commit https://github.com/VUnit/vunit/compare/vc-template#diff-722cac3044b13447113d7459cf4a5c0b8ec5e27c212c1aaa092af1152950d6ee to avoid unnecessary changes when that is merged

eschmidscs: I think I don't completely understand yet.
Should I implement the part with the signal_checker_logger / signal_checker_checker? That would be feasible.
But the std_cfg_t does not seem merged, so the functionality based on this has to wait.

Backward-compatibility should probably be kept, right?

#703 is more flexible, as it is not modifying an existing VC. Should I try to get that one closer to the template?

eine: @eschmidscs, the point is that #520 will likely be merged before accepting further modifications to verification components. #520 is ready to be merged already; we are just lacking some time for final testing/reviewing. Therefore, it would be desirable for PRs such as this one to be based on #520.

> Backward-compatibility should probably be kept, right?

VCs are still *beta*: http://vunit.github.io/verification_components/user_guide.html. The purpose of #520 is to go for the first release including non-beta VCs. Hence, some backwards compatibility breaks might happen.

LarsAsplund (commented):
We've have always been restrictive when it comes to disable errors but we do have some generally available functions. Have you seen http://vunit.github.io/logging/user_guide.html#disabled-logs? When using that you must also do an extra step at `test_runner_cleanup` (http://vunit.github.io/run/user_guide.html#counting-errors-with-vunit-logging-check-libraries). Interestingly we do not have an enable procedure... Would your use case be solved if such a procedure was added?

eschmidscs:
At least I did not have the disabled logs in active memory...
Having an enable_logs might work, yes. If the test_runner would not fail on the disabled logs, but still fail on the later, enabled logs.

LarsAsplund (commented):
Looks good!

vunit/vhdl/verification_components/src/signal_checker_pkg.vhd
LarsAsplund: I was mostly thinking about the behavior of logger and checker depending  on what's being provided when calling the new function. The core of the behavior is

``` vhdl
    result.p_logger                     := logger when logger /= null_logger else default_logger;
   
    if checker = null_checker then
      if logger = default_logger then
        result.p_checker := default_checker;
      else
        result.p_checker := new_checker(logger);
      end if;
    else
      result.p_checker := checker;
    end if;
```

where

```
  constant signal_checker_logger : logger_t := get_logger("vunit_lib:signal_checker_pkg");
  constant signal_checker_checker : checker_t := new_checker(signal_checker_logger);
```

eschmidscs: @eine Thanks for the background! It seems wiser to wait for #520 then, but this one has been around for quite a while...
I can definitely rebase this one and #703 on #520 and try to get them compatible with the rules there. Is it stable enough for this?
I guess I cannot help with reviews, but what is left for testing in #520?

hughesjs (approved):
Looks good to me

PR #709, opened 2021-02-28
Adding ability to define the actor on new_axi_slave function

rftafas:
This change should not affect any current usage, but adds possibility to define the actor for the axi_slave being created. For some reason on a test needing several AXI Slave VCIs, modelsim put all of them under the same "id" and things got awkward.

This has solved it.

### DIFF
--- a/vunit/vhdl/verification_components/src/axi_slave_pkg.vhd
+++ b/vunit/vhdl/verification_components/src/axi_slave_pkg.vhd
@@ -33,7 +33,7 @@ package axi_slave_pkg is
 
   constant axi_slave_logger : logger_t := get_logger("vunit_lib:axi_slave_pkg");
   impure function new_axi_slave(memory : memory_t;
-                                actor : actor_t := new_actor;
+                                actor : actor_t := null_actor;
                                 address_fifo_depth : positive := 1;
                                 write_response_fifo_depth : positive := 1;
                                 check_4kbyte_boundary : boolean := true;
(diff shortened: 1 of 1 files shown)


LarsAsplund:
There is work being done to standardize some parts of VCs and VCIs to improve interoperability. One of the rules of that standard is to have the ability you're looking for. All present VCs have been modified to comply with this standard but that work hasn't been merged yet. It's been on hold for a while but is number two on my prio list and I will get to it shortly.

We can do this before that as it doesn't break any backward compatibility but please have a look at https://github.com/VUnit/vunit/blob/ff6737ade82856575fb0b64052fa8ff7caf92467/vunit/vhdl/verification_components/src/axi_slave_pkg.vhd#L168 and modify such that your default value is `null_actor` to better match that standard.

rftafas:
I made the changes, but It is still having trouble on some seemingly unrelated checks.

rftafas:
It seems that the change from 'new_actor' to 'null_actor' caused this. Most probably, the examples will have to be adapted to include an actor now, as it does not do that anymore.

vunit/vhdl/verification_components/src/axi_slave_pkg.vhd
```diff
@@ -147,7 +149,7 @@ package body axi_slave_pkg is
                                 max_response_latency : delay_length := 0 ns;
                                 logger : logger_t := axi_slave_logger) return axi_slave_t is
   begin
-    return (p_actor => new_actor,
+    return (p_actor => actor,
```
LarsAsplund, line 152: Sorry, I was a bit unclear here. You still need to call `new_actor`. `p_actor` has to be set to

`actor when actor /= null_actor else new_actor`

LarsAsplund: Once you've fixed that you can also rebase on master and then, hopefully, the remaining lint issues will disappear.

rftafas: I thought of that at first, but got worried that it would not be the acceptable way. Will do it.

PR #717, opened 2021-05-03
Added example Gitlab CI integration

hughesjs:
I integrated VUnit with Gitlab CI for one of my projects. Figured it makes sense to have a directory full of these examples for different CI tools such as Jenkins, Github Actions etc.

### DIFF
--- a/examples/ci-integration/.gitlab-ci.yml
+++ b/examples/ci-integration/.gitlab-ci.yml
@@ -1,0 +1,19 @@
+image: ghdl/vunit:llvm
+stages:
+  - test
+
+before_script:
+  - cd $CI_PROJECT_DIR/src/tests
+
+test:
+  tags:
+    - linux-docker
+  stage: test
+  script:
+    - python3 ./run.py --xunit-xml ./test_output.xml
+  artifacts:
+    when: always
+    paths:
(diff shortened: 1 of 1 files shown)


umarcor:
Hi @hughesjs! This is an interesting contribution! However, I think that it belongs in https://github.com/VUnit/tdd-intro (or maybe https://github.com/VUnit/vunit_action renamed to vunit-ci), rather than here.

This PR could be repurposed for actually explaining/documenting the example. See http://vunit.github.io/ci/script.html. So, a "Gitlab" subsection might be created. Initially, it'd be enough with a few refs to the GitLab Docs.

BTW, VUnit is used by the VASG for testing standard and IEEE libs. See https://gitlab.com/IEEE-P1076/packages/-/pipelines. On the one hand, that might be a good additional reference. On the other hand, you might want to suggest some enhancements to https://gitlab.com/IEEE-P1076/packages/-/blob/develop/.gitlab-ci.yml.

hughesjs:
I didn't realise those repos existed. I think tdd-intro might be the best place. I'll get this moved over when I'm back at my desk

hughesjs:
Sorry, I totally forgot to do this... Just had a look at the `tdd-intro` repo, looks like you've got this covered since

PR #718, opened 2021-05-12
initial values on outgoing ram_master ports

embed-me:
Avoid undefined signals on outgoing ports of the ram interface.

### DIFF
--- a/vunit/vhdl/verification_components/src/ram_master.vhd
+++ b/vunit/vhdl/verification_components/src/ram_master.vhd
@@ -22,9 +22,9 @@ entity ram_master is
   port (
     clk : in std_logic;
     en : out std_logic := '0';
-    we : out std_logic_vector(byte_enable_length(bus_handle)-1 downto 0);
-    addr : out std_logic_vector(address_length(bus_handle)-1 downto 0);
-    wdata : out std_logic_vector(data_length(bus_handle)-1 downto 0);
+    we : out std_logic_vector(byte_enable_length(bus_handle)-1 downto 0) := (others => '0');
+    addr : out std_logic_vector(address_length(bus_handle)-1 downto 0) := (others => '0');
+    wdata : out std_logic_vector(data_length(bus_handle)-1 downto 0) := (others => '0');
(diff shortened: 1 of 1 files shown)


umarcor:
I'm not sure it is correct to avoid undefined signals. The undefined value exist for it to be used... See https://github.com/stnolting/neorv32/issues/128#issuecomment-880546512.

embed-me:
Undefined initial signals propagate to the UUT. Simple checks like "check_not_unknown" will fail. Other master modules (eg. [ram_master](https://github.com/VUnit/vunit/blob/master/vunit/vhdl/verification_components/src/ram_master.vhd), [axi_stream_master](https://github.com/VUnit/vunit/blob/master/vunit/vhdl/verification_components/src/axi_stream_master.vhd),...) all have their output values defined, so I still think it makes sense to add them.

eschmidscs:
As you mention axi_stream_master: You have the option to have the outputs undefined if invalid.
The ram_master already has the en output initialized to zero. If the UUT fails due to undefined values while en is zero, the UUT should be fixed to not consider any other signals while en is zero.

embed-me:
In general, this makes sense of course, however, the enable port for blk-mem is optional (at least for Xilinx, but I think also for Altera/Intel). If you do not use this signal, for example, because the Memory is always enabled (like it is the case for me), default values would help.
Also, if you follow your logic, why does the "axis_stream_master" define default values for "tdata" (and other ports) if "tvalid" is already set to a default value of '0'?

eschmidscs:
Well, the possibility to drive invalids was added on our request. So the default is to have it drive zeros (for backward compatibility), but you can make it drive X. Which would solve your issue, if ram_master would behave the same way.
I personally would prefer to have the default drive X or U and add an option to drive 0. But we came late and are not the initiators, so it is the other way round ;)

I have to admit that I never used ram_master, so I'm not really competent when it comes to its application. I was triggered by axi_stream_master in your post, and wanted to point out that this block can drive U as default as well.

The enable port becomes a philosophical topic indeed, because the default assignment of the Xilinx memory is '1', but the ram_masters default assignment is '0'. Maybe the default of '0' is not good either...

Issue #719, opened 2021-05-18
non-blocking check_axi_stream does not respect stall_config


eschmidscs:
**Symptom**
Using check_axi_stream with blocking=>false and a non-zero stall_config keeps TREADY high for all time.

**Cause**
The stall_config works fine for all pop*-functions.
However, the check_axi_stream with blocking => false is not based on pop functions, but issues a specific message to the slave. The implementation of that part does not respect the stall_config. It is simply not implemented.

eschmidscs:
I'm working on a test and fix now.

eschmidscs:
Fixed with #720

PR #720, opened 2021-05-18
Bugfix/axi stream slave nonblocking check

eschmidscs:
Fix for #719.
I also had to fix partial non-blocking checks...


eschmidscs:
The checks all seem to fail with the same python error.
Is that something that I caused or something else?

LarsAsplund:
@eschmidscs I fixed the new problems on master so if you rebase on that it will probably pass.

eine:
Thanks @eschmidscs! 😃

PR #724, opened 2021-07-22
Use MAJOR and MINOR constants to check supported Python version

std-max:
Use MAJOR and MINOR constants to check supported Python version.


eine:
Thanks!

PR #731, opened 2021-08-30
log_deprecated_pkg: fix compilation issues with Cadence tools

umarcor:
This PR picks a commit from #684, which is required according to #728.

/cc @powARman @rodrigomelo9

### DIFF
--- a/vunit/vhdl/logging/src/log_deprecated_pkg.vhd
+++ b/vunit/vhdl/logging/src/log_deprecated_pkg.vhd
@@ -101,7 +101,7 @@ package body log_deprecated_pkg is
         end if;
       end function;
 
-      impure function get_logger_display_handler return log_handler_t is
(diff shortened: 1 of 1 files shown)


LarsAsplund:
I think it's time to remove this deprecated stuff altogether now that we're about to make a new major release

umarcor:
@LarsAsplund, agree. However, we have two major changes coming up:

1. Addition of the "rule based" VC plumbing.
2. Moving some VCs to separated repos and submoduling them.

I think that we might want to release v4.6.0 after 1, and v5.0.0 after 2. We, might deprecate stuff between those. Alternativaly, we could bump to v5.0.0, but then maybe we should go to v6.0.0 to be semver compliant. That is because, strictly, reworking the VCs in a backwards incompatible way is ok, as long as those are in Beta. However, deprecations and submoduling forces a major version bump.

PR #734, opened 2021-08-31
Add draft of ghdl/rivierapro simulation with precompiled artifacts.

solsjo:
This simplifies dependency resolution, caching of artifacts and tests in
buildsystems such as bazel.


examples/vhdl/docker_run_precompiled.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
+
+cd $(dirname "$0")
+
+docker run --rm -t \
+  -v /$(pwd)/../..://work \
+  -w //work/examples/vhdl \
+  -e PYTHONPATH=//work \
+  ghdl/vunit:llvm-master sh -c ' \
+    VUNIT_SIMULATOR=ghdl; \
```
umarcor, line 10: Since the shell commands are half of the file, I'd recommend saving it to a different file. That will make it easier to read.
Note that the container is executed in the same place as the script. Hence, `... ghdl/vunit:llvm-master ./myscript.sh` should work.

vunit/sim_if/ghdl.py
```diff
@@ -273,35 +273,39 @@ def compile_vhdl_file_command(self, source_file):
...
     ):
         """
         Return GHDL simulation command
         """
         cmd = [str(Path(self._prefix) / self.executable)]
-
-        if ghdl_e:
-            cmd += ["-e"]
+        if precompiled:
+            self._vhdl_standard = config._design_unit.source_file._vhdl_standard
```
umarcor, line 283: Why does the standard need to be set here?

vunit/sim_if/ghdl.py
```diff
@@ -317,32 +321,35 @@ def _get_command(  # pylint: disable=too-many-branches
             elif self._gtkwave_fmt == "vcd":
                 sim += ["--vcd=%s" % wave_file]
 
-        if not ghdl_e:
-            cmd += sim
-            if elaborate_only:
-                cmd += ["--no-run"]
+        if not precompiled:
```
umarcor, line 324: Maybe return fast if precompiled. I.e. invert the logic of the condition and deindent the rest of the function.

vunit/sim_if/ghdl.py
```diff
@@ -375,8 +381,14 @@ def simulate(  # pylint: disable=too-many-locals
             gcov_env["GCOV_PREFIX"] = coverage_dir
             self._coverage_test_dirs.add(coverage_dir)
 
+        if precompiled:
+            precompiled_binary_path = Path(precompiled)
```
umarcor, line 385: Maybe let precompiled always be a Path (add a type hint), and handle it wherever this function is called.

vunit/sim_if/ghdl.py
```diff
@@ -375,8 +381,14 @@ def simulate(  # pylint: disable=too-many-locals
...
+            cwd = precompiled_binary_path.parent.absolute()
```
umarcor, line 386: Since precompiled_binary_path is used once only, I think it's reasonable to do:

`cwd = precompiled.parent.absolute() if precompiled else None`

vunit/sim_if/vsim_simulator_mixin.py
```diff
@@ -312,10 +312,14 @@ def _run_persistent(self, common_file_name, load_only=False):
         except Process.NonZeroExitCode:
             return False
 
-    def simulate(self, output_path, test_suite_name, config, elaborate_only):
+    def simulate(self, output_path, test_suite_name, config, elaborate_only, precompiled=None):
         """
         Run a test bench
         """
+
+        with open(self._sim_cfg_file_name, "w") as ofile:
```
umarcor, line 320: What's the motivation for this change?

vunit/test/bench.py
```diff
@@ -106,19 +106,18 @@ def _check_architectures(design_unit):
                     )
                 )
 
-    def create_tests(self, simulator_if, elaborate_only, test_list=None):
+    def create_tests(self, simulator_if, elaborate_only, test_list=None, precompiled=None):
```
umarcor, line 109: Is `precompiled` something to be set per test? I'd say it's something related to the tool/backend, which should be unrelated to the tests.

vunit/sim_if/ghdl.py
solsjo: The has_valid_exit_code() function accesses the self._vhdl_standard, where it crashes because the 'other' parameter in the VHDLStandard object for __lt__(self, other) is None.

I guess it could be set elsewhere? But since I'm not well-versed with this code base other than the user API, do you have a suggestion as to where?

vunit/sim_if/ghdl.py
solsjo: Agree!

vunit/sim_if/ghdl.py
solsjo: Good point

vunit/sim_if/vsim_simulator_mixin.py
solsjo: This was a half-baked commit, for just presenting the general idea, in the end I want to support all the simulators.
Though ghdl is nice to start with. So as to delimit the commit.
That being said, by lint also complains about differing function signature and between the overridden base function and the subclass function, so all 'simulate' functions will need to be updated

vunit/test/bench.py
solsjo: This was sort of a wack-a-mole implmentation, I just wanted to make the precompiled param available for the sim_if classes. So I let it trickle down the same path as elaborate_only.

I did that as I'm not that familiar with the inner workings of this code base. Could I do it in another way?
Perhaps I could add a set-function for precompiled on the sim_if function, and perhaps at the same time add one for the elaborate_only param?

vunit/sim_if/ghdl.py
solsjo: I'll look into it

examples/vhdl/docker_run_precompiled.sh
solsjo: Good point

vunit/test/bench.py
solsjo: Or add it to the configuration object? since it seems to handle sim_options

vunit/sim_if/ghdl.py
solsjo: Done

vunit/sim_if/vsim_simulator_mixin.py
solsjo: I updated the VUnit argument to only allow --precompiled for ghdl and riviera pro.
Riviera pro set experimental.

vunit/test/bench.py
solsjo: Updated the constructor for the sim_if classes to accept both elaboration and precompiled.

Since both of these won't change during a VUnit session.

examples/vhdl/docker_run_precompiled.sh
solsjo: @umarcor Anything else you want me to fix for this commit?

solsjo:
@eine, @umarcor,
I understand that this pull-request hasn't been updated in a while but I just wanted to notify you that I've created bazel rules for both ghdl and vunit and setup an example here:

https://github.com/solsjo/rules_ghdl
https://github.com/solsjo/rules_vunit
https://github.com/solsjo/ghdl-example

if you want to try it out :)

PR #737, opened 2021-09-06
examples/vhdl/user_guide: add VHDL 1993 variant

umarcor:
I was chatting with @rodrigomelo9, as it seems that the usage of the user guide with simulators which don't support 2008 completely is unclear.

In this PR, subdir `vhdl1993` is added to `examples/vhdl/user_guide`, and the sources are adapted to use VHDL 1993. Precisely, the optional `vhdl_standard` argument is used when calling `from_argv`, and the `context` in VHDL sources is replaced with use statements.

By the way, an skipIf decorator is added to the 2008 version in the acceptance tests, so that it is skipped with simulators that don't support contexts.

/cc @rodrigomelo9


rodrigomelo9:
Hi @umarcor. Sorry, I was unclear. `context` is supported by Xcelium and the user guide which is written in VHDL 2008 is fine. I think that "my" confusion, was related to the fact that the acceptance test is prepared to check VHDL 2008. Using 1993, more tests fail.

PR #743, opened 2021-09-22
Use f-strings for formatting strings

umarcor:
Currently, we are using a deprecated syntax (`"some text %s" % variable`) for formatting strings. CI linting is complaining about it.

There are three alternatives:

- `"some text %s".format(variable)"`
- `f"some text {variable}"`
- `"some text" + str(variable)`

It seems that the official syntax going forward is the second one only (f-strings). I tried using the first one, but linters would still complain. With regard to the last one, that is not desirable I think, however, in some cases I found it necessary because I couldn't make it work using f-strings.

In this PR, most of the sources in the codebase are modified for using f-strings instead of `"some text %s" % variable` or `"some text %s".format(variable)"`. The remaining sources belong to `vunit/sim_if`. I "fixed" those already, but I cannot test all of them. Hence, I decided to include the `ghdl.py` only, and handle those in an upcoming PR.

I followed the following guidelines:

- Replace `%s` with `{VARIABLE!s}`.
- Replace `%r` with `{VARIABLE!r}`.
- Replace `%i` with `{VARIABLE:d}`.
- Replace `%.1f` with `{VARIABLE:.1f}`.
- Replace `"%s\n" % " ".join(cmd)` with `" ".join(cmd) + "\n"`, and `"%s\n" % ("=" * (max(max_len + 25, 0)))` with `("=" * (max(max_len + 25, 0))) + "\n"`.

### DIFF
--- a/tools/create_release_notes.py
+++ b/tools/create_release_notes.py
@@ -38,7 +38,7 @@ def create_release_notes():
 
     with (source_path / "release_notes.rst").open("w") as fptr:
         fptr.write(
-            """
+            f"""
 .. _release_notes:
 
 Release notes
@@ -46,10 +46,9 @@ Release notes
 
 .. NOTE:: For installation instructions read :ref:`this <installing>`.
 
-`Commits since last release <https://github.com/VUnit/vunit/compare/%s...master>`__
+`Commits since last release <https://github.com/VUnit/vunit/compare/{latest_release.tag!s}...master>`__
 
 """
-            % latest_release.tag
         )
 
         fptr.write("\n\n")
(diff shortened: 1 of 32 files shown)


Paebbels:
What linter are you using?

umarcor:
@Paebbels, see:

- https://github.com/VUnit/vunit/blob/master/pyproject.toml#L45
- https://github.com/VUnit/vunit/tree/master/tests/lint
  - https://github.com/VUnit/vunit/blob/master/tests/lint/test_pycodestyle.py
  - https://github.com/VUnit/vunit/blob/master/tests/lint/test_pylint.py
  
In this case, pycodestyle is passing, and pylint is failing. So:

https://github.com/VUnit/vunit/blob/14b5ad16dc80a5e5b6c165f09f010c60ade94808/tests/lint/test_pylint.py#L26-L33

and https://github.com/VUnit/vunit/blob/master/tests/lint/pylintrc.

PR #745, opened 2021-09-23
parsing/tokenizer: partial revert of 5141f7c6 (#735)

### LINKED ISSUE #735: Massive delay before test actually starts when using Modelsim
Hello,

I am using VUnit with Modelsim Intel free edition. Ever since I started using VUnit, I experience a massive delay (several minutes) whenever I launch a test, before the test is actually run. What is the reason for this massive wait time? When I do not use VUnit, the test starts immediately, so I am sure this has nothing to do with the Modelsim free edition code size limitations.

umarcor:
Close #735.

As discussed in #735, this improves performance significantly.

The previous change was done because of some linter issue. However, this PR seems not to trigger any new warning. Therefore the original issue might have been a temporary rule in the linter.

### DIFF
--- a/vunit/parsing/tokenizer.py
+++ b/vunit/parsing/tokenizer.py
@@ -13,8 +13,11 @@ import re
 from vunit.ostools import read_file, file_exists, simplify_path
 
 
+TokenType = collections.namedtuple("Token", ["kind", "value", "location"])
+
+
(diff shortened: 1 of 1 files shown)


LarsAsplund:
Or maybe I just messed up :-)

PR #747, opened 2021-09-28
Use more f-strings for formatting strings

umarcor:
This PR complements #743. The remaining f-string related pylint issues are fixed. However, I could not test these, because I don't have access to all those simulators. Therefore, I would be grateful if other users could test this branch and report any issue.


umarcor:
FTR, @@NanooooK confirmed that acceptance tests are passing with ModelSim. Two of the examples are failing for legit reasons, unrelated to this PR.

NanooooK:
I apologize but I didn't run the tests correctly (wrong branch). I need to re-run them.

LarsAsplund:
@umarcor Riviera-PRO works with exception of the `array_axis_vcs` example. The reason is the PSL included. This is a known exception but maybe we should exclude this test unless the simulator is GHDL. Alternatively, we can remove the PSL since it's not essential for show casing VUnit.

LarsAsplund:
@umarcor Active-HDL has the same result as Riviera-PRO.

NanooooK:
@umarcor For me with ModelSim using `tox -e py39-acceptance-modelsim
`, most of tests fail because of the following:
```
# ** Warning: (vsim-3351) Invalid value "{s_value!s}" for generic "runner_cfg".  Not using this value.
# ** Fatal: (vsim-3350) Generic "runner_cfg" has not been given a value.
```

or

```
#             0 ps - runner               -    INFO - Test case: Test that pass
#             0 ps - runner               -    INFO - Test case: Test that stops the simulation on first error
# ERROR - Error message
```

vunit/sim_if/modelsim.py
```diff
@@ -410,9 +410,9 @@ def encode_generic_value(value):
     """
     s_value = str(value)
     if " " in s_value:
-        return '{"%s"}' % s_value
+        return f'{"{s_value!s}"}'
```
LarsAsplund, line 413: Should be `return f'{{"{s_value!s}"}}'`

umarcor: Thanks! I applied the fix and force-pushed.

umarcor:
@NanooooK, can you please run the acceptance tests again?

umarcor:
> @umarcor Riviera-PRO works with exception of the `array_axis_vcs` example. The reason is the PSL included. This is a known exception but maybe we should exclude this test unless the simulator is GHDL. Alternatively, we can remove the PSL since it's not essential for show casing VUnit.

On the one hand, this was a conscious decision. The used syntax is valid VHDL 2008, so it should be supported by simulators which are compliant. However, there are some simulators which don't support the syntax, but we don't know certainly which. Also, we don't know if the feature is missing because of the license tier. Therefore, having this test work in CI but fail when running acceptance tests with other simulators is made to raise awareness of the capabilities that VHDL 2008 (PSL) provides and which we are not using in the community.

Nonetheless, I agree that it can be painful, and from VUnit's perspective it might not be desirable that new users find this error and they are forced to understand the motivation. I am OK with disabling it for simulators other than GHDL, but I would like some some accurate handling of the case. For instance, we can mark it as `xfail` for simulators other than GHDL. That will attempt to run it, but will not consider it a failure in the summary (so acceptance-tests will be green). Moreover, we can convert the block to a conditional generate controlled by a generic, which we set from `run.py`. Some simulators might be able to parse the source, even though they cannot handle the block. Others simulators will probably choke on parsing.

Therefore, if you are ok, I will mark it as XFAIL when simulators other than GHDL are used, and I will open an issue for tracking an improvement of the reasons for each sim.

NanooooK:
> @NanooooK, can you please run the acceptance tests again?

Two tests failed but as you said they are out of scope for this PR (verilog_ams, and the assert in the axis_vcs_example)

LarsAsplund:
@NanooooK I think that verilog_ams requires a good enough license. What ModelSim are you using?

vunit/sim_if/modelsim.py
```diff
@@ -410,9 +410,9 @@ def encode_generic_value(value):
     """
     s_value = str(value)
     if " " in s_value:
-        return '{"%s"}' % s_value
+        return f'{{"{s_value!s}"}}'
     if "," in s_value:
-        return '"%s"' % s_value
+        return f'"{value!s}"'
```
LarsAsplund, line 415: I don't think `value` or `s_value` matters here but line 413 and 415 should be consistent.

LarsAsplund (commented):
I reviewed all the files now.

vunit/sim_if/modelsim.py
umarcor: Fixed!

NanooooK:
@LarsAsplund I am using ModelSim DE.

The error for AMS test is the following: `** Error (suppressible): (vlog-1902) Option "-ams" is either unknown, requires an argument, or was given with a bad argument.`

LarsAsplund:
@NanooooK Yeah, it might not be sufficient for AMS. Siemens has Questa-ADMS which might be the only product supporting this.

PR #762, opened 2021-10-23
docs: add subdir 'logo'; add banner

umarcor:
In this PR, the SVG sources of the logo are moved to subdir 'logo', so that `docs/_static` contains the sources to be used in the Sphinx site only. At the same time, since I got to install the correct font, `logo/banners.svg` is added, which includes multiple possible banners. One of them is used in the README. By the way, the icons in the shields/badges are updated. See https://github.com/dbhi/vunit/tree/docs-logo.


umarcor:
I added a thin stroke around the white border, in order to preserve the sign look/feel on white backgrounds:



image

I also updated the banner to a "wide sign": https://github.com/dbhi/vunit/blob/docs-logo/README.md

@LarsAsplund if you are ok with these styles, I think we can merge this even though the generation is not automated yet.

LarsAsplund (approved):
Looks good to me

PR #764, opened 2021-10-23
remove compile_builtins and vunit.verilog

### LINKED ISSUE #559: Do not add builtins by default

umarcor:
Close #559.

This PR is the complement of #757. There, deprecation are announced. Here the features are removed.
This PR will be kept as a draft, because it is expected to be merged in future releases, not in the next one.


umarcor:
@LarsAsplund after the discussion in #757, I updated this PR to include a larger and coloured note about the breaking change. I also updated the documentation to make the content about (V)HDL libraries/utilities easier to find and understand. In order to do so, I propose to create a "Guides" section:

image

Issue #767, opened 2021-10-24
Building VUnit's verification components and OSVVMLibraries at the same time


umarcor:
This is an spin-off coming from #754. That issue is about using OSVVMLibraries in VUnit, after compiling them externally (using OSVVM's `.pro` files, GHDL's vendor scripts or any other solution). It also covers builting OSVVMLibraries in a `run.py` script, while disabling (not building) VUnit's verification components. This one is a complementary issue to discuss how to actually build VUnit's verification components and OSVVMLibraries at the same time.

> https://github.com/VUnit/vunit/issues/754#issuecomment-950287485
> @umarcor What happens if we have the latest OSVVM as a submodule and that is included with `add_osvvm` or `add_verification_components` and then just the rest are compiled manually. Wouldn't that work?

@LarsAsplund, that's an interesting question. Yes, it needs to work, meaning it is technically possible and not complex to achieve for someone experienced with both VUnit and OSVVM. However, it is not easy to understand for an arbitrary newcomer.

- VUnit master is using OSVVM 2021.09, because I bumped it some days ago: https://github.com/VUnit/vunit/commit/d1956b0db3c94e99bb13d005dea8b32197581465.
- In OSVB, I am installing OSVVM using three different approaches:
  - GHDL's vendor scripts (bash).
  - OSVVM's `.pro` files (tclsh).
  - A VUnit `run.py` file (python).

OSVVM's `.pro` files and VUnit's `run.py` file can handle OSVVM 2021.09, but GHDL's vendor scripts were not updated yet, so they require OSVVM 2021.06 (see ghdl/ghdl#1900). Therefore, the default CI in OSVB is using OSVVM 2021.06 at the moment: https://github.com/umarcor/osvb/tree/main/mods. In that context, mixing VUnit's OSVVM core (2021.09) with the submoduled OSVVMLibraries (2021.06) is likely to crash. Hence, I had to do several modifications:

- Ensure that the checked out version of OSVVMLibraries is matching the version of OSVVM submoduled in VUnit. There is no programmatic procedure to achieve it. The user needs to read VUnit's changelog to find it out. It is possible to know the commit sha of the OSVVM core that VUnit is using, but that belongs to a submodule of OSVVMLibraries. I'm not sure you can find the commit in the "parent" that matches a bump of the "child". Therefore, it's mainly a manual step.
- Do not use GHDL's vendor scripts.
- Update the `run.py` script to add `VU.add_verification_components()` and *not* to add lib OSVVM manually.

[...]

PR #771, opened 2021-10-26
builtins: skip addition of builtins if the library is added previously

umarcor:
As discussed in #767, this PR modifies the functions for adding the built OSVVM and JSON-for-VHDL.  Currently, a try...except block is used:

https://github.com/VUnit/vunit/blob/7879504ba6a97be82137199e8819f770e4017681/vunit/builtins.py#L139-L142

That approach has two main caveats. On the one hand, the `try` statement is not case sensitive. Hence, if a library named `OSVVM` exists, it does not get it when asking for `osvvm`. That falls back to the exception, which crashes because adding a library is case insensitive. If no crash is produced, additional sources are added to the existing library.

In this PR, `_add_library_if_not_exist` is added, which lowercases library names to check if it exists. Furthermore, if the library exists already, adding builtins is skipped, instead of mixing the content.

I didn't find any function in the API for retrieving the names of the libraries declared already. Hence, I used `[library.lower() for library in self._vunit_obj._project._libraries]`. However, we might want to add it in the project class.

### DIFF
--- a/vunit/builtins.py
+++ b/vunit/builtins.py
@@ -11,8 +11,13 @@ Functions to add builtin VHDL code to a project for compilation
 from pathlib import Path
 from glob import glob
 from warnings import warn
+import logging
+
 from vunit.vhdl_standard import VHDL, VHDLStandard
 from vunit.ui.common import get_checked_file_names_from_globs
+
+
+LOGGER = logging.getLogger(__name__)
 
 VHDL_PATH = (Path(__file__).parent / "vhdl").resolve()
 VERILOG_PATH = (Path(__file__).parent / "verilog").resolve()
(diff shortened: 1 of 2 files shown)


vunit/builtins.py
```diff
@@ -130,16 +135,26 @@ def _add_verification_components(self):
...
         Add osvvm library
         """
-        library_name = "osvvm"
-
-        try:
-            library = self._vunit_obj.library(library_name)
-        except KeyError:
-            library = self._vunit_obj.add_library(library_name)
+        library = self._add_library_if_not_exist(
+            "osvvm", "Library 'OSVVM' previously defined. Skipping addition of builtin OSVVM (2021.09)."
```
LarsAsplund, line 154: How can we make sure that we will remember to update this message when OSVVM is bumped?

umarcor: I'm not sure... By navigating the codebase through the GitHub web, you don't get any explicit indication that OSVVM is bumped to a tag. It just points to the commit: https://github.com/OSVVM/OSVVM/tree/71faba779aebe79d3a48ce0abc0a9adc272f99ae.

We might use a test for checking out the history of OSVVM, using git for comparing the commit hash with the tags and find an exact match. Then, in the `builtins.py`, have the version moved to a variable which the check can import for comparison. However, I'm not sure that complexity is worth. Particularly, whether that should be part of any "regular" test.

We might just add a subsection to http://vunit.github.io/contributing.html about "How to bump OSVVM", explaining that we want to bump to tagged commits only, and that whoever does it should remember to update the string.

When dependabot or some other similar solution supports bumping submodules to tags only, we might actually need an automated procedure.

LarsAsplund: I think the instruction is good enough for now. If we find out we keep missing to update the string we can come up with another solution.

umarcor: This PR now adds a subsection to the docs about bumping submodules.

LarsAsplund (approved):

PR #775, opened 2021-10-30
Support hierarchical library mapping in ModelSim

ThomasWismer:
The library section of the modelsim.ini file allows to refer to a subordinate INI file using an "others" clause (subordinate files may also contain an "others" clause). The current implementation for collecting all mapped libraries ignores the "others" clause. With this change, all library mappings from all subordinate files are included.

### DIFF
--- a/vunit/sim_if/modelsim.py
+++ b/vunit/sim_if/modelsim.py
@@ -218,13 +218,17 @@ class ModelSimInterface(VsimSimulatorMixin, SimulatorInterface):  # pylint: disa
         cfg.set("Library", library_name, path)
         write_modelsimini(cfg, self._sim_cfg_file_name)
 
-    def _get_mapped_libraries(self):
+    def _get_mapped_libraries(self, cfg_file_name=None):
         """
         Get mapped libraries from modelsim.ini file
         """
-        cfg = parse_modelsimini(self._sim_cfg_file_name)
+        cfg_file_name = cfg_file_name if cfg_file_name is not None else self._sim_cfg_file_name
+        cfg = parse_modelsimini(cfg_file_name)
         libraries = dict(cfg.items("Library"))
         if "others" in libraries:
+            other_libraries = self._get_mapped_libraries(libraries["others"])
+            # current mappings take precedence over others
+            libraries = {**other_libraries, **libraries}
             del libraries["others"]
         return libraries
 


vunit/sim_if/modelsim.py
```diff
@@ -218,13 +218,17 @@ def create_library(self, library_name, path, mapped_libraries=None):
         cfg.set("Library", library_name, path)
         write_modelsimini(cfg, self._sim_cfg_file_name)
 
-    def _get_mapped_libraries(self):
+    def _get_mapped_libraries(self, cfg_file_name=None):
```
umarcor, line 221: Is the addition of this argument related to the description of the PR? It seems to be a different feature.

vunit/sim_if/modelsim.py
```diff
@@ -218,13 +218,17 @@ def create_library(self, library_name, path, mapped_libraries=None):
...
         Get mapped libraries from modelsim.ini file
         """
-        cfg = parse_modelsimini(self._sim_cfg_file_name)
+        cfg_file_name = cfg_file_name if cfg_file_name is not None else self._sim_cfg_file_name
+        cfg = parse_modelsimini(cfg_file_name)
         libraries = dict(cfg.items("Library"))
         if "others" in libraries:
+            other_libraries = self._get_mapped_libraries(libraries["others"])
+            # current mappings take precedence over others
+            libraries = {**other_libraries, **libraries}
```
umarcor, line 231: Can this be a single line?

```py
libraries = {**(self._get_mapped_libraries(libraries["others"])), **libraries}
```

vunit/sim_if/modelsim.py
ThomasWismer: This allows the same method to be used for subordinate INI files, see recursive call below on line 229.

vunit/sim_if/modelsim.py
ThomasWismer: Sure. I'd prefer two statements for readability, but I can change this if you like.

vunit/sim_if/modelsim.py
umarcor: Ok, I overlooked that the call was recursive.

vunit/sim_if/modelsim.py
umarcor: It's ok. Just checking that `other_libraries` is not used somewhere else (it's local to the if cluse).

dalex78:
@LarsAsplund this as been reviewed by Umarcor, I suggest to merge and close.

LarsAsplund:
I think we should probably add a test for this.

dalex78:
You are right. I don't mind working on that but I do not understand how the tests work. I guess I need to modify the `tests/unit/test_modelsim_interface.py` but where are specified the content of the tested modelsim.ini ? 
If anyone can explain me how the tests are working I would appreciate.

LarsAsplund:
@dalex78 The tests create their own ini files. These doesn't have to be real and complete files but just  enough to verify that VUnit handles the files as expected. For example, the test setup creates a file containing a single line: https://github.com/VUnit/vunit/blob/fbb46c60b3b6b150bfaeb6e4fb30afa8ccdaf5a1/tests/unit/test_modelsim_interface.py#L328

Some tests, i.e. https://github.com/VUnit/vunit/blob/fbb46c60b3b6b150bfaeb6e4fb30afa8ccdaf5a1/tests/unit/test_modelsim_interface.py#L257, don't even use valid ini files to verify behavior

In this case you would need to create ini files containing the others clause and verify that VUnit collects the information correctly.

ThomasWismer:
As the author of this pull request, I felt obliged to submit a unit test. @dalex78 Thanks anyway for your offer to take on this "burden".

std-max:
@LarsAsplund This one might be merged aswell

PR #779, opened 2021-11-14
docs: add section Overview, including a diagram

umarcor:
This PR adds subsection 'Overview' to section 'What is VUnit?'. A diagram based on [umarcor.github.io/osvb/intro](https://umarcor.github.io/osvb/intro) is included.

image


LarsAsplund:
I think we need to elaborate a bit more on what the VHDL libraries are. The existing main features section is a bit shallow and do not mention datatypes, com, verification etc. Without that it becomes a bit unclear what's in the boxes of the new overview. Other than that I think the picture is good.

umarcor:
I updated and reordered the "Main features" section:

image

LarsAsplund:
The only thing I'm missing is a line saying that the location preprocessor isn't needed if the simulator has VHDL-2019 call path support.

Issue #772, opened 2021-10-27
Rethinking the runner configuration interface: integration of the OSVVM methodology into VUnit


umarcor:
The most popular open source VHDL verification projects are cocotb, OSVVM and UVVM, along with VUnit. As discussed in [larsasplund.github.io/github-facts](https://larsasplund.github.io/github-facts/), there are some philosophical differences between them: OSVVM and UVVM are defined as "VHDL verification methodologies", cocotb is for (Python) co-routine co-simulation, and VUnit is a Python-aided framework of HDL utilities. Naturally, there are some overlapping capabilities because all of them provide basic features such as logging and building/simulation. Therefore, methodologies can be seen as bundles of utilities (frameworks), and some users might refer to using the VUnit runner as a methodology. Nonetheless, it is in the DNA of VUnit to be non-intrusive and allow users to pick features one by one, including reusing the methodologies they are used to.

Currently, it is possible to use OSVVM utilities/libraries in a VUnit environment. Although there are still some corner cases to fix (#754, #767, #768), it is usable already. In fact, some of VUnit's features do depend on OSVVM's core. However, it is currently not possible to use the OSVVM methodology as-is within VUnit. The OSVVM methodology uses top-level entities without generics or ports, and the entrypoints are VHDL configurations. Meanwhile, VUnit needs a top-level generic of type `string` in order to pass data from Python to VHDL.

Most simulators do support calling a configuration instead of an entity as the primary simulation unit. It should, therefore, be trivial to support OSVVM's configurations as entrypoints in VUnit. I am unsure about VUnit's parser supporting configurations in the parser and dependency scanning features; but that should not be the main challenge anyway.

The main challenge we need to address is that passing generics to VHDL Configurations is not supported in the language. If that was possible, the runner string might be forwarded to the entiy within the configuration. For the next revision, we might propose enhancements in this regard, since revisiting the limitations of configurations is one of the expected areas to work on. Nevertheless, that will/would take several months or years until made available in simulators.

Yesterday, I had a meeting with @JimLewis and he let me know that he's been thinking about implementing some mechanism for passing data between the TCL scripts (`.pro` files) and the VHDL testbenches. We talked about `.ini`, `.yml` and `.json`, and I suggested to use the latter because there is a JSON reader library available already: [Paebbels/JSON-for-VHDL](https://github.com/Paebbels/JSON-for-VHDL). In fact, JSON-for-VHDL is submoduled in VUnit, in order to pass very complex generics to the testbench.

[...]

Issue #179, opened 2016-07-08
Simulate VHDL configurations


imd1:
There's currently a "entity" method in the library class, which I use to get a test handle

tb_vunit = lib.entity("tb_vunit")
test = tb_vunit.test("Test")

I have examples of a single VHDL test entity with 2 VHDL configurations defining 2 different tests, and so I'd to use a library "configuration" method to access each configuration and hence the test handle for each configuration.

Is this possible in a future release?

kraigher:
First of all configurations would need to be able to be considered as runnable test bench top levels which they are not today. After that it would make sense to add a Python method you describe.

joshrsmith:
I would also be interested in this functionality -- to have a configuration be a runnable top level for VUnit.

joshrsmith:
I have been trying to figure out if there is a work-around for this issue.

I thought you might be able to make a wrapper testbench and instantiate a desired "top level" testbench configuration in the architecture of the wrapper. But the problem then is that I do not know how to get the _runner_cfg_ from the testbench wrapper generics to push down through the instantiated configuration.

MortenZdk:
Hi Josh, 
Please find an example attached that I made while trying to do a work-around.  
Best regards Morten

[VHDL - config_tryout.zip](https://github.com/VUnit/vunit/files/893158/VHDL.-.config_tryout.zip)

kraigher:
The work around for this would be to create a wrapper test bench and use the scan_tests_from_file public method.

LarsAsplund:
Implemented by #951

PR #784, opened 2021-12-30
Configuration support

LarsAsplund:
Related to #772 and #179

Supports automatic discovery of VHDL configurations of testbench entities. These are added as VUnit configurations to the testbench. These VUnit configurations have a newly introduced `vhdl_configuration_name` attribute that, if set, makes VUnit call the simulator with the VHDL configuration name rather than the testbench entity + architecture names.

Currently VUnit doesn't allow an existing VUnit configuration to be modified. Since a VUnit configuration based on a VHDL configuration is added automatically during the discovery phase there is no way to extend that VUnit configuration with `pre_config` and `post_check` functions. Removing this limitation should be added before closing this PR.


LarsAsplund:
Branch broken. Closing this PR to open a new one

PR #792, opened 2022-01-11
Fix port type starting with signal

### LINKED ISSUE #791: parser fails to accept port types starting with signal
It seems I have found a parsing issue: I have a block declaring a port input with a type whos name starts with "signal". As soon as I change the port type name to not start with signal it works.
I would like to debug this, but I could not find any Python code tests. Are there unit tests for the Python functions?

eschmidscs:
Trying to fix #791.
I don't know/understand the replacement of "signal", so the padding with a space might not be the correct solution.


vunit/vhdl_parser.py
```diff
@@ -637,7 +637,7 @@ def parse(cls, code, is_signal=False):
         """
         if is_signal:
             # Remove 'signal' string if a signal is beeing parsed
```
LarsAsplund, line 639: beeing -> being while we at it

tests/unit/test_vhdl_parser.py
```diff
@@ -298,7 +298,8 @@ def test_parsing_entity_with_ports(self):
             """\
 entity name is
    port (clk : in std_logic;
-         data : out std_logic_vector(11-1 downto 0));
+         data : out std_logic_vector(11-1 downto 0);
+         data2 : in signal_type);
```
LarsAsplund, line 302: You should also add a port which starts with `signal name : ...`. Not often that you that but it's legal. Then you should add one port that has a tab character or another whitespace between `signal` and `name` :smile:

LarsAsplund (changes requested):

tests/unit/test_vhdl_parser.py
eschmidscs: Good points - but before I try to catch the rest of the world here: What is the intention of removing the "signal" string in this function? I think this is parsed as a port declaration, so no signal keyword here anyway, right?

eschmidscs: Added a regex to catch this quite generic.

dalex78:
@LarsAsplund, this looks good to me.

dalex78:
I did not see the ''change requested". As I do not believe I can directly modify this PR, I will create a new one with the requested changes.

eschmidscs:
@dalex78 I think I have fixed the requests - that's why I requested another review.

dalex78:
Unless I am mistaken, I see in test_vhdl_parser.vhd:
```vhdl
port (clk : in std_logic;
         data : out std_logic_vector(11-1 downto 0));
         data : out std_logic_vector(11-1 downto 0);
         signal_data2 : in std_logic;
         data3 : in signal_type;
         data4_signal : in std_logic;
         data5 : in type_signal
);
```

And Lars was asking to add something like that:
```vhdl
entity name is
port (
    clk : in std_logic;
    data : out std_logic_vector(11-1 downto 0);
    signal_data2 : in std_logic;
    data3 :\tin signal_type;
    data4_signal : in std_logic;
    data5\t: in type_signal;
    signal clk2 : in std_logic;
    signal\tdata7 : out std_logic_vector(11-1 downto 0);
    signal signal_data8 : in std_logic;
    signal data9 : in signal_type;
    signal data10_signal :\tin std_logic;\t
    signal data11 : in type_signal
);
end entity;
```
(I am not sure I the '\t' can be added like that.)

dalex78:
Ypu can see the changes here also: https://github.com/dalex78/vunit/tree/port_type_signal

dalex78:
@LarsAsplund I suggest to close without merging this PR.
The *new* PR is #826.

PR #797, opened 2022-01-18
Fix axi_lite_master wait behaviour if idle

eschmidscs:
The axi_lite_master did wait forever on wait_for_idle if it was already idle.
I tried to fix it here.

### DIFF
--- a/vunit/vhdl/verification_components/src/axi_lite_master.vhd
+++ b/vunit/vhdl/verification_components/src/axi_lite_master.vhd
@@ -50,6 +50,7 @@ end entity;
 
 architecture a of axi_lite_master is
   constant reply_queue, message_queue : queue_t := new_queue;
+  signal idle : boolean := true;
 begin
   main : process
     variable request_msg : msg_t;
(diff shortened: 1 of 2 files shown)


umarcor:
> If I understand the failing task, it is not related to my change. Can anyone check this please?

It seems that Sphinx was updated to v.4.4.0 and it is now showing that error. Please, replace the reference reported in the error with `:vunit_example:`vhdl/array`.

LarsAsplund:
Looks good from my point of view. The only alternative solution I can come to think of is to push an extra dummy "token" into the queue after the request message and then pop that at the end of the bus process. The component is idle if the queue is empty. I can't say that one is better or more beautiful than the other so you can decide.

eschmidscs:
@umarcor Tried to do that, the new link does not seem to work. A similar error shows up now, but I can't even find that static link in examples.rst?

umarcor:
@eschmidscs I fixed doc issues in `master`. If you rebase, it should be green.

creiter64:
I'm hitting this error, can this PR be merged?

Issue #798, opened 2022-02-10
Question about axi_stream_protocol_checker rule 5.


bewimm:
[Rule 5](https://github.com/VUnit/vunit/blob/master/vunit/vhdl/verification_components/src/axi_stream_protocol_checker.vhd#L115) of the AXI stream protocol checker ensures that `tdata` is valid when `tvalid` is asserted.
I am wondering if that should also be the case when not all bytes are valid (i.e. `tkeep /= "1111"`).

For example the following trace results in an error:
`#      62490000 ps - master:rule 5            -   ERROR - Not unknown check failed for tdata when tvalid is high - Got ----_----_----_----_----_----_1111_0000.`
![image](https://user-images.githubusercontent.com/3818590/153488807-4fb3cfcd-7f76-4e7f-805d-ce02de3c4fe1.png)

The workaround is trivial, but I like setting unused byte lanes to '-' to easily see downstream issues.

Is the current behavior of the checker intentional or just an oversight?

umarcor:
I'd say it's an oversight. `tkeep` should be taken into account.

LarsAsplund:
The rules are created from ARM's own document "AMBA® 4 AXI4™, AXI4-Lite™, and
AXI4-Stream™ Protocol Assertions" (https://documentation-service.arm.com/static/5f106bd90daa596235e808ed?token=). Nevertheless, I agree that this looks like an oversight. I will give it some thinking to see if there are some drawbacks of making this change.

LarsAsplund:
I think the problem here, and the reason why ARM made the rule the way they did, is that `tkeep` is an optional signal, i.e. the receiver of `tdata` containing null bytes (the bytes with `-` in your case) may not support `tkeep` and can't ignore any byte.

I see that the suggested change is a useful extension but it should be made an optional extension, `allow_x_in_null_bytes`

Also note that there can be null bytes if `tkeep` is high and `tstrb` is low. That scenario should also be covered by such an extension..

bewimm:
> I think the problem here, and the reason why ARM made the rule the way they did, is that `tkeep` is an optional signal, i.e. the receiver of `tdata` containing null bytes (the bytes with `-` in your case) may not support `tkeep` and can't ignore any byte.
> 
> I see that the suggested change is a useful extension but it should be made an optional extension, `allow_x_in_null_bytes`

Can you describe how you imagined this to work some more? I was looking for an existing component that implements an extension or something similar but couldn't find anything.
[...]

PR #799, opened 2022-02-11
Stream checker rule5 fix

bewimm:
Fix for #798

### DIFF
--- a/vunit/vhdl/verification_components/src/axi_stream_pkg.vhd
+++ b/vunit/vhdl/verification_components/src/axi_stream_pkg.vhd
@@ -34,39 +34,42 @@ package axi_stream_pkg is
   type axi_stream_component_type_t is (null_component, default_component, custom_component);
 
   type axi_stream_protocol_checker_t is record
-    p_type        : axi_stream_component_type_t;
-    p_actor       : actor_t;
-    p_data_length : natural;
-    p_id_length   : natural;
-    p_dest_length : natural;
-    p_user_length : natural;
-    p_logger      : logger_t;
-    p_max_waits   : natural;
+    p_type                      : axi_stream_component_type_t;
+    p_actor                     : actor_t;
+    p_data_length               : natural;
+    p_id_length                 : natural;
+    p_dest_length               : natural;
+    p_user_length               : natural;
+    p_logger                    : logger_t;
+    p_max_waits                 : natural;
+    p_allow_x_in_non_data_bytes : boolean;
   end record;
 
   constant null_axi_stream_protocol_checker : axi_stream_protocol_checker_t := (
-    p_type        => null_component,
-    p_actor       => null_actor,
-    p_data_length => 0,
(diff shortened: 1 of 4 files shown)


vunit/vhdl/verification_components/test/tb_axi_stream_protocol_checker.vhd
```diff
@@ -42,7 +42,7 @@ architecture a of tb_axi_stream_protocol_checker is
   constant logger           : logger_t                      := get_logger("protocol_checker");
   constant protocol_checker : axi_stream_protocol_checker_t := new_axi_stream_protocol_checker(
     data_length => tdata'length, id_length => tid'length, dest_length => tdest'length, user_length => tuser'length,
-    logger => logger, actor => new_actor("protocol_checker"), max_waits => max_waits
+    logger => logger, actor => new_actor("protocol_checker"), max_waits => max_waits, allow_x_in_null_bytes => true
```
bewimm, line 45: I'm not very happy with this solution.
I think it would be better to not mix the extension with the rest, but I'm not sure what the cleanest way of doing that is

vunit/vhdl/verification_components/src/axi_stream_pkg.vhd
```diff
@@ -469,24 +473,26 @@ package body axi_stream_pkg is
...
-      actor       : actor_t  := null_actor;
-      max_waits   : natural  := 16
+      data_length           : natural;
+      id_length             : natural  := 0;
+      dest_length           : natural  := 0;
+      user_length           : natural  := 0;
+      logger                : logger_t := axi_stream_logger;
+      actor                 : actor_t  := null_actor;
+      max_waits             : natural  := 16;
+      allow_x_in_null_bytes : boolean := false
```
LukasVik, line 483: Align 

```suggestion
      allow_x_in_null_bytes : boolean   := false
```

vunit/vhdl/verification_components/src/axi_stream_protocol_checker.vhd
```diff
@@ -69,6 +69,20 @@ architecture a of axi_stream_protocol_checker is
...
   signal not_tvalid  : std_logic;
+
+  signal tdata_normalized : std_logic_vector(tdata'range);
+
+  function normalize_tdata(data, strb, keep : std_logic_vector) return std_logic_vector is
+    variable ret : std_logic_vector(data'range);
+  begin
+    ret := data;
+    for i in keep'range loop
+      if keep(i) = '0' and strb(i) = '0' then
```
LukasVik, line 80: Per discussion in #798 this should perhaps be

```suggestion
      if keep(i) = '0' or strb(i) = '0' then
```

LukasVik:
I looked through it again, and after @bewimm's fixes it looks all good to me. I agree with the nomenclature and functionality, per the discussion in #798, and I agree with the implementation. IMHO ready to merge.

PR #802, opened 2022-02-18
Drive invalid in AXI VCs

LukasVik:
When the 'valid' signal is low for an AXI (-Lite) channel, the signals of the channel should be driven with 'X'.
They should not keep their value from the last time the channel was valid.
This can hide errors in the DUT if the DUT samples values in the wrong clock cycle.

This change is largely inspired by the AXI-Stream VCs.

### DIFF
--- a/vunit/vhdl/verification_components/src/axi_lite_master.vhd
+++ b/vunit/vhdl/verification_components/src/axi_lite_master.vhd
@@ -19,10 +19,12 @@ context work.com_context;
 context work.vunit_context;
 
 entity axi_lite_master is
-  generic(
-    bus_handle : bus_master_t
+  generic (
+    bus_handle : bus_master_t;
+    drive_invalid : boolean := true;
+    drive_invalid_val : std_logic := 'X'
   );
-  port(
+  port (
     aclk : in std_logic;
 
     arready : in std_logic;
@@ -45,13 +47,15 @@ entity axi_lite_master is
 
     bvalid : in std_logic;
     bready : out std_logic := '0';
-    bresp : in axi_resp_t := axi_resp_okay);
+    bresp : in axi_resp_t := axi_resp_okay
+  );
 end entity;
 
 architecture a of axi_lite_master is
   constant reply_queue, message_queue : queue_t := new_queue;
   signal idle : boolean := true;
 begin
+
   main : process
     variable request_msg : msg_t;
     variable msg_type : msg_type_t;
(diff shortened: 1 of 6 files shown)


vunit/vhdl/verification_components/src/axi_lite_master.vhd
```diff
@@ -70,73 +74,116 @@ begin
...
+    variable wdata_this_transaction : std_logic_vector(wdata'range) := (others => '0');
   begin
-    wait until rising_edge(aclk) and not is_empty(message_queue);
+    -- Initialization
+    drive_ar_invalid;
+    drive_aw_invalid;
+    drive_w_invalid;
+
+    loop
+      wait until rising_edge(aclk) and not is_empty(message_queue);
```
LukasVik, line 114: This is mainly an indentation change due to the ``loop`` statement. Github renders i really bad.

LukasVik:
Note that this changes the default behavior of the three VCs, so in some sense it is a breaking change. This could be changed by setting the default value `drive_invalid` generic to `false`. Though having it set to `true` is symmetric with the AXI-Stream VCs. And in my personal opinion it should be enabled in pretty much all cases, to get a stronger test.

vunit/vhdl/verification_components/test/tb_axi_lite_master.vhd
```diff
@@ -370,6 +370,31 @@ begin
     end if;
   end process;
 
+  check_not_valid : process
+    constant a_addr_invalid_value : std_logic_vector(araddr'range) := (others => 'X');
+    constant wdata_invalid_value : std_logic_vector(wdata'range) := (others => 'X');
+    constant wstrb_invalid_value : std_logic_vector(wstrb'range) := (others => 'X');
+    constant bresp_invalid_value : std_logic_vector(bresp'range) := (others => 'X');
```
LudvigVidlid, line 377: Unused here

LudvigVidlid:
Looks good to me! Added a small comment on an unused constant.

vunit/vhdl/verification_components/test/tb_axi_lite_master.vhd
LukasVik: Fixed, thanks!

LukasVik:
@LarsAsplund Sorry to ping you like this, but would it be possible to merge this? It has been reviewed by Ludvig.

std-max:
LGTM too.
I would even say that this should not be a generic settings as it should be the default behaviour. Maybe for v5.0?
But for the next minor release, yeah we need that generic.

LarsAsplund:
@LukasVik Sorry for the delay. It's in the pipe but unfortunately I haven't had time to work on that pipe. Been more of that lately so I will deal with shortly... I hope

LukasVik:
@LarsAsplund No problem at all, I know how it is.

@std-max Well with the default value `true` of the `drive_invalid` generic it **is** the default behavior. Which is a breaking change since the ports used to keep their value before. The generic configuration, and the default behavior, is now symmetric with the AXI-Stream verification components.

With that said, I agree with you in the sense that I personally will always leave the `drive_invalid` generic as `true`. Off the top of my head I can not think of a scenario where I would not. But perhaps I am missing something.

LukasVik:
I rebased the source branch upon upstream master in order to solve a merge conflict. No functional changes apart from that. Is ready for review/merge.

PR #804, opened 2022-03-06
Add feature to select simulator progamatically

nfrancque:
Add `set_simulator` method exposed to user for simpler switching between simulators


nfrancque:
Closing due to starting off very old version of master, will be recreated

PR #805, opened 2022-03-06
Add feature to select simulator programatically 

nfrancque:
Add set_simulator method exposed to user for simpler switching between simulators

### DIFF
--- a/tests/unit/test_ui.py
+++ b/tests/unit/test_ui.py
@@ -1143,6 +1143,16 @@ end architecture;
         ui = self._create_ui()
         self.assertEqual(ui.get_simulator_name(), "mock")
 
+    def test_set_simulator(self):
+        ui = self._create_ui()
+        prefix = ui.get_simulator_prefix()
+        # Just change it
+        prefix = Path("mock")
+        ui.set_simulator("modelsim", "modelsim_2021_2", prefix)
+        self.assertEqual(ui.get_simulator_name(), "modelsim")
+        self.assertEqual(ui.get_simulator_id(), "modelsim_2021_2")
+        self.assertEqual(ui.get_simulator_prefix(), prefix)
+
     def _create_ui(self, *args):
         """Create an instance of the VUnit public interface class"""
         with mock.patch(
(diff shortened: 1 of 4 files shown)


nfrancque:
@LarsAsplund Please review.  I don't have much experience with tox but it seems natural to change the environment variable settings in the `pyproject.toml` for the acceptance tests to use the new interface.  I ran a quick few tests locally adding `set_simulator` in the `run.py` of `acceptance/artificial`.  Seems to work but wasn't sure what the best way would be.

LarsAsplund:
Today the selection of simulator is done in two steps based on the user's PATH and VUNIT_SIMULATOR:

1. The simulator class is selected early when the VUnit object is created and then there are a number of other things based on that selection. https://github.com/VUnit/vunit/blob/master/vunit/ui/__init__.py#L156-L176
2. The simulator instance is created late when `main` has been called (https://github.com/VUnit/vunit/blob/master/vunit/ui/__init__.py#L775)

Rearranging these event makes it harder to see if there are changes in behavior that would break backward compatibility. For that reason I suggest that the programmatic approach to changing the simulator is the same as doing step 1 and that step 2 remains where it is today. You may need to hold on to some of the arguments in the `set_simulator` call until the simulator instance is created in step 2.

Note that `set_simulator` must be called early after the VUnit object is created. At least `add_library` uses ` `_simulator_output_path`  which is updated in step 1.

GlenNicholls:
One request I have is to be able to also select the simulator from the CLI. When I'm debugging, it's often that I change the simulator from the command line using bash aliases. However, it'd be nice to have the ability to do this from the CLI instead of in the `run.py` file all the time.

My use-cases are:
1. Debugging simulator bugs/issues when they pop up and other misc things that one simulator reports better than others
2. Running tests that require encrypted IP be compiled. E.g. using GHDL for parallelizing all generic unit tests, then running modelsim to run tests that use e.g. PCIe core
3. Running with e.g. modelsim+gui locally, and configuring CI to use GHDL. In that case, it's simpler to just have `--sim modelsim|ghdl|rivierapro|...`. I could probably add a custom CLI arg, but this seems like something VUnit should support directly

For 2., I would probably just use Python to run with two different simulators for all the tests to run back-to-back. But, I switch often enough that it'd be nice to change from the CLI as well.

PR #820, opened 2022-04-26
Handle PermissionError while listing available simulators

### LINKED ISSUE #819: VUnit does not handle file permission exception
When I was running VUnit, I got the following error (with WSL)
`PermissionError: [Errno 13] Permission denied: '/mnt/c/Windows/system32/config/systemprofile/AppData/Local/Microsoft/WindowsApps/vsim'`

At the start of a run.py script, VUnit checks for the available simulators by looking into each path in the PATH variable.
However, sometimes we do not have permission to read a path. And it fails :'(

I will file a PR today.

std-max:
Fix #819

### DIFF
--- a/vunit/sim_if/__init__.py
+++ b/vunit/sim_if/__init__.py
@@ -316,7 +316,10 @@ def isfile(file_name):
     Case insensitive Path.is_file()
     """
     fpath = Path(file_name)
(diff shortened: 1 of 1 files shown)


umarcor (approved):
LGTM!

PR #824, opened 2022-05-10
Detect dual test case run

### LINKED ISSUE #795: Detect dual test case run
Issue #794 highlighted an improper use of VUnit where a single test case is run twice. The bad side effects of such use was fixed in that issue but the root cause is still not flagged as an error. Adding such detection of dual test case runs is the purpose of this issue.

std-max:
Fix #795

Prevent from calling twice the same test case.
Reset the number of run tests when test_runner_setup is called.
Add tests.


LarsAsplund:
The error detection looks good but I don't think we should make a the ability to re-run a test case (by calling `test_runner_setup` a second time) a new feature. It could be something that works just because of the way our procedures and functions are designed but the proposed error message is advertising it as a feature.

I'm afraid that unorthodox use of VUnit might lead into unforeseen complexities (just as calling a test case twice did) so I rather keep the core of VUnit structure very clean. Did you have a specific use case for this in mind?

std-max:
@LarsAsplund In fact I added a test that calls `test_runner_setup` just because it was done in a previous test like the one at line 318 (`banner("Should loop over enabled_test_case once and in order unless re-initialized.");`
I can remove that of course as I agree with you this might be unusual !

LarsAsplund:
@std-max Ok, I understand. I can't really recall if that test was added with a feature in mind or just to make sure that `test_runner_setup` is well-designed. After all, it is reasonable to expect that a setup works regardless of what happened before it. Still, I wouldn't like to make it a supported feature so if you could remove that part it would be great.

std-max:
Done @LarsAsplund :)

std-max:
No joke, I literally fell asleep while waiting for the checks to pass 
You can now merge @LarsAsplund haha

PR #826, opened 2022-05-10
Port type signal

dalex78:
The changement are those implemented by @eschmidscs in pull request #792 with the changes requested by Lars.
Note that there is also a modification of the regex from:
```python
code = re.sub("\bsignal\b", "", code)
```
to
```python
code = re.sub(r"\bsignal\b", "", code)
```
otherwise the tests fail. The [python documentation about the word boundary](https://docs.python.org/3/howto/regex.html#more-metacharacters) states:
> There are two subtleties you should remember when using this special sequence. 
> First, this is the worst collision between Python’s string literals and regular expression sequences. In Python’s string literals, \b is the backspace character, ASCII value 8. If you’re not using **raw strings**, then Python will convert the \b to a backspace, and your **RE won’t match as you expect it to**.
> Second, inside a character class, where there’s no use for this assertion, \b represents the backspace character, for compatibility with Python’s string literals.

### DIFF
--- a/tests/unit/test_vhdl_parser.py
+++ b/tests/unit/test_vhdl_parser.py
@@ -403,6 +403,7 @@ end entity;
         self.assertEqual(ports[11].subtype_indication.code, "type_signal")
         self.assertEqual(ports[11].subtype_indication.type_mark, "type_signal")
         self.assertEqual(ports[11].subtype_indication.constraint, None)
+
     def test_parsing_simple_package_body(self):
         package_body = self.parse_single_package_body(
             """\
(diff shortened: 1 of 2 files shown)


dalex78:
The 517 tests all pass (using `tox -v -e py39-unit-modelsim`).

dalex78:
@LarsAsplund I suggest to merge and close this PR and to close without merging #792.

eschmidscs:
@dalex78 Thanks for fixing the fix and properly testing it!

umarcor (approved):
LGTM

PR #827, opened 2022-05-11
OSVVM: bump to 2022.04

umarcor:
This PR bumps the OSVVM submodule in VUnit.


g0t00:
@umarcor This is shown as being released as in 4.7.0 in the release notes.
If i manually checked out the v4.7.0 tag, It still seems to have the submodule set to 2021.12.
The same seems to be true for the released pip module.

umarcor:
@g0t00 you are correct. We unwantedly reverted it back to 2021.12 (https://github.com/VUnit/vunit/commit/04622a76634648f775c808887c0556826b27e1dd#diff-8eabb1bbd50f80cf0a09c390b7f45cdd4564234e9a074a0e5178cb864ccc7692). I guess that @LarsAsplund had that version checked out locally and did `commit -a` without noting it.

I pushed an update to branch `master`. @g0t00, does that work for you or do you want us to push v4.7.1 to PyPI with this bugfix?

g0t00:
@umarcor Thanks a lot for the quick answer. I think thats fine, luckily pip can directly track git commits hashes, so this is pinned for now in our CI setup.

umarcor:
@g0t00 if you are using "git features" to pin requirements in pip, you might want to use `https://github.com/VUnit/vunit/archive/master.zip` (as in https://github.com/VHDL/Compliance-Tests/blob/main/requirements.txt) or `https://github.com/VUnit/vunit/archive/v4.7.zip` (which is a branch, not a tag, so it will move forward if more bugfixes are applied).

Issue #830, opened 2022-05-14
Extend `dict_t` to handle any type


LarsAsplund:
Just like queue the `dict_t` should support values of any standard and IEEE type in a type safe manner.

LarsAsplund:
For v5.0.0 the overloaded subprograms should be added that create ambiguity in current code.

PR #835, opened 2022-05-30
Make dict_t type generic

LarsAsplund:
Solves #830


std-max:
That's going to be really useful!
So are you going to merge it for 4.7.0? or all at once with overloaded functions for 5.0?

LarsAsplund:
@std-max I also added support for sending dict_t in com messages. Once this passes all checks I will merge to the master branch. For v5.0 I will add the overloaded subprograms that create potential ambiguity in existing programs using dicts with the original string value support

Issue #808, opened 2022-03-23
vhdl_parser.py: components should not be converted to lower-case while instantiating Verilog modules


krishnan-gopal:
I have a case where a VHDL entity (say vhdl_top) is instantiating a Verilog module with upper-case letters (say VERILOG_MODULE):

VERILOG:

```
module VERILOG_MODULE (
.....
endmodule
```

The component declaration and instantiation of this would be done in the vhdl_top:

VHDL:
```
component VERILOG_MODULE
...
end component;

inst_module: VERILOG_MODULE
....
```

The vhdl_parser.py in VUnit is parsing the vhdl_top, and converts the component declaration of VERILOG_MODULE to lower-case assuming that it is a VHDL component. However, the Verilog module name remains unconverted in case. 
Therefore, we get warnings from the VUnit parser that look like this:

> WARNING - vhdl_top.vhd: failed to find a primary design unit 'verilog_module' in library 'verilog_lib'

Since Verilog is case-sensitive, this creates a problem for VUnit to find the right dependencies and results in a wrong compilation order.

asicnet:
Hello
A suggested solution would be:
When parsing the Verilog module, also convert the module name to lowercase and thus resolve the dependencies. 
Since Verilog is case sensitive, if the module exists in Verilog with different "same" names, an error will occur anyway when instantiating the module in VHDL.
A count in a dict module[name]+=1 and an error message if module[name]>1 would be sufficient and in case of no error the dependencies would be correct

BR Helmut

LarsAsplund:
As a workaround you can either change the Verilog name to lowercase or you can use the [add_dependency_on](http://vunit.github.io/py/vunit.html?highlight=add_dep#vunit.ui.source.SourceFile.add_dependency_on) method to enforce a correct compile_order.

krishnan-gopal:
Changing the verilog name to lowercase:
- This would be highly difficult while these are vendor/tool-generated models, and changing the case would mean also changing it in each instantiation(s).

add_dependency_on method:
- We use VUnit along with FuseSOC, meaning the run.py is generated. Doing this in the VUnit hooks/runner file would mean that every user should know these models/instances which come from the hierarchy of IPs designed by another user.

We (@asicnet and me) can make an attempt with a pull-request.

LarsAsplund:
[...]

PR #837, opened 2022-05-31
proposal for vhdl_parser to fix the issue #808 with an dummy  example…

asicnet:
…/mixed/* ; see special_storage.vhd line 392:394; each is runnig
cd  example/mixed/special_storage
python run.py --clean -f     ->  pf_dpsram_special.v is to late !!!

try all posibilities in  special_storage.vhd line 392:394;


std-max:
Hi!
You should update your master branch before filing the PR -> `This branch is 1 commit ahead, 1106 commits behind VUnit:master.`
It is hard to see what you have changed in the codebase.

PR #839, opened 2022-06-01
proposal for vhdl_parser to fix the issue #808

asicnet:
…/mixed/* ; see special_storage.vhd line 392:394; each is runnig

must comment line 25 in 

      vunit/vhdl/data_types/src/data_types_private_pkg.vhd

for succesfull test.


std-max:
@eine @LarsAsplund Do you think it is ok to have source code copyrighted Siemens Gmbh?

LarsAsplund:
@std-max @asicnet All code must have the license and the copyright we have today. The reason is that if we want to change the license, for example to a more modern and HW-oriented FOSS license, we can do that without tracking down all copyright holders to ask for permission.

License and copyright is also the only information we keep in the header. Things like author, dates, history etc is for Git to keep track of.

PR #858, opened 2022-08-21
Bugfix/axis slave back to back transfers

### LINKED ISSUE #573: Regression Bug in AXI-Stream slave when upgrading from v4.0.8

bundi78:
This PR shall fix the back-to-back transfer issue for axi-stream master/slave. This may fix #573.

### DIFF
--- a/vunit/vhdl/verification_components/src/axi_stream_master.vhd
+++ b/vunit/vhdl/verification_components/src/axi_stream_master.vhd
@@ -97,12 +97,15 @@ begin
         drive_invalid_output(tdata, tkeep, tstrb, tid, tdest, tuser);
       end if;
 
-      -- Wait for messages to arrive on the queue, posted by the process above
-      wait until rising_edge(aclk) and (not is_empty(message_queue) or areset_n = '0');
-
       if (areset_n = '0') then
         tvalid <= '0';
+        wait until areset_n = '1' and rising_edge(aclk);
       else
+        if is_empty(message_queue) then
+          -- Wait for messages to arrive on the queue, posted by the process above
+          wait until (not is_empty(message_queue) or areset_n = '0') and rising_edge(aclk);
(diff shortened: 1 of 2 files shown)


LarsAsplund (approved):

vunit/vhdl/verification_components/src/axi_stream_master.vhd
```diff
@@ -97,12 +97,15 @@ begin
...
-      -- Wait for messages to arrive on the queue, posted by the process above
-      wait until rising_edge(aclk) and (not is_empty(message_queue) or areset_n = '0');
-
       if (areset_n = '0') then
         tvalid <= '0';
+        wait until areset_n = '1' and rising_edge(aclk);
       else
+        if is_empty(message_queue) then
+          -- Wait for messages to arrive on the queue, posted by the process above
+          wait until (not is_empty(message_queue) or areset_n = '0') and rising_edge(aclk);
```
LarsAsplund, line 106: All this looks more correct than it was before but I see that the asynchronous reset is basically checked at the rising edge of the clock such that its assertion becomes synchronous rather than asynchronous. Deassertion is correctly synchronous. I think that can be fixed in a separate PR.

LarsAsplund (approved):

PR #866, opened 2022-11-01
Working Xcelium + Indago support with single command commpilation plus a basic rerun script for Xcelium

Hardolaf:
This work started from the work that @rodrigomelo9 presented in his PR: https://github.com/VUnit/vunit/pull/728.


vunit/sim_if/xcelium.py
```diff
@@ -0,0 +1,694 @@
...
+
+    def __init__(  # pylint: disable=too-many-arguments
+        self, prefix, output_path, gui=False, log_level=None, cdslib=None, hdlvar=None, use_indago=False
+    ):
+        SimulatorInterface.__init__(self, output_path, gui)
+        self._global_compile_options = {}
+        self._prefix = prefix
+        self._libraries = []
+        self._log_level = log_level
+        if cdslib is None:
```
GlenNicholls, line 121: Why check for None instead of doing:

```
if cdslib:
   ...
else:
   ...
```

Shorter if it fits reasonably in VUnit's line length limit (if there is one):

```
x = a if cdslib else b
```

vunit/sim_if/xcelium.py
```diff
@@ -0,0 +1,694 @@
...
+        except subprocess.CalledProcessError:
+            return None
+
+    def _create_cdslib(self):
+        """
+        Create the cds.lib file in the output directory if it does not exist
+        """
+        cds_root_virtuoso = self.find_cds_root_virtuoso()
+
+        if cds_root_virtuoso is None:
```
GlenNicholls, line 156: I've never understood driving TCL interfaces with Python strings. Stuff like this is way clearer when using Jinja templates as the flow can be followed in a single file with all the for, if/else, etc. constructs instead of following python strings between functions/methods

GlenNicholls: For multi-line strings, I like using `inspect.cleandoc` instead to avoid `\`'s and also to keep the string inline with the code

vunit/sim_if/xcelium.py
```diff
@@ -0,0 +1,694 @@
...
+            return self.compile_verilog_file_command(source_file)
+
+        LOGGER.error("Unknown file type: %s", source_file.file_type)
+        raise CompileError
+
+    def _compile_all_source_files(self, source_files_by_library: Dict[Library, List[SourceFile]], printer):
+        """
+        Compiles all source files and prints status information
+        """
+        try:
```
GlenNicholls, line 233: Is the output consistent with the rest of VUnit's reporting?

vunit/sim_if/xcelium.py
```diff
@@ -0,0 +1,694 @@
...
+        Return a command to compile all source files in a library
+        """
+        args = []
+        args += ["-makelib %s" % library.directory]
+        search_list = []
+
+        for source_file in source_files:
+            unique_args = []
+            if source_file.is_vhdl:
+                unique_args += ["%s" % self._vhdl_std_opt(source_file.get_vhdl_standard())]
```
GlenNicholls, line 334: f-string would be clearer

vunit/sim_if/xcelium.py
```diff
@@ -0,0 +1,694 @@
...
+                unique_args += source_file.get_compile_option("xcelium.xrun_verilog_flags", include_global=False)
+                for include_dir in source_file.include_dirs:
+                    if include_dir not in search_list:
+                        search_list.append(include_dir)
+                for key, value in source_file.defines.items():
+                    unique_args += ["-define %s=%s" % (key, value.replace('"', '\\"'))]
+
+            args += [source_file.name]
+            if len(unique_args) > 0:
+                # args += ["-filemap %s" % source_file.name]
```
GlenNicholls, line 347: It'd be nice to remove commented code if it's not needed anymore

vunit/sim_if/xcelium.py
```diff
@@ -0,0 +1,694 @@
...
+        Return a command to compile all source files
+        """
+        first_library = True
+        launch_gui = self._gui is not False
+        use_indago = self._use_indago is not False
+
+        cmd = str(Path(self._prefix) / "xrun")
+        args = []
+
+        args += ["-compile"]
```
GlenNicholls, line 377: It would be clearer to just do 

```
args = [
  "foo",
  "bar",
]
```

GlenNicholls (commented):
Whoops, didn't mean to start review.

vunit/sim_if/xcelium.py
```diff
@@ -0,0 +1,694 @@
...
+        )
+        write_file(argsfile, "\n".join(args))
+        return [cmd, "-F", argsfile, "-clean"]
+
+    def compile_vhdl_file_command(self, source_file):
+        """
+        Returns command to compile a VHDL file
+        """
+        cmd = str(Path(self._prefix) / "xrun")
+        args = []
```
GlenNicholls, line 431: It would be clearer to do:

```
args = [
  "foo",
  "bar",
]
```

vunit/ui/__init__.py
```diff
@@ -795,6 +818,11 @@ def _main_run(self, post_run):
         Main with running tests
         """
         simulator_if = self._create_simulator_if()
+        try:
+            simulator_if.set_global_compile_options(self._global_compile_options)
```
GlenNicholls, line 822: Why is every python exception caught and not the ones you care about? This could mask potentially important issues in the code that would make debugging difficult

tools/xcelium_verilog_fixup.py
```diff
@@ -0,0 +1,60 @@
...
+
+def main():
+    """
+    Remove xcelium incompatabilities from source code
+    """
+    where = "../vunit/verilog"
+    root = os.path.abspath(os.path.join(os.path.dirname(__file__), where))
+    for base, _, files in os.walk(root):
+        for file_name in files:
+            if file_name.endswith(".sv") or file_name.endswith(".svh"):
```
GlenNicholls, line 53: Why is `os.path` used here when `Path` is used elsewhere? Also, why not just use `Path.rglob`? I don't see any directory filtering here so I'm not sure why walking the file system is needed over just recursively globbing.

vunit/sim_if/incisive.py
```diff
@@ -38,22 +38,7 @@ class IncisiveInterface(SimulatorInterface):  # pylint: disable=too-many-instanc
...
-            "--cdslib",
-            default=None,
-            help="The cds.lib file to use. If not given, VUnit maintains its own cds.lib file.",
-        )
-        group.add_argument(
-            "--hdlvar",
-            default=None,
-            help="The hdl.var file to use. If not given, VUnit does not use a hdl.var file.",
-        )
+    # NOTE: Incisive shares the command-line arguments with Xcelium
```
GlenNicholls, line 41: This comment relates to the code that was removed, this should be in a commit message instead. I personally don't think this comment is needed, but I would clean it up to include `add_argument is excluded because X`

vunit/sim_if/xcelium.py
```diff
@@ -0,0 +1,694 @@
...
+            ]  # promote to error: "bad boolean literal in generic association"
+            args += [
+                "-xmerror EVBSTR"
+            ]  # promote to error: "bad string literal in generic association"
+            args += [
+                "-xmerror EVBNAT"
+            ]  # promote to error: "bad natural literal in generic association"
+            args += [
+                '-xmlibdirname "%s"' % (str(Path(self._output_path) / "libraries"))
+            ]  # @TODO: ugly
```
GlenNicholls, line 598: with f-string, you wouldn't need to type-cast as str

vunit/sim_if/xcelium.py
```diff
@@ -0,0 +1,694 @@
...
+        cds.write(self._cdslib)
+
+    def _get_mapped_libraries(self):
+        """
+        Get mapped libraries from cds.lib file
+        """
+        cds = CDSFile.parse(self._cdslib)
+        return cds
+
+    def simulate(  # pylint: disable=too-many-locals, too-many-branches
```
GlenNicholls, line 541: This whole function is really complex and confusing. I would break things up a bit or find a better way to do all the branching.

vunit/sim_if/xcelium.py
```diff
@@ -0,0 +1,694 @@
...
+
+    @staticmethod
+    def _generic_args(entity_name, generics):
+        """
+        Create xrun arguments for generics/parameters
+        """
+        args = []
+        for name, value in generics.items():
+            if _generic_needs_quoting(value):
+                args += ['''-gpg "%s.%s => \\"%s\\""''' % (entity_name, name, value)]
```
GlenNicholls, line 687: First, this seems appropriate for list comprehension. Second, why are there tripple `'''`?? This file has been using `"` for most quotes so this was awkward at first to read.

vunit/source_file.py
```diff
@@ -8,11 +8,10 @@
 Functionality to represent and operate on VHDL and Verilog source files
 """
 from pathlib import Path
-from typing import Union
+from typing import Union, Dict
 import logging
 from copy import copy
 import traceback
-from vunit.sim_if.factory import SIMULATOR_FACTORY
```
GlenNicholls, line 15: Why was this imported to the funcs/methods? It has the potential to be imported many, many times now which seems pointless to me.

tests/unit/test_xcelium_interface.py
```diff
@@ -0,0 +1,1150 @@
...
+            read_file(args_file).splitlines(),
+            [
+                "-compile",
+                "-nocopyright",
+                "-licqueue",
+                "-nowarn DLCPTH",
+                "-nowarn DLCVAR",
+                "-v200x -extv200x -inc_v200x_pkg",
+                "-work work",
+                '-cdslib "%s"' % str(Path(self.output_path) / "cds.lib"),
```
GlenNicholls, line 61: I strongly dislike mixing quote types for no reason. Use f-string or format to clean this up.

tests/unit/test_xcelium_interface.py
```diff
@@ -0,0 +1,1150 @@
...
+from vunit.test.bench import Configuration
+from vunit.vhdl_standard import VHDL
+
+
+class TestXceliumInterface(unittest.TestCase):
+    """
+    Test the Xcelium interface
+    """
+
+    @mock.patch("vunit.sim_if.xcelium.XceliumInterface.find_cds_root_virtuoso")
```
GlenNicholls, line 31: Similar mocks are used for most test methods. I would apply all the common ones to the class. For any duplicated code, you can set up a static or class method to configure the default mocks or leave customized mocks local to the test methods if it's not shared.

vunit/sim_if/xcelium.py
cmarqu: I think this code was originally mine (back when f-strings were too new), and the triple single-quotes make this more readable IMO - it's a string with quotes that contains escaped quotes.
I agree with the use of a list comprehension and would also use f-strings now

tests/unit/test_xcelium_interface.py
cmarqu: I'm not sure, but it could be that the simulator really wants to see doublequotes here.

vunit/sim_if/xcelium.py
cmarqu: This is actually as per https://peps.python.org/pep-0008/#programming-recommendations, but it could be that an unset `cdslib` should have a default of `""` which would then be tested as you show.

vunit/sim_if/xcelium.py
cmarqu: Is Jinja already a VUnit dependency? If so, I agree, otherwise it's not quite so clear if that code warrants an extra dependency.
I didn't know about `inspect.cleandoc`, interesting. I often see `textwrap.dedent` used (assuming the tabs --> spaces transformation is not needed).

tools/xcelium_verilog_fixup.py
cmarqu: Probably old code from back when `pathlib` was not widely used.

vunit/sim_if/incisive.py
cmarqu: IMO we could even remove all traces of Incisive, its last release was in 2015, everything after that was called Xcelium.

tools/xcelium_verilog_fixup.py
```diff
@@ -0,0 +1,60 @@
...
+    text = re.sub(r"(\$error\(.*\))", "\\1; $finish(1)", text)
+
+    with Path(file_name).open("w", encoding="iso-8859-1") as fptr:
+        fptr.write(text)
+
+
+def main():
+    """
+    Remove xcelium incompatabilities from source code
+    """
```
cmarqu, line 48:
```suggestion
    """Remove Xcelium incompatibilities from source code."""
```

vunit/sim_if/xcelium.py
```diff
@@ -0,0 +1,694 @@
...
+            default=None,
+            help="The cds.lib file to use. If not given, VUnit maintains its own cds.lib file.",
+        )
+        group.add_argument(
+            "--hdlvar",
+            default=None,
+            help="The hdl.var file to use. If not given, VUnit does not use a hdl.var file.",
+        )
+        group.add_argument(
+            "--use_indago",
```
cmarqu, line 77:
```suggestion
            "--use-indago",
```
Python tools generally don't use underscores in options.

vunit/sim_if/xcelium.py
```diff
@@ -0,0 +1,694 @@
...
+            cdslib=args.cdslib,
+            hdlvar=args.hdlvar,
+            use_indago=args.use_indago
+        )
+
+    @classmethod
+    def find_prefix_from_path(cls):
+        """
+        Find xcelium simulator from PATH environment variable
+        """
```
cmarqu, line 102:
```suggestion
        """Find Xcelium simulator from ``PATH`` environment variable."""
```

vunit/sim_if/xcelium.py
```diff
@@ -0,0 +1,694 @@
...
+            default=False,
+            help="Indicate whether to use Indago instead of Simvision when running in GUI mode.",
+            action="store_true",
+        )
+
+    @classmethod
+    def from_args(cls, args, output_path, **kwargs):
+        """
+        Create new instance from command line arguments object
+        """
```
cmarqu, line 87:
```suggestion
        """Create new instance from command line arguments object."""
```

vunit/sim_if/xcelium.py
```diff
@@ -0,0 +1,694 @@
...
+        Find xcelium simulator from PATH environment variable
+        """
+        return cls.find_toolchain(["xrun"])
+
+
+    @staticmethod
+    def supports_vhdl_contexts():
+        """
+        Returns True when this simulator supports VHDL 2008 contexts
+        """
```
cmarqu, line 110:
```suggestion
        """Return ``True`` when this simulator supports VHDL-2008 contexts."""
```

vunit/sim_if/xcelium.py
```diff
@@ -0,0 +1,694 @@
...
+        self._log_level = log_level
+        if cdslib is None:
+            self._cdslib = str((Path(output_path) / "cds.lib").resolve())
+        else:
+            self._cdslib = str(Path(cdslib).resolve())
+        self._hdlvar = hdlvar
+        self._use_indago = use_indago
+        self._cds_root_xrun = self.find_cds_root_xrun()
+
+
```
cmarqu, line 129:
```suggestion

```

vunit/sim_if/xcelium.py
```diff
@@ -0,0 +1,694 @@
...
+    def find_cds_root_xrun(self):
+        """
+        Finds xrun cds root
+        """
```
cmarqu, line 133:
```suggestion
        """Find xrun cds root."""
```

vunit/sim_if/xcelium.py
```diff
@@ -0,0 +1,694 @@
...
+        return subprocess.check_output(
+            [str(Path(self._prefix) / "cds_root"), "xrun"]
+        ).splitlines()[0].decode()
+
+    def find_cds_root_virtuoso(self):
+        """
+        Finds virtuoso cds root
+        """
```
cmarqu, line 141:
```suggestion
        """Find virtuoso cds root."""
```

vunit/sim_if/xcelium.py
```diff
@@ -0,0 +1,694 @@
...
+                [str(Path(self._prefix) / "cds_root"), "virtuoso"],
+                stderr=subprocess.STDOUT
+            ).splitlines()[0].decode()
+        except subprocess.CalledProcessError:
+            return None
+
+    def _create_cdslib(self):
+        """
+        Create the cds.lib file in the output directory if it does not exist
+        """
```
cmarqu, line 153:
```suggestion
        """Create the ``cds.lib`` file in the output directory if it does not exist."""
```

vunit/sim_if/xcelium.py
cmarqu: If nothing else, this should use f-strings.
I think I was using Python 3.2 when I originally wrote this which apparently didn't even support `{}` with auto-numbering.

vunit/sim_if/xcelium.py
```diff
@@ -0,0 +1,694 @@
...
+        contents += "# project libraries \n"
+        write_file(self._cdslib, contents)
+
+    def set_global_compile_options(self, global_options: Dict):
+        self._global_compile_options = global_options
+
+    def get_global_compile_option(self, name):
+        """
+        Return a copy of the compile option list
+        """
```
cmarqu, line 199:
```suggestion
        """Return a copy of the compile option list."""
```

vunit/sim_if/xcelium.py
```diff
@@ -0,0 +1,694 @@
...
+        if name not in self._global_compile_options:
+            self._global_compile_options[name] = []
+
+        return copy(self._global_compile_options[name])
+
+    def setup_library_mapping(self, project: Project):
+        """
+        Compile project using vhdl_standard
```
cmarqu, line 207: This docstring seems incorrect.

vunit/sim_if/xcelium.py
```diff
@@ -0,0 +1,694 @@
...
+        mapped_libraries = self._get_mapped_libraries()
+
+        for library in project.get_libraries():
+            self._libraries.append(library)
+            self.create_library(library.name, library.directory, mapped_libraries)
+
+    def compile_source_file_command(self, source_file):
+        """
+        Returns the command to compile a single source file
+        """
```
cmarqu, line 219:
```suggestion
        """Return the command to compile a single source file."""
```

vunit/sim_if/xcelium.py
```diff
@@ -0,0 +1,694 @@
...
+        if source_file.is_any_verilog:
+            return self.compile_verilog_file_command(source_file)
+
+        LOGGER.error("Unknown file type: %s", source_file.file_type)
+        raise CompileError
+
+    def _compile_all_source_files(self, source_files_by_library: Dict[Library, List[SourceFile]], printer):
+        """
+        Compiles all source files and prints status information
+        """
```
cmarqu, line 232:
```suggestion
        """Compile all source files and print status information."""
```

vunit/sim_if/xcelium.py
```diff
@@ -0,0 +1,694 @@
...
+            printer.write("One or more source files failed to compile.\n")
+            printer.write("Compile failed\n", fg="ri")
+            raise CompileError
+
+
+    @staticmethod
+    def _vhdl_std_opt(vhdl_standard):
+        """
+        Convert standard to format of xrun command line flag
+        """
```
cmarqu, line 311:
```suggestion
        """Convert VHDL standard to format of xrun command line flag."""
```

vunit/sim_if/xcelium.py
```diff
@@ -0,0 +1,694 @@
...
+
+        if vhdl_standard == VHDL.STD_1993:
+            return "-v93"
+
+        raise ValueError("Invalid VHDL standard %s" % vhdl_standard)
+
+    def _compile_all_files_in_library_subcommand(self, library, source_files):
+        """
+        Return a command to compile all source files in a library
+        """
```
cmarqu, line 326:
```suggestion
        """Return a command to compile all source files in a library."""
```

vunit/sim_if/xcelium.py
```diff
@@ -0,0 +1,694 @@
...
+            / ("xrun_compile_library_%s.args" % library.name)
+        )
+        write_file(argsfile, "\n".join(args))
+        return ["-F", argsfile]
+
+
+    def compile_all_files_command(self, source_files: Dict[Library, List[SourceFile]]):
+        """
+        Return a command to compile all source files
+        """
```
cmarqu, line 369:
```suggestion
        """Return a command to compile all source files."""
```

vunit/sim_if/xcelium.py
```diff
@@ -0,0 +1,694 @@
...
+            Path(self._output_path)
+            / ("xrun_compile_all.args")
+        )
+        write_file(argsfile, "\n".join(args))
+        return [cmd, "-F", argsfile, "-clean"]
+
+    def compile_vhdl_file_command(self, source_file):
+        """
+        Returns command to compile a VHDL file
+        """
```
cmarqu, line 429:
```suggestion
        """Return command to compile a VHDL file."""
```

vunit/sim_if/xcelium.py
```diff
@@ -0,0 +1,694 @@
...
+            Path(self._output_path)
+            / ("xrun_compile_vhdl_file_%s.args" % source_file.library.name)
+        )
+        write_file(argsfile, "\n".join(args))
+        return [cmd, "-F", argsfile]
+
+    def compile_verilog_file_command(self, source_file):
+        """
+        Returns commands to compile a Verilog file
+        """
```
cmarqu, line 467:
```suggestion
        """Return commands to compile a Verilog file."""
```

vunit/sim_if/xcelium.py
```diff
@@ -0,0 +1,694 @@
...
+            Path(self._output_path)
+            / ("xrun_compile_verilog_file_%s.args" % source_file.library.name)
+        )
+        write_file(argsfile, "\n".join(args))
+        return [cmd, "-F", argsfile]
+
+    def create_library(self, library_name, library_path, mapped_libraries=None):
+        """
+        Create and map a library_name to library_path
+        """
```
cmarqu, line 516:
```suggestion
        """Create *library_name* and map it to *library_path*."""
```

vunit/sim_if/xcelium.py
```diff
@@ -0,0 +1,694 @@
...
+        cds = CDSFile.parse(self._cdslib)
+        return cds
+
+    def simulate(  # pylint: disable=too-many-locals, too-many-branches
+        self, output_path, test_suite_name, config:Configuration, elaborate_only=False
+    ):
+        """
+        Elaborates and Simulates with entity as top level using generics
+        """
```
cmarqu, line 546:
```suggestion
        """Elaborate and simulate with entity as top level using generics."""
```

vunit/sim_if/xcelium.py
```diff
@@ -0,0 +1,694 @@
...
+
+        script_path = str(Path(output_path) / self.name)
+        launch_gui = self._gui is not False and not elaborate_only
+        use_indago = self._use_indago is not False
+
+
```
cmarqu, line 552:
```suggestion

```

vunit/sim_if/xcelium.py
```diff
@@ -0,0 +1,694 @@
...
+        rerun_content+=["set -euo pipefail"]
+        rerun_content+=["echo 'running compile all'"]
+        rerun_content+=["xrun -F %s/xrun_compile_all.args" % Path(self._output_path)]
+        rerun_content+=["echo 'running elaborate'"]
+        rerun_content+=["xrun -F %s/xrun_elaborate.args" % script_path]
+        write_file(rerun_file_path, "\n".join(rerun_content))
+        st = os.stat(rerun_file_path)
+        os.chmod(rerun_file_path, st.st_mode | stat.S_IEXEC)
+
+
```
cmarqu, line 565:
```suggestion

```

vunit/sim_if/xcelium.py
```diff
@@ -0,0 +1,694 @@
...
+                env=self.get_env(),
+            ):
+                return False
+
+        return True
+
+    def _hdlvar_args(self):
+        """
+        Return hdlvar argument if available
+        """
```
cmarqu, line 674:
```suggestion
        """Return hdlvar argument if available."""
```

vunit/sim_if/xcelium.py
```diff
@@ -0,0 +1,694 @@
...
+        if self._hdlvar is None:
+            return []
+        return ['-hdlvar "%s"' % self._hdlvar]
+
+    @staticmethod
+    def _generic_args(entity_name, generics):
+        """
+        Create xrun arguments for generics/parameters
+        """
```
cmarqu, line 683:
```suggestion
        """Create xrun arguments for generics/parameters."""
```

vunit/ui/__init__.py
cmarqu: Also, it could be written more succinctly using `with suppress(Exception):` (and `from contextlib import suppress`)

tools/xcelium_verilog_fixup.py
Hardolaf: The style used throughout the rest of VUnit is what we used.

vunit/sim_if/xcelium.py
GlenNicholls: Well yeah, but only if you need to know that the value is actually `None`. However, all that is being checked here is if the value is set, otherwise it's being set to the default.

In this case, it is written as an anti-pattern. My comment here is implicitly recommends avoiding this pattern by using the more logical pattern of handling `None` and other "empty" objects:

```
if not value:
  do this
else:
  do that
```

Why would you check for "false" first if you can just do the following which is less confusing to read (more important for complicated conditionals):

```
if value:
  do that
else:
  do this
```

`None` is falsy meaning you can safely use `if [str, None]` and it will never be true for `None`. The other advantage of not checking `None` here is that it will also catch empty objects like `""`, `[]`, `()`, `{}`, etc. and fallback to the default behavior. As it is written, those objects will require the user to read through a code traceback when it inevitably fails because the user used the wrong empty type. I think it's important here because there's no type hint, so I'm not actually sure what `cdslib` is supposed to be. 

Regarding the PEPs warning:

>Also, beware of writing if x when you really mean if x is not None – e.g. when testing whether a variable or argument >that defaults to None was set to some other value. The other value might have a type (such as a container) that could >be false in a boolean context!

It doesn't recommend one way or another. However, I use type-hints for this reason so the user knows what is expected and what isn't. In that case, it is safe to make assumptions about how the data can be handled as the interfaces requirements are explicit.

vunit/sim_if/xcelium.py
GlenNicholls: I'm not sure, it's been a while since I've forked and modified VUnit. `textwrap` probably makes more sense, I know `cleandoc` because I've needed some other stuff it can do that `dedent` can't.

vunit/ui/__init__.py
GlenNicholls: Do you know why this is catching all errors though? My concern is things like `KeyError` somewhere below that should be raised and might cause confusing errors later. My opinion is that this should only catch what it cares about to avoid masking VUnit bugs that affect the user.

tools/xcelium_verilog_fixup.py
GlenNicholls: For new code, I would vote for consistency with the latest repo preferences. I have no authority over this repo, but just wanted to raise the question for whoever ends up reviewing the code.

vunit/sim_if/xcelium.py
GlenNicholls: That's fair. I would honestly run this file through [black](https://black.readthedocs.io/en/stable/) with preview features enabled so it reformats strings and see what it comes up with.

tests/unit/test_xcelium_interface.py
GlenNicholls: I mentioned black earlier (I won't reference it more after this), but their docs/impl might have their str opinions about this case. Not sure if the PEPs have a recommendation for this. It doesn't look like VUnit enabled preview features so that's why it's probably not touching strings.

vunit/sim_if/xcelium.py
GlenNicholls: I don't think VUnit does either.

vunit/sim_if/xcelium.py
GlenNicholls: If flake8 is used with pydocstyle and D is enabled, it should raise an error about this.

GlenNicholls:
Another quick thing I noticed is there is already https://github.com/VUnit/vunit/pull/728. Does anyone know how this differs?

tools/xcelium_verilog_fixup.py
LarsAsplund: Yes, pathlib is preferred

vunit/sim_if/xcelium.py
Hardolaf: The style will be handled by Black which I believe formats how we had it.

vunit/sim_if/xcelium.py
Hardolaf: The style will be handled by Black.

vunit/sim_if/xcelium.py
Hardolaf: The style will be handled by Black.

vunit/sim_if/xcelium.py
Hardolaf: `cdslib` defaults to `None`, this is the correct check. In terms of type-hinting, we didn't feel it was our place to universally add type-hinting throughout all of VUnit as much as it pained us to not do so.

vunit/sim_if/xcelium.py
Hardolaf: The style will be handled by Black.

vunit/sim_if/xcelium.py
Hardolaf: The style will be handled by Black.

vunit/sim_if/xcelium.py
Hardolaf: The style will be handled by Black.

vunit/sim_if/xcelium.py
Hardolaf: The style will be handled by Black.

vunit/sim_if/xcelium.py
Hardolaf: Ideally, I'd just convert this to format strings. But I don't see a need to change it unless people are very unhappy with it.

vunit/sim_if/xcelium.py
Hardolaf: The style will be handled by Black.

vunit/sim_if/xcelium.py
Hardolaf: The style will be handled by Black.

vunit/sim_if/xcelium.py
Hardolaf: If I had to guess, the answer would be no.

vunit/sim_if/xcelium.py
Hardolaf: The style will be handled by Black.

vunit/sim_if/xcelium.py
Hardolaf: The style will be handled by Black.

vunit/sim_if/xcelium.py
Hardolaf: We left this commented code in because we didn't have time to setup a test for it but we'd ideally like to support per-source file arguments as is supported by Xcelium.

vunit/sim_if/xcelium.py
Hardolaf: The style will be handled by Black.

vunit/sim_if/xcelium.py
Hardolaf: The style will be handled by Black.

vunit/sim_if/xcelium.py
Hardolaf: The style will be handled by Black.

vunit/sim_if/xcelium.py
Hardolaf: The style will be handled by Black.

vunit/sim_if/xcelium.py
Hardolaf: The style will be handled by Black.

vunit/sim_if/xcelium.py
Hardolaf: The style will be handled by Black.

vunit/sim_if/xcelium.py
Hardolaf: The style will be handled by Black.

vunit/ui/__init__.py
Hardolaf: I believe this was added because we didn't care if it failed which is probably not the behavior we'd actually want in production quality code. I'll have to dig in and see what possible failures actually exist and see if we even care to catch any of them.

tools/xcelium_verilog_fixup.py
```diff
@@ -0,0 +1,60 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
Hardolaf, line 1: This file should be removed as it is no longer necessary.

tools/xcelium_verilog_fixup.py
Hardolaf: Removing file.

tools/xcelium_verilog_fixup.py
Hardolaf: Removing file.

Hardolaf:
@GlenNicholls As mentioned in the opening of this PR, this is based on #728 which was incomplete. If you look at the commit history for this PR, it has the full history from #728 included.

vunit/sim_if/xcelium.py
Hardolaf: I should also clarify that `""` is a valid input for this field. If the user does not specify, or does not wish to specify the `cds.lib` file, then Xcelium will infer the location of the file and handle it automatically.

rodrigomelo9:
Hi. I started #728, but I'm not a VUnit user, so when it started to be complicated, I left it unfinished...
Happy to see that somebody continues this work and hopefully, it will be part of VUnit finally! Thanks.

vunit/sim_if/xcelium.py
GlenNicholls: >In terms of type-hinting, we didn't feel it was our place to universally add type-hinting throughout all of VUnit as much as it pained us to not do so.

I totally understand that. 

`""`, at least to me, is not obvious that it's supposed to be CWD unless I knew the implementation. Yes, `Path("").resolve()` resolves to the CWD, but this to me isn't clear. I would think that if a user wanted CWD, they would pass `"."`. If I were reading the code and `""` was used, I would read it as the user letting the code choose the default. My preference is to be explicit, not implicit and in this case I think treating `""` as a valid path is implicit.

vunit/sim_if/xcelium.py
GlenNicholls: I see. In that case, I'd still remove it, but open a new issue with some of this code and where it should go unless you plan to add support.

GlenNicholls:
> @GlenNicholls As mentioned in the opening of this PR, this is based on #728 which was incomplete. If you look at the commit history for this PR, it has the full history from #728 included.

I see, thanks for the clarification!

Hardolaf:
I did a first pass on cleaning up the review per the comments. I've addressed most issues so far locally, but I haven't had time to finish my changes and verify them against xcelium this week. I'll probably get a new commit up sometime next week, time permitted.

vunit/sim_if/incisive.py
LarsAsplund: Would the Incisive from 2015 work with the new Xcelium interface?

vunit/ui/__init__.py
```diff
@@ -489,6 +492,26 @@ def add_compile_option(self, name: str, value: str, allow_empty: Optional[bool]
         for source_file in check_not_empty(source_files, allow_empty, "No source files found"):
             source_file.add_compile_option(name, value)
 
+
+    def set_global_compile_options(self, name: str, value: str, allow_empty: Optional[bool] = False):
```
LarsAsplund, line 496: Is this a way to add compile options to future added source files? Is that necessary for Xcelium support or a separate feature?

vunit/verilog/vunit_pkg.sv
```diff
@@ -79,7 +79,11 @@ class test_runner;
       end
 
       if (index == -1) begin
-         $error("Internal error: Cannot find 'enabled_test_cases' key");
```
LarsAsplund, line 82: Generally speaking we have controlled the behavior on errors by configuring the simulator. The reason is that we want the error to behave the same way regardless if it is caused by user code or VUnit features. Is that not possible with Xcelium?

Hardolaf: I believe we added this because it was the easiest way to ensure consistent behavior between Questa/Modelsim and Xcelium. We can look into configuring the max errors before stopping in Xcelium but I haven't done it before.

vunit/ui/__init__.py
Hardolaf: This was needed to achieve acceptable performance for Xcelium via single file compilation (We saw compilation times go from ~2m to ~11s on our NFS share). The other option was adding all compilation options into a hash map and then de-duping them, but in early testing that proved to be very error prone based on how users might decide to compose command line options in Python.

tools/xcelium_verilog_fixup.py
Hardolaf: No. It was added by @rodrigomelo9 in their original PR and was made irrelevant by our inclusion of the patches to the files in this PR.

vunit/sim_if/incisive.py
Hardolaf: I honestly don't know. Our codebase is incompatible with Incisive due to reliance on newer System Verilog features.

vunit/source_file.py
Hardolaf: We were having an issue with it for some reason. I have it on my list of things to clean up/fix.

JFMarten:
> I did a first pass on cleaning up the review per the comments. I've addressed most issues so far locally, but I haven't had time to finish my changes and verify them against xcelium this week. I'll probably get a new commit up sometime next week, time permitted.

Any updates regarding this MR? I would like to use VUnit with Xcelium in an upcoming project. If you need any assistance, let me know. Maybe i can somehow contribute.

bryankerr1986:
FYI I decided to try this branch with Xcelium 2303 using the Uart example run.py - Failed. Seemed like OSVVM wasn't even compiling correctly.

Hardolaf:
@JFMarten Due to leaving the position where I worked on this, I won't be able to continue the work on this for contractual reasons. I wasn't able to get this into a good linted state before leaving the job. The branch _does_ work for System Verilog and Verilog. For VHDL, we didn't have anything other than toy examples to try and those _appeared_ to work.

@bryankerr1986 I believe that is likely an OSVVM + Cadence Xcelium issue not an issue with this branch if I'm remembering the communications with Cadence correctly. There were a lot of OSVVM issues in Cadence and they got fixed after this branch was created so we never got the fixes in the git submodule into this branch.

For anyone who is capable of picking this up, please feel free to do so.

bryankerr1986:
@Hardolaf 

Ok, well I forked your repo and rebased to the latest VUnit. I also added an option to pass compile arguments. Now I have a pretty large vhdl sim working.

One thing I noticed is that your compiling the source files a little differently than incisive was doing this right? I noticed run.py --compile and run.py -m we're not working for me. Does this sound familiar?

PR #868, opened 2022-11-06
Replace inspect method removed in Python 3.11

oscargus:
`getargspec` was removed in 3.11 (after a long deprecation period) and can be replaced with `getfullargspec`.


LarsAsplund:
Thanks for your first contribution!

PR #871, opened 2022-11-07
Fix some Vivado issues (SystemVerilog support, space in file name, environment variable)

### LINKED ISSUE #686: run_vivado fails when tcl_file_name contains spaces in the path
Unfortunately Python can install to Program Files on Windows, causing the path passed to `run_vivado` by `create_compile_order_file` to also contain a space because `extract_compile_order.tcl` is contained within the install.  The problems caused by this probably vary, but looks like it has caused batch to split up the commands and fail entirely.

I think that surrounding `tcl_file_name` and also each of `tcl_args` independently with quotes for the string passed into check_call would solve it for both OSes?

### LINKED ISSUE #796: No SystemVerilog support for add_from_compile_order_file in vivado.py
There is no support for SystemVerilog files when using the [add_from_compile_order_file](https://github.com/VUnit/vunit/blob/d99fe6270493532a81fe2f2cf97423b7786a03b4/vunit/vivado/vivado.py#L16) function. It currently only support VHDL, Verilog and Verilog Headers. Currently it fails on an assert if a SystemVerilog file type exist in the generated `compile_order.txt` file.

Support can be added by changing:
```python
vunit/vivado/vivado.py

from:
30    with_dependency_scan = []
31    for library_name, file_name in compile_order:
32        is_verilog = file_name.endswith(".v") or file_name.endswith(".vp")

to:
30    with_dependency_scan = []
31    for library_name, file_name in compile_order:
32        is_verilog = file_name.endswith(".v") or file_name.endswith(".vp") or file_name.endswith(".sv")
```

and

```python
vunit/vivado/vivado.py

from:
91        for line in ifile.readlines():
92            library_name, file_type, file_name = line.strip().split(",", 2)
93            assert file_type in ("Verilog", "VHDL", "Verilog Header")
94            libraries.add(library_name)

to:
91        for line in ifile.readlines():
92            library_name, file_type, file_name = line.strip().split(",", 2)
93            assert file_type in ("Verilog", "VHDL", "Verilog Header",  "SystemVerilog")
94            libraries.add(library_name)
```

### LINKED ISSUE #811: Add VUNIT_VIVADO_PATH env variable
The vivado.py python script pulls the path to the vivado binary from either the user's path or from method argument, rather than from a env variable like the simulator paths. Modifying the `run_vivado` method under vivado.py to pull from an env variable would make it much easier to work with multiple local vivado installs on a machine, and unify this code with how VUnit specifies other binaries. 

The `run_vivado` method might look something like this

```
def run_vivado(tcl_file_name, tcl_args=None, cwd=None, vivado_path=None):
    """
    Run tcl script in Vivado in batch mode.
    Note: the shell=True is important in windows where Vivado is just a bat file.
    """
    vivado = os.environ.get("VUNIT_VIVADO_PATH", None) if vivado_path is None else str(Path(vivado_path).resolve() / "bin" / "vivado")
    cmd = f"{vivado} -nojournal -nolog -notrace -mode batch -source {str(Path(tcl_file_name).resolve())}"
    if tcl_args is not None:
        cmd += " -tclargs " + " ".join([str(val) for val in tcl_args])

    print(cmd)
    check_call(cmd, cwd=cwd, shell=True)
```

oscargus:
Closes #686
Closes #796 
Closes #811

Not sure where it can be tested. Didn't really find a test file for vivado.py?

The vivado module is not in the documentation? So not really clear where I should document the environment variable. (But I can improve the documentation for `run_vivado`.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -30,6 +30,7 @@ needs_sphinx = "3.0"
 
 extensions = [
     "sphinx.ext.autodoc",
+    "sphinx.ext.autosummary",
     "sphinx.ext.extlinks",
     "sphinx.ext.intersphinx",
     "sphinx.ext.todo",
@@ -113,7 +114,7 @@ htmlhelp_basename = "VUnitDoc"
 # -- InterSphinx --------------------------------------------------------------
 
 intersphinx_mapping = {
-    "python": ("https://docs.python.org/3.8/", None),
+    "python": ("https://docs.python.org/3/", None),
     "pytest": ("https://docs.pytest.org/en/latest/", None),
     "osvb": ("https://umarcor.github.io/osvb", None),
 }
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -49,6 +49,7 @@ often"* approach through automation. :ref:`Read more <about>`
    py/ui
    hdl_libraries
    examples
+   tool_integration/index
 
 .. toctree::
    :caption: Continuous Integration
--- a/vunit/vivado/__init__.py
+++ b/vunit/vivado/__init__.py
@@ -13,3 +13,5 @@ from vunit.vivado.vivado import (
     add_from_compile_order_file,
     create_compile_order_file,
 )
+
+__all__ = ["run_vivado", "add_from_compile_order_file", "create_compile_order_file"]
--- a/vunit/vivado/vivado.py
+++ b/vunit/vivado/vivado.py
@@ -9,7 +9,7 @@ Utilities for integrating with Vivado
 """
 
 from subprocess import check_call
-from os import makedirs
+from os import makedirs, environ
 from pathlib import Path
 
 
@@ -17,7 +17,13 @@ def add_from_compile_order_file(
     vunit_obj, compile_order_file, dependency_scan_defaultlib=True, fail_on_non_hdl_files=True
 ):  # pylint: disable=too-many-locals
     """
-    Add Vivado IP:s from a compile order file
+    Add Vivado IP:s from a compile order file.
+
+    :param project_file: :class:`~vunit.ui.VUnit`
+    :param compile_order_file: Compile-order file (from :func:`create_compile_order_file`)
+    :param dependency_scan_defaultlib: Whether to do VUnit scanning of ``xil_defaultlib``.
+    :param fail_on_non_hdl_files: Whether to fail on non-HDL files.
+
     """
     compile_order, libraries, include_dirs = _read_compile_order(compile_order_file, fail_on_non_hdl_files)
 
@@ -30,8 +36,9 @@ def add_from_compile_order_file(
 
     no_dependency_scan = []
     with_dependency_scan = []
+    verilog_file_endings = (".v", ".vp", ".sv")
     for library_name, file_name in compile_order:
-        is_verilog = file_name.endswith(".v") or file_name.endswith(".vp")
+        is_verilog = file_name.endswith(verilog_file_endings)
 
         # Optionally use VUnit dependency scanning for everything in xil_defaultlib, which
         # typically contains unencrypted top levels that instantiate encrypted implementations.
@@ -63,7 +70,13 @@ def add_from_compile_order_file(
 
 def create_compile_order_file(project_file, compile_order_file, vivado_path=None):
     """
-    Create compile file from Vivado project
+    Create compile file from Vivado project.
+
+    :param project_file: Project filename.
+    :param compile_order_file: Filename for to write compile-order file.
+    :param vivado_path: Path to Vivado install directory. If ``None``, the
+        environment variable ``VUNIT_VIVADO_PATH`` is used if set. Otherwise,
+        rely on that ``vivado`` is in the path.
     """
     print(f"Generating Vivado project compile order into {str(Path(compile_order_file).resolve())} ...")
 
@@ -81,18 +94,19 @@ def create_compile_order_file(project_file, compile_order_file, vivado_path=None
 
 def _read_compile_order(file_name, fail_on_non_hdl_files):
     """
-    Read the compile order file and filter out duplicate files
+    Read the compile order file and filter out duplicate files.
     """
     compile_order = []
     unique = set()
     include_dirs = set()
     libraries = set()
 
+    valid_file_types = ("Verilog", "VHDL", "Verilog Header", "SystemVerilog")
     with Path(file_name).open("r", encoding="utf-8") as ifile:
         for line in ifile.readlines():
             library_name, file_type, file_name = line.strip().split(",", 2)
 
-            if file_type not in ("Verilog", "VHDL", "Verilog Header"):
+            if file_type not in valid_file_types:
                 if fail_on_non_hdl_files:
                     raise RuntimeError(f"Unsupported compile order file: {file_name}")
                 print(f"Compile order file ignored: {file_name}")
@@ -119,12 +133,22 @@ def run_vivado(tcl_file_name, tcl_args=None, cwd=None, vivado_path=None):
     """
     Run tcl script in Vivado in batch mode.
 
-    Note: the shell=True is important in windows where Vivado is just a bat file.
+    :param tcl_file_name: Path to tcl file
+    :param tcl_args: tcl arguments passed to Vivado via the ``-tclargs`` switch
+    :param cwd: Passed as ``cwd`` to :func:`subprocess.check_call`
+    :param vivado_path: Path to Vivado install directory. If ``None``, the
+        environment variable ``VUNIT_VIVADO_PATH`` is used if set. Otherwise,
+        rely on that ``vivado`` is in the path.
     """
-    vivado = "vivado" if vivado_path is None else str(Path(vivado_path).resolve() / "bin" / "vivado")
-    cmd = f"{vivado} -nojournal -nolog -notrace -mode batch -source {str(Path(tcl_file_name).resolve())}"
+    vivado = (
+        str(Path(vivado_path).resolve() / "bin" / "vivado")
+        if vivado_path is not None
+        else environ.get("VUNIT_VIVADO_PATH", "vivado")
+    )
+    cmd = f'"{vivado}" -nojournal -nolog -notrace -mode batch -source "{Path(tcl_file_name).resolve()}"'
     if tcl_args is not None:
-        cmd += " -tclargs " + " ".join([str(val) for val in tcl_args])
+        cmd += " -tclargs " + " ".join([f'"{val}"' for val in tcl_args])
 
     print(cmd)
+    # shell=True is important in Windows where Vivado is just a bat file.
     check_call(cmd, cwd=cwd, shell=True)
--- a/docs/tool_integration/index.rst
+++ b/docs/tool_integration/index.rst
@@ -1,0 +1,11 @@
+.. _tool_integration:
+
+Tool Integration
+================
+There are additional integration support available for selected tools.
+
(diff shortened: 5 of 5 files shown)


LarsAsplund:
Yes, please do something about the documentation. `vivado.py` started out more as an example as we generally do not maintain tool integrations other than what's related to simulators and we generally do not document the examples that much. Since this is becoming more than just an example I suggest that it also gets some documentation.

oscargus:
Now there is some sort of documentation for the functions in the module. Located at the bottom of the "Documentation" section (below "Examples").

oscargus:
Sorry to bother you @LarsAsplund , but can you please see if anything else is required (or if you prefer the Vivado menu entry elsewhere).

docs/index.rst
```diff
@@ -47,6 +47,7 @@ often"* approach through automation. :ref:`Read more <about>`
    py/ui
    vhdl_libraries
    examples
+   py/vivado
```
LarsAsplund, line 50: This doesn't feel like a top-level thing. I would prefer we have "Tool Integrations" at this level and the "Vivado Integration" one level down. While we don't strive to maintain a lot of tool integrations we can use this section to also describe things that are special to the various simulators we support.

vunit/vivado/vivado.py
```diff
@@ -30,8 +36,9 @@ def add_from_compile_order_file(
 
     no_dependency_scan = []
     with_dependency_scan = []
+    verilog_file_endings = (".v", ".vp", ".sv")
     for library_name, file_name in compile_order:
-        is_verilog = file_name.endswith(".v") or file_name.endswith(".vp")
+        is_verilog = file_name.endswith(verilog_file_endings)
```
LarsAsplund, line 41: Nice, didn't know it takes tuples.

oscargus:
Sorry for letting this slip. Will try to find some time "soon" to complete it.

oscargus:
Finally got around to this. Not sure this is exactly what you wanted, but at least a "Tool Integration" menu item and then Vivado shows up there. Not clear to me how the "dynamic menus" work and if this will be an entry there (or if that it a better solution).

SzymonHitachi:
Hello, bump on this one @oscargus @LarsAsplund . Any chance this could get pulled into mainline? This is very useful for using with Verification IPs

docs/conf.py
```diff
@@ -113,7 +114,7 @@
 # -- InterSphinx --------------------------------------------------------------
 
 intersphinx_mapping = {
-    "python": ("https://docs.python.org/3.8/", None),
+    "python": ("https://docs.python.org/3/", None),
```
umarcor, line 117: This belongs to a separated PR or, at least, to a separated commit within this PR.

docs/conf.py
```diff
@@ -30,6 +30,7 @@
 
 extensions = [
     "sphinx.ext.autodoc",
+    "sphinx.ext.autosummary",
```
umarcor, line 33: I don't see where is the `.. autosummary::` used in the docs. Is this extension used anywhere?

vunit/vivado/__init__.py
```diff
@@ -13,3 +13,5 @@
     add_from_compile_order_file,
     create_compile_order_file,
 )
+
+__all__ = ["run_vivado", "add_from_compile_order_file", "create_compile_order_file"]
```
umarcor, line 17: As far as I am aware, the purpose of `__all__` seems to be to reduce the number of imported elements when `from <module> import *` is used. However, doing so is generally considered a not good programming practice. In fact, we recently merged a PR to reduce the "catch them all" imports in VHDL (#992). Therefore, I'm not sure we want to support/enforce those kind of imports in Python.

vunit/vivado/vivado.py
umarcor, line 41: Rather than using `.endswith`, I would suggest using Pathlib's `.suffix` (https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.suffix) because semantically we are cheking the extension of a filename not the end of an arbitrary string. At the same time, since the tuple/list is used once only, the declaraction of the variable feels redundant. I would suggest `Path(file_name).suffix in ['.v', '.vp', '.sv']`.

vunit/vivado/vivado.py
```diff
@@ -81,18 +94,19 @@ def create_compile_order_file(project_file, compile_order_file, vivado_path=None
...
     include_dirs = set()
     libraries = set()
 
+    valid_file_types = ("Verilog", "VHDL", "Verilog Header", "SystemVerilog")
     with Path(file_name).open("r", encoding="utf-8") as ifile:
         for line in ifile.readlines():
             library_name, file_type, file_name = line.strip().split(",", 2)
 
-            if file_type not in ("Verilog", "VHDL", "Verilog Header"):
+            if file_type not in valid_file_types:
```
umarcor, line 109: I would use a list instead of a tuple.

vunit/vivado/vivado.py
```diff
@@ -119,12 +133,22 @@ def run_vivado(tcl_file_name, tcl_args=None, cwd=None, vivado_path=None):
...
+        else environ.get("VUNIT_VIVADO_PATH", "vivado")
+    )
+    cmd = f'"{vivado}" -nojournal -nolog -notrace -mode batch -source "{Path(tcl_file_name).resolve()}"'
     if tcl_args is not None:
-        cmd += " -tclargs " + " ".join([str(val) for val in tcl_args])
+        cmd += " -tclargs " + " ".join([f'"{val}"' for val in tcl_args])
 
     print(cmd)
+    # shell=True is important in Windows where Vivado is just a bat file.
     check_call(cmd, cwd=cwd, shell=True)
```
umarcor, line 154: Probably not to be done in this PR, but `check_call`'s `args` parameter can be either a list of a string: https://docs.python.org/3/library/subprocess.html#subprocess.CompletedProcess.args.
I feel it would be cleaner if we managed `cmd` as a list, so we could append `tcl_args` without explicitly using `" ".join`. See https://docs.python.org/3/library/subprocess.html#converting-an-argument-sequence-to-a-string-on-windows.

umarcor (commented):
This PR seems to address multiple issues in a single commit. @oscargus can you please split it into three commits so it's easier to review? It's ok to keep them in a single PR tho.

vunit/vivado/__init__.py
oscargus: If I remember correctly: this will also make the documentation clearer, but I'll play around with it and see if it is actually the case.

docs/conf.py
oscargus: You are correct. I probably thought it was required for `automodule` at the time.

vunit/vivado/vivado.py
```diff
@@ -119,12 +133,22 @@ def run_vivado(tcl_file_name, tcl_args=None, cwd=None, vivado_path=None):
...
+        environment variable ``VUNIT_VIVADO_PATH`` is used if set. Otherwise,
+        rely on that ``vivado`` is in the path.
     """
-    vivado = "vivado" if vivado_path is None else str(Path(vivado_path).resolve() / "bin" / "vivado")
-    cmd = f"{vivado} -nojournal -nolog -notrace -mode batch -source {str(Path(tcl_file_name).resolve())}"
+    vivado = (
+        str(Path(vivado_path).resolve() / "bin" / "vivado")
+        if vivado_path is not None
+        else environ.get("VUNIT_VIVADO_PATH", "vivado")
+    )
```
oscargus, line 147: I think it is easier to comment line by line.

This line introduces the environment variable (documented above).

vunit/vivado/vivado.py
```diff
@@ -119,12 +133,22 @@ def run_vivado(tcl_file_name, tcl_args=None, cwd=None, vivado_path=None):
...
+    cmd = f'"{vivado}" -nojournal -nolog -notrace -mode batch -source "{Path(tcl_file_name).resolve()}"'
     if tcl_args is not None:
-        cmd += " -tclargs " + " ".join([str(val) for val in tcl_args])
+        cmd += " -tclargs " + " ".join([f'"{val}"' for val in tcl_args])
```
oscargus, line 150: More quoting of stuff in case of spaces.

vunit/vivado/vivado.py
oscargus: There lines fix #796

vunit/vivado/vivado.py
oscargus: Fixes #811

vunit/vivado/vivado.py
```diff
@@ -119,12 +133,22 @@ def run_vivado(tcl_file_name, tcl_args=None, cwd=None, vivado_path=None):
...
+        rely on that ``vivado`` is in the path.
     """
-    vivado = "vivado" if vivado_path is None else str(Path(vivado_path).resolve() / "bin" / "vivado")
-    cmd = f"{vivado} -nojournal -nolog -notrace -mode batch -source {str(Path(tcl_file_name).resolve())}"
+    vivado = (
+        str(Path(vivado_path).resolve() / "bin" / "vivado")
+        if vivado_path is not None
+        else environ.get("VUNIT_VIVADO_PATH", "vivado")
+    )
+    cmd = f'"{vivado}" -nojournal -nolog -notrace -mode batch -source "{Path(tcl_file_name).resolve()}"'
```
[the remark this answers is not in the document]
oscargus: This and below fixes #686

vunit/vivado/vivado.py
```diff
@@ -63,7 +70,13 @@ def add_from_compile_order_file(
...
 def create_compile_order_file(project_file, compile_order_file, vivado_path=None):
     """
-    Create compile file from Vivado project
+    Create compile file from Vivado project.
+
+    :param project_file: Project filename.
+    :param compile_order_file: Filename for to write compile-order file.
+    :param vivado_path: Path to Vivado install directory. If ``None``, the
+        environment variable ``VUNIT_VIVADO_PATH`` is used if set. Otherwise,
+        rely on that ``vivado`` is in the path.
```
oscargus, line 79: Better documentation related to environment variable.

vunit/vivado/vivado.py
```diff
@@ -9,15 +9,21 @@
...
 ):  # pylint: disable=too-many-locals
     """
-    Add Vivado IP:s from a compile order file
+    Add Vivado IP:s from a compile order file.
+
+    :param project_file: :class:`~vunit.ui.VUnit`
+    :param compile_order_file: Compile-order file (from :func:`create_compile_order_file`)
+    :param dependency_scan_defaultlib: Whether to do VUnit scanning of ``xil_defaultlib``.
+    :param fail_on_non_hdl_files: Whether to fail on non-HDL files.
+
```
oscargus, line 26: Documentation of parameters/arguments.

vunit/vivado/vivado.py
oscargus: Sure.

oscargus: Part of fixing #796

vunit/vivado/vivado.py
umarcor: Conceptually, I understand the tuples as an array of a fixed size, while a list is kind of expected to grow/reduce. Hence, I tend to use lists in other places of the codebase.

However, if there is a performance benefit, it is ok to use tuples here, and I will take it into account when touching other parts of the codebase. It's always nice to learn something!

oscargus:
Closes as this is moved to a separate repo. See https://github.com/VUnit/vunit-vivado/pull/4

Issue #881, opened 2022-12-21
Improve changelog


GlenNicholls:
I've been finding it somewhat painful to read the changelog as it contains a lot of dev-only information. I am fine with this info sitting in a misc section, but it would be nice to have breaking and features sections to make it easier to see what is actually different from a user perspective from one release to the next.

In 4.6 for example, I've highlighted a few entries that are irrelevant to most people:

```
Add Python 3.9 and 3.10 to classifiers.

-> Use MAJOR and MINOR constants to check supported Python version. [#724](https://github.com/VUnit/vunit/issues/724/)

-> Fix pylint issues.

-> Use f-strings for string formatting. [#743](https://github.com/VUnit/vunit/issues/743/) [#747](https://github.com/VUnit/vunit/issues/747/)

-> Specify encoding when using ‘open’. [#748](https://github.com/VUnit/vunit/issues/748/)

-> Set black line-length to 120 characters. [#736](https://github.com/VUnit/vunit/issues/736/)

-> Use Path from pathlib, instead of open().

Add support for log location based on VHDL-2019 call paths. [#729](https://github.com/VUnit/vunit/issues/729/)

GHDL supports VHDL package generics. [#753](https://github.com/VUnit/vunit/issues/753/)

Bump OSVVM to 2021.09.

-> [Tox] Use pytest for collecting coverage, add py310.
...
```

`Fix pylint issues` for example is something I don't find useful, that's a commit message that clutters the relevant release information in a changelog.

I've been using [towncrier](https://towncrier.readthedocs.io/en/stable/) which automates the annoyingness of a changelog. If this is of interest, I'm happy to submit a PR, but Tox and some other projects that use it are good examples as well.

LarsAsplund:
Yes, we haven't put any effort into this but from previous experiences with inhouse solutions I prefer a solution where the clarity of the release notes is based on clarity of commit messages and pull requests. To achieve that we should work more on templates. Please go ahead with a PR on this with some examples.

GlenNicholls:
One question I have is can we remove `Commits since previous release` in the release notes? I see that more as a tool for developers, but I only really see value in that when there's a regression. In that case, a user is likely just going to open an issue instead of digging through commits to find the cause.

GlenNicholls:
PR is ready to be reviewed and has examples attached.

Issue #750, opened 2021-09-29
d ?= dx or d ?= du  - Both are always FALSE.   


JimLewis:
In https://github.com/VUnit/vunit/blob/master/examples/vhdl/array_axis_vcs/src/fifo.vhd#L44
    assert always (not rst and wr -> not (d ?= dx or d ?= du))@rising_edge(clkw)
      report "wrote X|U to FIFO";

With `"?="` when one operand is an `'X'` or a `'U'` it only returns `'1'` when the other value is a `'-'`, otherwise, it returns `'X'` or '`U`'.  Probably not what you want

I think you need a regular `"="`

PR #885, opened 2023-01-06
Improved changelog to be automatically managed by towncrier

GlenNicholls:
# The Goal
This PR removes the manual changelog/release notes management by using [towncrier](https://towncrier.readthedocs.io/en/stable/). User-facing "newsfragments" live in a specific location (`docs/changelog.d`). For VUnit, it is set up such that the reST news files are of the form `<issue #>.<type>.rst` where the type can be one of (all config in `pyproject.toml` and template format in `docs/changelog.d/template.rst.j2`):

* `breaking` - breaking changes, section header in release notes will be `Breaking Changes`
* `bugfix` - A bug has been fixed, section header will be `Bug Fixes`
* `change` - A backwards compatible change has been made, section header will be `Changes`
* `deprecation` - A feature is deprecated, section header will be `Deprecations`
* `misc` - Misc. changes that may be relevant to users, section header will be `Miscellaneous`

Note that all `type`'s are supposed to be user-facing, but, the `misc` is for everything else that **might** be notable. For example, this PR for #881 might look like `881.misc.rst` that states how the changelog will be changing in future releases.

# What Changed?
This PR removes the complicated release notes process and validation. Instead, developers add newsfragments for notable, user-facing features as single files to a directory Towncrier searches. `towncrier build` will then construct the changelog based on those files, stage those files for removal, and the user commits those changes.

The other feature added is that the `py*-docs` plans will not do anything with towncrier directly. Instead `sphinxcontrib-towncrier` will add the unreleased changes to make them easier to fix syntax issues before a release is cut. 

The only process that will actually run `towncrier build` for a release is `tools/release.py`. This is because towncrier will stage newsfragment files it adds to the changelog for removal in Git. Because the release_notes.rst file exists and towncrier manages it, no validation is done. The template handles populating a default no notable changes when there is no news to add to the release notes.

Lastly, the old changelog release notes file contents have been added to `docs/release_notes.rst` and that file is now tracked by git. `tools/release.py` will also add/commit everything relevant for this process.

# Outstanding Questions
Can we remove the `Commits since previous release` and `Commits since last release`? Those IMO are not relevant in a changelog since it's developer facing for debug. I've never seen that before in release notes and it is straightforward for a developer to figure that out with git directly or in the GitHub UI. As a compromise, I changed those for future releases to `Commits since this release`, but I still don't think it's all that useful to link in an area where *most* people reading it would open an issue instead of attempting to fully debug a regression.

# Final TODO's

* [ ] add instructions to contributing section of docs (reference [pytest's changelog readme](https://github.com/pytest-dev/pytest/tree/main/changelog))
* [ ] Make sure that `scripts/release.py` works properly with these changes. I have not tested this yet as I don't know the release process as looking at GHA, it looks like releases are cut manually instead of detecting them from `vunit.__version__`.
* [ ] Provide examples for: release with no news fragments, adding multiple types of news, anything else?

# Examples

* Draft version using #881 as an example: [docsbuild.zip](https://github.com/VUnit/vunit/files/10356735/docsbuild.zip)


tools/build_docs.py
```diff
@@ -36,11 +36,36 @@ def get_theme(path: Path, url: str):
...
+            Version to set the section to for all newsfragments. Newsfragments will be
+            added to the release notes and the now old newsfragments will be staged for
+            removal.
     """
-    Build documentation/website
+    print(f"Adding newsfragment enteries to release notes for release {version}")
+    check_call(shlex.split("towncrier build --version {version} --yes"))
+
+
+def main(version=None):
```
GlenNicholls, line 52: @LarsAsplund where in the repo does it define the path to the documentation to upload to RTD? I've looked around for a while and haven't been able to figure this out. Do you manually upload the docs to RTD after a release?

In essence, I need to modify the sphinx call here to define a different location to save the release version of the docs as I am getting errors when running the release script with the sys.argv since I now have to call this main func in release.py.

GlenNicholls: It seems to me like it is expected that the tox py*-docs env is run before release.py, but I'm not seeing where a script or GHA is actually pointing to the HTML to upload to RTD.

GlenNicholls:
This is ready to be reviewed now. Looks like there are failures with the license date ranges, that should be updated on master.

LarsAsplund:
From your experience with the tool, what are the benefits of separate files? A summary of the change and the type of change should already be in the PR/commit so there will be a manual copy and paste action that seems to break single source of information. 

If we have a PR/issue template saying that you should provide the type of change and a summary of that change we would avoid that step. The information towncrier needs could be extracted from GitHub.

How should commits not related to an issue or PR be handled. Maybe not a big issue. Most of that tends to be things not very interesting to the user.

GlenNicholls:
>From your experience with the tool, what are the benefits of separate files?

IMO, I think it's just easier to put the news in separate files with the issue/PR# and type of news in the name in a volatile location that is automatically removed by towncrier when a release is cut. This way, if the PR submits a relevant change for end users you can say "please add a news/changelog entry with a quick summary of your change" that can be reviewed before the PR is merged.

>A summary of the change and the type of change should already be in the PR/commit so there will be a manual copy and paste action that seems to break single source of information.

I see your point, but generally, a PR/issue has more information than a user cares to know about. What it looks like y'all currently do is scrub the commit history since a previous release for notable changes and add them to e.g. `5.0.0.rst` as quick summaries and the release script goes and adds everything to the release notes. Effectively, you're doing the same thing; copying/pasting relevant information for the release notes and also providing a link to the original issue if someone needs to look more in-depth at the issue.

This change effectively does the exact same thing, however, it is done immediately by the author or a maintainer when it's realized something is relevant for end users and doesn't require extra time during a release to dig in further because everything has been reviewed and approved before a merge to master. The difference, though, is that the repetitive steps are automated by a popular tool for other Python projects: content for changes is copied to the release notes, issues are categorized into sections to make finding information easier, and links are added to the origin of the change if a user needs to investigate or if they're curious about the internal changes.

The new workflow is basically this:

1. dev opens an issue and submits a PR
2. maintainer sees that it affects end users and instructs them to create a news entry in `changelog.d` using either the issue or PR #, or the maintainer can add the entry to the PR before merging, whatever makes the most sense in the context
3. news is versioned in `changelog.d` until a release
4. when a release is cut, towncrier adds all entries to a new section at the beginning of the release notes with everything in `changelog.d`
5. towncrier (optionally, removing `--yes` will prompt you to remove files, `--keep` will not remove them and let you do so manually) removes now old news from changelog.d and stages it for removal. Because these are staged for removal, the commit's run in `release.py` has these committed as well

My understanding of your release process is that this change adjusts when the news is created for the release notes to be before a PR is merged such that the news can be created by a maintainer or reviewed by them instead of scrubbing the commit history and trying to remember important details or reviewing code changes/comments again to see if there's something relevant that commit messages don't mention.

My argument against your last point is that this does not break the "single source of information". Towncrier will automatically add a link to the issue/pr based on the file name and categorize it in the release notes based on the type of change. The goal of this is to change the focus of the release notes to be user-facing instead of developer-facing. In essence, the files in `changelog.d` are concise summaries of what the overall change is, not a bunch of commit messages like "fixed CI failures" or "fixed typing issues". When I install the latest version of VUnit and something breaks, the first thing I'm going to look at is the changelog. Having entries like "changed from os.path to pathlib" or "updated date range for license" clutters relevant information that might relate to my problem, e.g. "dropped python3.6 support" or some breaking change to the public API. The commit history and thorough issue details are available to dig through for developers if they need it, but IMO, users shouldn't be bothered with that information because it simply isn't relevant to them.

TLDR: check out [keep a changelog](https://keepachangelog.com/en/1.0.0/), that'll be way more concise than what I've described here.

>If we have a PR/issue template saying that you should provide the type of change and a summary of that change we would avoid that step. The information towncrier needs could be extracted from GitHub.

I don't think you would want that, though. Again, my strong opinion about changelogs is that they're for conveying relevant information to users, commit history and issue/pr titles/descriptions/comments are not focused that way. They are for developers, not users.

Like I mentioned in #881, I don't want to see any commit history in the changelog as a user if it doesn't affect me because it's distracting. If you deprecate a feature, introduce a breaking change, or add new features to the public API, I care about seeing those. But, I don't want to see anything about a CI failure that was fixed or a fix to lint issues because they don't affect me. I go to a changelog to see what actually changed from my perspective as a user and to determine if I need to open an issue if something broke, for anything else I'll go look at closed issues if I care for developer commentary. I did add a misc section for anything that *might* be relevant to users, but at least now it's categorized into sections that will hopefully make it easier to find important information.

In any case, towncrier doesn't force you to comply with opinionated rules about managing a changelog. It simply has a convention to automate some of the repetitive steps of doing so, so I figured it'd be easy enough to submit a PR to let someone else's code take the burden off maintaining it and also provide some personal insight into how I, as a user, use your release notes.

>How should commits not related to an issue or PR be handled. Maybe not a big issue. Most of that tends to be things not very interesting to the user.

Agreed, I haven't been able to think of an example where a change doesn't have an issue/PR# that is also something the user cares about. However, towncrier does support this by using an ["orphan prefix"](https://towncrier.readthedocs.io/en/stable/configuration.html#all-options) (which defaults to `+` I believe) as the first character in the `issue` part of the file name. For example, the file `docs/changelog.d/+security.misc.rst` would not add an issue link to the end of the content in that file when it's added to the release notes.

GlenNicholls:
You can also see examples from other projects that use this package for their changelog:
* [attrs](https://www.attrs.org/en/stable/changelog.html)
* [BuildBot](https://buildbot.readthedocs.io/en/v3.7.0/relnotes/index.html)
* [pip](https://pip.pypa.io/en/stable/news/)
* [pytest](https://docs.pytest.org/en/7.2.x/changelog.html)

LarsAsplund:
@GlenNicholls I completely agree that Towncrier is a step forward compared to what we have today but it may be possible to take it one step further. Today, we don't have any templates for issues and PRs and that is also something we need to improve. Quite often we have to ask the basic questions about what version is used, what simulator, what OS, how can I reproduce etc. 

A typical template for a PR starts with a short summary describing what bug is being fixed or what user value is added. There are also labels to categorize the PR. In many cases the summary + the label + the PR number = the news file if the label is "news worthy".

The tricky part which may kill that idea is if you have a PR adding X such that Y becomes obsolete. The PR summary would be "Added X and deprecated Y" but this might deserve two news files. One for the addition and one for the deprecation. Do you use it like that or can news files have several types?

Anyway, PR-generated news files would be a separate step before running Towncrier and can be added later if it turns out to be valuable. For now I think it's sufficient to keep it in mind. For example, Towncrier types should be aligned with PR labels.

We should also create a PR template that explains that a news file is needed before the PR can be merged. We also need a job that fails if a PR is created without such a file.

GlenNicholls:
@LarsAsplund 

>but it may be possible to take it one step further. Today, we don't have any templates for issues and PRs and that is also something we need to improve. Quite often we have to ask the basic questions about what version is used, what simulator, what OS, how can I reproduce etc.

I agree. I know GHDL and other projects can have certain fields populated such that a GHA can be kicked off to reproduce the problem based on a MWE. However, I think that this is a very large effort outside the scope of the issue. I'm willing to help in that area, but I don't know that it is a good thing to attempt that here.

>In many cases the summary + the label + the PR number = the news file if the label is "news worthy".

I'm still not sold that a PR description is appropriate as user news. Most PR descriptions I see might be empty or contain lower-level details. It's all valuable when you're debugging, but not for simply seeing what's changed or getting hints about where to look when something breaks. *Maybe* it would make sense to have a required section when there is a "news worthy" or "user-facing" label that would populate changelog entries. However, IMO it's the same amount of work requiring the same review; regarding what tradeoff is there, I'm not sure. With that said, it could be at least automated, but if there's a CI check for news entries when there's a specific label, then I don't think it makes sense to complicate or deal with the complexity of the GitHub hosting side of things.

>The tricky part which may kill that idea is if you have a PR adding X such that Y becomes obsolete. The PR summary would be "Added X and deprecated Y" but this might deserve two news files. One for the addition and one for the deprecation. Do you use it like that or can news files have several types?

News files cannot have several types. That feature might be in their roadmap, not sure. But, in that case, you would create two news fragments, e.g. `123.deprecation.rst` and `123.change.rst`. That way, you might put some verbiage in the `123.change.rst` that states the user should reference the deprecations/breaking changes for something that change breaks or plans to remove in the future.

>For example, Towncrier types should be aligned with PR labels.
>
>We should also create a PR template that explains that a news file is needed before the PR can be merged. We also need a job that fails if a PR is created without such a file.

100% agree. Are the current types satisfactory? In the above case, I think it makes sense that a PR would allow for multiple labels and CI would check that all labels have created a news file for that issue with those types. For the labels (assuming you are satisfied with the current types and the section verbiage), the following make sense to me:

* `news:breaking`
* `news:bugfix`
* `news:change`
* `news:deprecation`
* `news:misc`

tools/build_docs.py
GlenNicholls: This action needs to be modified: https://github.com/VUnit/vunit/blob/master/.github/workflows/docs.yml

GlenNicholls:
@LarsAsplund so I got the validation for news mostly working. For some reason, `github.event.issue.number` is returning nothing and I'm not sure why... I'm not too familiar with GHA, is there a different way to get the issue number?

GlenNicholls:
@LarsAsplund sorry for the spam, ran into a bunch of issues setting the matrix since GHA doesn't support evaluating the matrix in a conditional `if`. Anyways, you can see the [test CI run](https://github.com/VUnit/vunit/actions/runs/3898360588) I created that checks the file and passes or fails depending on which `news:*` labels are present. When no `news:*` labels are present, none of the validators will show up which is an unfortunate consequence of not being able to evaluate the matrix expression in a conditional.

To test this PR for the automated checks for the news, you can add the `news:misc`, review the `news` validator to make sure it passes, and repeat by adding `news:change` which should only fail the `news:change` validator.

tools/build_docs.py
umarcor: `tox py*-docs` generates the HTML output in https://github.com/VUnit/vunit/blob/master/.github/workflows/docs.yml#L45. The website is updated by creating an orphan branch, adding the content and pushing it to repo vunit.github.io using a private SSH key: https://github.com/VUnit/vunit/blob/master/.github/workflows/docs.yml#L48-L61.

Docs are updated after each push to master, regardless of it being a release: https://github.com/VUnit/vunit/blob/master/.github/workflows/docs.yml#L41.

tools/build_docs.py
```diff
@@ -36,26 +37,60 @@ def get_theme(path: Path, url: str):
...
         [
             sys.executable,
             "-m",
             "sphinx"
-        ] + ([] if len(argv) < 2 else argv[2:]) + [
+        ] + ([] if len(sys.argv) < 2 or version else sys.argv[2:])
+        + (["-t", "release_notes_draft"] if version is None else []) + [
             "-TEWanb",
             "html",
-            Path(__file__).parent.parent / "docs",
```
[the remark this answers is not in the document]
umarcor: Probably just needs to be wrapped in `str()`.

GlenNicholls:
>It seems that this PR/branch gathered a bunch of modifications and tests! I suggest we split it into smaller PRs, to more easily review and merge.

Yeah... because the original idea had a scope change request in the PR.

>changing the current flow to use towncrier instead of manually writing one rst file for each release

That was the goal. Providing VUnit with a way to mange a changelog that was user facing, not developer facing. With that said, I see no issue with creating a rst file per release. You add a rst file whenever there's something relevant to the user and towncrier handles deleting it when it's released... To be clear, adding new files to the release section of the repo is the whole point of towncrier. 

At the end of the day, my compliant is that a changelog reporting commit history is absolutely, 100%, useless. I can see commit history in commits in GitHub, GitLab, etc.. What I expect to see in changelogs is relevant information for a user like API changes, REST API differences, etc. Specifically, I care about user-facing changes in changelogs, not developer information.

>Let's change that: let's create a subdir in docs where we put one file for each meaningful change

This PR already does that - removes developer notes and commit messages from user facing release notes and allows a single ReST file for anything useful created in `news.d`.

>We need to release v4.7.0. Using the same approach as in the last 4y, I would go through https://github.com/VUnit/vunit/compare/v4.6.0...master and would manually create two files: the rst to be included in the docs and the same in markdown syntax to be used in the description in GitHub Releases. Let's change that: let's create a subdir in docs where we put one file for each meaningful change.

Again, I don't see the benefit. [99% of the current release notes](https://vunit.github.io/release_notes.html) are not useful... Let's take 4.6 as an exmaple an examine what's acutally useful for the average user. Examine the below and let me know what you would find useful as an end user? I found exactly ONE change that I found useful.......

As you'll see, theres 1 (ONE) change that's actually useful for me...

- ~Add Python 3.9 and 3.10 to classifiers.~

- ~Use MAJOR and MINOR constants to check supported Python version. [#724](https://github.com/VUnit/vunit/issues/724/)~

- ~Fix pylint issues.~

- ~Use f-strings for string formatting. [#743](https://github.com/VUnit/vunit/issues/743/) [#747](https://github.com/VUnit/vunit/issues/747/)~

- ~Specify encoding when using ‘open’. [#748](https://github.com/VUnit/vunit/issues/748/)~

- ~Set black line-length to 120 characters. [#736](https://github.com/VUnit/vunit/issues/736/)~

- ~Use Path from pathlib, instead of open().~

- ~Add support for log location based on VHDL-2019 call paths. [#729](https://github.com/VUnit/vunit/issues/729/)~

- ~GHDL supports VHDL package generics. [#753](https://github.com/VUnit/vunit/issues/753/)~

- Bump OSVVM to 2021.09.

- ~[Tox] Use pytest for collecting coverage, add py310.~

- ~[Tests] mark array_axis_vcs and verilog_ams examples as xfail. [#751](https://github.com/VUnit/vunit/issues/751/)~

- ~[Logging/log_deprecated_pkg] fix compilation issues with Cadence tools. [#731](https://github.com/VUnit/vunit/issues/731/)~

- ~[Parsing/tokenizer] partial revert of 5141f7c [#735](https://github.com/VUnit/vunit/issues/735/) [#745](https://github.com/VUnit/vunit/issues/745/)~

- ~[UI] make glob search recursive by default.~

- ~[VCs] bugfix AXI stream slave nonblocking check. [#720](https://github.com/VUnit/vunit/issues/720/)~

- ~[Examples] add shebang to run scripts. [#738](https://github.com/VUnit/vunit/issues/738/)~

- ~[Example/vhdl/user_guide] add VHDL 1993 variant, clean use statements, skip in acceptance tests if VHDL context not supported. [#737](https://github.com/VUnit/vunit/issues/737/)~

- ~[Examples/vhdl/array_axis_vcs] Fix PSL check for valid fifo in data during write. [#750](https://github.com/VUnit/vunit/issues/750/) [#766](https://github.com/VUnit/vunit/issues/766/)~

- ~[Docs] bump sphinx_btd_theme to v1, revert temporary pinning of Sphinx and docutils, remove redundant delete message call from com user guide example, fix ref to Travis CI (deprecated) (GitHub Actions is used now), add section about envvars, document VUNIT_VHDL_STANDARD, use ‘exec’ directive to generate content in examples, update ‘Credits and License’, add refs to Tratex. [#730](https://github.com/VUnit/vunit/issues/730/) [#739](https://github.com/VUnit/vunit/issues/739/) [#761](https://github.com/VUnit/vunit/issues/761/)~

- ~[CI] add emojis/icons, avoid deployments from forks, fix deploy condition event, add job using setup-ghdl-ci, update images from Debian Buster to Debian Bullseye, do not overload image tags.~

Why waste time or release notes to cover internal notes? This PR fixes that by providing a template to fix developer vs user facing release notes.

If you don't want it, that's fine. However, I would appreciate a thorough plan if you want any more time from me to fix this integration.

umarcor:
> Yeah... because the original idea had a scope change request in the PR.

I understand. Yet, I would like to discuss technical/implementation details about the various features. Discussing all of it here would be hard to follow. Since we can split it into smaller chunks, I believe we'd better do it step by step.

> At the end of the day, my compliant is that a changelog reporting commit history is absolutely, 100%, useless. I can see commit history in commits in GitHub, GitLab, etc.. What I expect to see in changelogs is relevant information for a user like API changes, REST API differences, etc. Specifically, I care about user-facing changes in changelogs, not developer information.

I understand and agree with the motivation. I would personally not be so minimalistic, but that's not relevant. I'm good with the selection of "relevant" changes proposed by you, Lars or anyone else. My concern is not conceptual, but pragmatic. We have ~175 items which we need to filter/select (see https://github.com/VUnit/vunit/pull/918/files) before we can execute towncrier to generate the summary/changelog/notes. Until we don't do that, we cannot know if the fields/categories/templates suffice to capture the knowledge we want to share with users. For instance, I believe that bumping dependencies might be a category itself.

> That was the goal. Providing VUnit with a way to mange a changelog that was user facing, not developer facing. With that said, I see no issue with creating a rst file per release. You add a rst file whenever there's something relevant to the user and towncrier handles deleting it when it's released... To be clear, adding new files to the release section of the repo is the whole point of towncrier.

I believe we have different points of view about what towncrier can do and what we want it to do. I agree with creating changelog/release notes that are user facing, rather than developer facing. I agree that towncrier can help by letting us write one file per change, using the filename to encode a backref to the issue/PR and the category/type. From there on, I don't think we want towncrier to delete anything, nor to manipulate/modify files/releases. Instead, we want it to assemble the individual change files and produce a single block of text that we can a) include in the sphinx documentation and b) include as the description of the Release in GitHub. That's mostly because letting towncrier do anything else involves unnecessarily increasing the complexity to start using it. We might want to improve that later, but at the moment we need a markdown file that we can copy and paste into the release description, and ideally we can include that same file in the sphinx documentation.

> This PR already does that - removes developer notes and commit messages from user facing release notes and allows a single ReST file for anything useful created in news.d.

I meant/mean "let's create a subdir in docs where we put one for each meaningful change **[since v4.6.0]**". I'm trying to reduce the technical complexity of adding towncrier and instead focus on creating the content that towncrier needs to process/use.

From this PR, I want to pick:

- Updating the `exclude_patterns` to ignore both `release_notes` and `news.d`.
- Adding 'Release Notes Intructions' to section 'Contributing' of the docs, but *not* removing the instructions to create the release notes file for a version (update them to *"copy/mv/rename the file generated by towncrier*").
- Adding `docs/news.d/template.rst.j2` (I think that the gitignore is not necessary), but adapted for markdown (see https://towncrier.readthedocs.io/en/stable/markdown.html).
- Adding the towncrier configuration to `.pyproject.toml`.
- Calling/executing towncrier when the documentation is built. After/if/when #912 is merged, we might get rid of `build_docs.py`, since we can do all of that in the `conf.py`. Hence, we can execute towncrier either in `build_docs.py` or in `conf.py`.

I would leave other modifications aside, for now:

- The CI workflow.
- `release_notes/*.rst`, `create_release_notes.py` and `release.py`.

By picking those changes, and having change files for the modifications since v4.6.0, we can start using towncrier and create the relase notes (changelog) for v4.7.0 with it.

> Again, I don't see the benefit. 99% of the current release notes are not useful... Let's take 4.6 as an exmaple an examine what's acutally useful for the average user. Examine the below and let me know what you would find useful as an end user? I found exactly ONE change that I found useful.......

> As you'll see, theres 1 (ONE) change that's actually useful for me...

> Why waste time or release notes to cover internal notes? This PR fixes that by providing a template to fix developer vs user facing release notes.

There is no benefit, and I don't want to waste a time I don't have. Yet, we need to create the change files for towncrier to generate any meaningful output. The sources of knowledge we have are the following:

- 117 commits in master since v4.6.0: https://github.com/VUnit/vunit/compare/v4.6.0...master
- 58 closed/merged issues/PRs in milestone v4.7.0: https://github.com/VUnit/vunit/milestone/2?closed=1

This is the first time we have issues/PRs milestoned; in all previous releases it needed to be checked by looking at the dates. This is to explain where we are coming from and where we are now; which we need to know if want to draw the path to where we want to be.

Please, let me know how do you suggest we create a summary of 1, 2, 5, 10 meaningul changes from those 175 items. My understanding is that one or multiple humans need to do that, because no one generated files for towncrier in the last year. So, we (you, me, whoever interested in using towncrier) need to go through the change files in https://github.com/VUnit/vunit/pull/918/files and elimite all the ones that are not meaningful, apart from rewording the ones that are to be kept. Naturally, I'm open to any suggestion to achieve the same result with less effort.

> If you don't want it, that's fine. However, I would appreciate a thorough plan if you want any more time from me to fix this integration.

I am all in with using towncrier. The thorough plan is:

- Not in a particular order:
  - Eliminate or reword the 175 change files in https://github.com/VUnit/vunit/pull/918/files. (first block of my previous comment)
  - Combine the change files from #918 (not the JSON neither the py) with the features of this PR listed above. (second block of my previous comment)
  - Change the configuration for towncrier to use markdown instead of rst.
- Merge the result to have a "Next" or "Development" section at the top of http://vunit.github.io/release_notes.html showing the content of an included markdown file generate by towncrier by reading the change files.
- Discuss whether we want to merge any other unrelated content before releasing.
- Release v4.7.0. As we would explain in the updated guidelines, a maintainer would:
  - Build the docs locally to generate the changelog file.
  - `mv changelog.md docs/release_notes/v4.7.0.md`
  - Follow the rest of the "releasing" guidelines, which will involve removing the change files at the same time that `v4.7.0.md` is added. NOTE: I believe we prefer release specific code to be in `release.py` (which is only executed by maintainers), rather than `build_docs.py` (which is executed every time the docs are built, either locally or in CI).
- Rebase this PR to further discuss the other two modifications:
  - Combining the version release notes files in a single one. (third block of my previous comment)
  - Having some CI to require change files to exist in PRs. (last block of my previous comment)

umarcor:
FTR, with towncrier's option `single_file = false` (https://towncrier.readthedocs.io/en/stable/configuration.html#all-options), we can generate a file for each release. So, `mv` for releasing is not necessary; we can just change the `--version`.

PR #887, opened 2023-01-14
Identity

LarsAsplund:


LarsAsplund:
Here is a preview of what the built user guide will look like. 
[Identity Package — VUnit documentation.pdf](https://github.com/VUnit/vunit/files/10417273/Identity.Package.VUnit.documentation.pdf)

Issue #889, opened 2023-01-17
Questa Sim 2022.4_2 changes default bit ordering in unconstrained arrays.  Test tb_uart_lib.tb_uart_rx.test_receives_one_byte broken.


nicdes:
Updating questasim from 2020.1_1 to 2022.4_2 breakes the test 'tb_uart_lib.tb_uart_rx.test_receives_one_byte':

```
#      73810000 ps - check                        -   ERROR - Got 1110_1110 (238). Expected 0111_0111 (119).
```

And I expect many others.  Note that the got value is the swapped version of the expected value.

The reason is that this procedure https://github.com/VUnit/vunit/blob/868eea10d187854a0b1adb02d2ff33df4b88293a/examples/vhdl/uart/src/test/tb_uart_rx.vhd#L54
ends up assigning the value x"77" to the unconstrained data port in the uart_master verification component:
https://github.com/VUnit/vunit/blob/868eea10d187854a0b1adb02d2ff33df4b88293a/vunit/vhdl/verification_components/src/uart_master.vhd#L29

Questa Sim has changed the bit ordering of expressions like x"77" when assigned to unconstrained std_logic_vector ports.

My fix is replacing https://github.com/VUnit/vunit/blob/868eea10d187854a0b1adb02d2ff33df4b88293a/vunit/vhdl/verification_components/src/uart_master.vhd#L43
with:
```
      for i in data'length-1 downto 0 loop
```

VUnit version: 4.6.0
Simulator: Questa Sim-64 vsim 2022.4_2 Simulator 2022.12 Dec  9 2022

LarsAsplund:
The first thing `push_string` does is to normalize the input. In this case to `7 downto 0`. When the UART master is looping `0 to data'length-1` it is sending the bits in 0x77 LSB first as it should. If you look at the master output you should see a start bit 0 and then 11101110. There are many other places where things can flip. To know I would have to try to recreate locally.

Flipping the order of the loop may fix things for your version of Questa but it will fail on all other versions and simulators

nicdes:
> There are many other places where things can flip. To know I would have to try to recreate locally.

Indeed, I just observed the flip in the uart_master VC component output.  I've run the whole vhdl test suite (vunit/examples/vhdl/docker_runall.sh) and no other errors showed up.  Not what I would expect if the flip happens elsewhere.

> Flipping the order of the loop may fix things for your version of Questa but it will fail on all other versions and simulators

[...]

PR #891, opened 2023-01-18
Patch questa 2022.4 2

nicdes:
Workaround for issue #889.  It adds a swap_bit_order argument to the uart_master_t record used to initialize the uart_master VC.

This allows to use:
```vhdl
  constant master_uart_a   : uart_master_t   := new_uart_master(swap_bit_order => true);
```
in a testbench that runs with questasim 2022.2_4.  Default (false) is the old behaviour.

Remaining issue: the example testcase tb_uart_lib.tb_uart_rx.test_receives_one_byte' would still fail if run with questasim 2022.4_2 because tb_uart_rx is unchanged.

### DIFF
--- a/vunit/vhdl/verification_components/src/uart_master.vhd
+++ b/vunit/vhdl/verification_components/src/uart_master.vhd
@@ -28,7 +28,8 @@ begin
   main : process
     procedure uart_send(data : std_logic_vector;
                         signal tx : out std_logic;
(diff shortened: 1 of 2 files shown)


LarsAsplund:
I will close this. Changing our APIs to workaround the bug in Questa is not a reasonable response. I hope @siemens will release a patch for this because:

image

If not, #892 is a better workaround.

PR #892, opened 2023-01-18
Work around for questasim 2022.4_2 bit ordering bug.

nicdes:
Fix for issue #889.

### DIFF
--- a/vunit/vhdl/verification_components/src/stream_master_pkg-body.vhd
+++ b/vunit/vhdl/verification_components/src/stream_master_pkg-body.vhd
@@ -23,7 +23,11 @@ package body stream_master_pkg is
     variable msg : msg_t := new_msg(stream_push_msg);
     constant normalized_data : std_logic_vector(data'length-1 downto 0) := data;
   begin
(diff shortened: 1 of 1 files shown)


vunit/vhdl/verification_components/src/stream_master_pkg-body.vhd
```diff
@@ -23,7 +23,11 @@ package body stream_master_pkg is
     variable msg : msg_t := new_msg(stream_push_msg);
     constant normalized_data : std_logic_vector(data'length-1 downto 0) := data;
   begin
-    push_std_ulogic_vector(msg, normalized_data);
+    if data'ascending then
+      push_std_ulogic_vector(msg, normalized_data);
+    else
+      push_std_ulogic_vector(msg, normalized_data);
+    end if;
```
oscargus, line 30: Should these really both be `normalized_data`? I may be missing something obvious though...

LarsAsplund:
There is a bug in this Questa version and a "random" rewrite of the code can have the effect of working around the bug. As explained in #889 it's not reasonable for VUnit to maintain such workarounds so I will close this.

PR #902, opened 2023-02-21
Allow recent versions of GCC to be recognized

dlschmidt:
When using VUnit with i.e. GCC 12.0.0 the compiler does not get recognized because VUnit tries to match "GCC back-end code generator" exactly. Atleast with GHDL 3.0.0 with GCC 12.0.0 will produce the following output:
```
GHDL 3.0.0-dev (tarball) [Dunoon edition]
 Compiled with GNAT Version: 12.1.0
 GCC 12.2.1 code generator```

I have updated the regular expression to fix this issue but remaining the old behaviour at the same time.


dlschmidt:
closed as this was already fixed

PR #917, opened 2023-03-08
ci: add NVC jobs

umarcor:
This PR adds NVC jobs to CI.

Yesterday, I added a container to [hdl.github.io/containers](https://hdl.github.io/containers/), which builds the `master` branch of NVC (see https://github.com/hdl/containers/commit/367faa2ca1c9a47912e31feadaedad35ae25abc1).

In this PR, the Images workflow is modified to build container image `ghcr.io/vunit/dev/nvc` based on `gcr.io/hdl-containers/nvc`. Then, in the Push workflow, a two job matrix is added to run acceptance and vcomponents tests using `ghcr.io/vunit/dev/nvc`.

By the way, two tests were not being executed by NVC because they were marked to be run with GHDL only. It seems that NVC can handle one of them (JSON in generics). Hence, I updated the xfail mark condition. The other one is related to PSL: https://github.com/VUnit/vunit/blob/master/examples/vhdl/array_axis_vcs/src/fifo.vhd#L38-L48. Maybe NVC can handle enough of PSL through some switch/option?

@LarsAsplund it seems that the image pushed by github actions was set to private by default (https://github.com/orgs/VUnit/packages). That's making CI fail when trying to pull with the default token (funny, since that was used for pushing it XD). Can you check whether you can change it to public? Maybe somewhere in https://github.com/VUnit/vunit/pkgs/container/dev%2Fnvc? BTW, I think we can remove https://github.com/VUnit/vunit/pkgs/container/vunit%2Fdev.

@nickg at first I used the same dependencies for building and runtime. However, in order to reduce the size of the container, I try reduce runtime dependencies. This is the list I came with through trial and error, since I could not find an explicit list: https://github.com/hdl/containers/blob/main/debian-bullseye/nvc/HDLC#L49-L54. As you can see in https://gcr.io/hdl-containers/nvc, I could reduce from 324 MB to 160 MB. For reference, the equivalent container with GHDL LLVM takes 148 MB (https://hdl.github.io/containers/ToolsAndImages.html). The pkg/ghdl is 10MB and pkg/nvc is 18MB, so that's where the difference is coming from. I think it is good enough, but please let me know if you would add or remove any.


nickg:
Shouldn't need clang as a runtime dependeny (`ld` from binutils is enough on Linux) and don't need libdwarf at all (we use libdw instead on Linux which you already have). But you do need the llvm runtime libs, maybe that's what clang is bringing in as a dependency?

umarcor:
@nickg thanks! https://github.com/hdl/containers/commit/fd54c4e40dda266821b25cf2e58a76f019228b1c reduced the size of the container image to 100 MB, and all VUnit tests do pass!

LarsAsplund:
@umarcor I made the package public and yes, I think you can remove that old package.

PR #918, opened 2023-03-08
Create release notes for v4.7.0 using towncrier

umarcor:
~**This is not to be merge yet!**~

This PR is related to #885.

In preparation for releasing v4.7.0, I run two queries in GitHub's GraphQL Explorer and I saved the results in two JSON files. Then, I wrote a Python script that takes those JSON files and generates a markdown file for each issue|pr or commit. I run the script and added all the files to the PR. Note that:

- "issues" are issues or PRs which are milestoned.
- "commits" are the last 100 commits in branch `master`. Hence, there are 17 commits missing (the first after tag v4.6.0).

Yet to be done:

- ~Configure towncrier to use markdown: https://towncrier.readthedocs.io/en/stable/markdown.html~
- Cleanup all the files added in this PR to only keep the ones that are meaningful for end-users.
- ~Add myst-parser to include the changelog in release_notes.~

/cc @GlenNicholls

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -250,15 +250,16 @@ Make sure to use full sentences in the **past or present tense** and use punctua
 Each file should be named like ``<issue #>.<type>.rst``, where ``<issue #>`` is the
 GitHub issue or pull request number, and ``<type>`` is one of:
 
-* ``breaking``: a change which may break existing functionality, such as feature removal
-  or behavior change
-* ``bugfix``: fixes a bug
-* ``change``: backwards compatible features or improvements
-* ``deprecation``: feature deprecation
-* ``misc``: a ticket was closed, but it is not necessarily important for the user
+* ``breaking``:may break existing functionality; such as feature removal or behavior change.
+* ``bugfix``: fixes a bug.
+* ``doc``: related to the documentation.
+* ``deprecation``: feature deprecation.
+* ``feature``: backwards compatible feature addition or improvement.
+* ``misc``: a ticket was closed, but it is not necessarily important for the user.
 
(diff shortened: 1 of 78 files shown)


umarcor:
This PR is now based on #921. I went through the ~200 issues|PRs|commits and I created ~75 news fragments.

I changed the categories/types to add `doc` and `feature`, and I removed `change`.

I found some limitations in towncrier:

- It's not possible to assign multiple PRs|issues to a single fragment. In order to get multiple refs in one item, the fragment file needs to be duplicated with different filenames. The content must be exactly the same (including line-endings). I think the duplication is acceptable until they improve it in towncrier.
- It's possible to have fragments with commit refs, instead of PRs|issues. For instance, `g758b9e47.deprecation.rst`. However, it's not possible to have them linked. This is because configuration option `issue_format` supports a simple replacement of variable `{issue}`; it does not support complex logic to select the format depending on the ID/prefix. Therefore, I decided to use `+758b9e47.deprecation.rst`, which shows those items without any ID/ref.
- Since the template or the issue_format cannot be overriden in CLI, it's not straightforward to support two outputs (one for GitHub Releases without sphinx features, and another one to be included in sphinx). Even if we used markdown only, we would still need to maintain two configuration files. Therefore, I propose to stop adding release notes to the GitHub Releases, and instead just add a link to the corresponding section in the docs. This PR adds labels before each subsection so we can use `https://vunit.github.io/release_notes.html#release-4-6-0`.

@LarsAsplund @GlenNicholls if you are good with the implementation in #921 and the news fragments introduced in this PR, we can introduce towncrier already and use it to handle the remaining PRs before releasing v4.7.0. Actually, we can merge #921 and #918 and keep changing the fragments until we decide to release v4.7.0.

PR #921, opened 2023-03-11
docs: improve changelog to be automatically generated by towncrier

umarcor:
This is a subset of #885, introducing minimal modifications to start using towncrier.

Instead of generating all of `docs/release_notes.rst`, a skeleton is checked in git. The skeleton includes two files:

- `news.inc`: generated by towncrier.
- `release_notes.inc`: generated by the existing logic in `create_release_notes.py`.

`towncrier` is executed in `create_release_notes`. Hence, in the same function all the news fragments are combined into `news.inc` and all the release notes are combined into `release_notes.inc`. This works even if no news fragments exist (right after a release) because towncrier will produce a sentence telling that no relevant news exist yet.

When a release needs to be done, a maintainer renames the `news.inc` to `release_notes/X.Y.Z.rst` and removes the news fragments before executing `release.py`.

It is to be noted that towncrier tries to add the generated file to git. It seems not possible to prevent it through the configuration nor the CLI. The only alternative seems to be `--draft`, which prints to stdout instead of writing to the file. That's why this PR uses a fixed file name for towncrier and it is added to `.gitignore`. The output of towncrier is always temporary (e.g. generated in CI) until a maintainer renames the file to be treated as release notes.

/cc @GlenNicholls

### DIFF
--- a/docs/.gitignore
+++ b/docs/.gitignore
@@ -1,4 +1,3 @@
 /_build/
 /*.inc
 /license.rst
-/release_notes.rst
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -51,7 +51,7 @@ release = ""
 
 language = "en"
 
-exclude_patterns = ["release_notes/*.*"]
+exclude_patterns = ["release_notes/*.*", "news.d/*.*"]
 
 todo_include_todos = False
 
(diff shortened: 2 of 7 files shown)


umarcor:
FTR twisted/towncrier#487

PR #927, opened 2023-04-14
Fix NVC version detection

nickg:
I bumped the version on the master branch to 1.10-devel but the comparison here is using floating point so 1.10 >= 1.9 is false. Changed determine_version to return a (major, minor) tuple instead.

### DIFF
--- a/vunit/sim_if/nvc.py
+++ b/vunit/sim_if/nvc.py
@@ -82,7 +82,9 @@ class NVCInterface(SimulatorInterface):  # pylint: disable=too-many-instance-att
         self._gtkwave_args = gtkwave_args
         self._vhdl_standard = None
         self._coverage_test_dirs = set()
-        self._supports_jit = self.determine_version(prefix) >= 1.9
+
+        (major, minor) = self.determine_version(prefix)
+        self._supports_jit = major > 1 or (major == 1 and minor >= 9)
 
         if self.use_color:
             environ["NVC_COLORS"] = "always"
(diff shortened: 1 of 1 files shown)


vunit/sim_if/nvc.py
```diff
@@ -105,12 +108,12 @@ def determine_version(cls, prefix):
...
-        return float(
-            re.match(
-                r"nvc ([0-9]*\.[0-9]*).*\(.*\)",
-                cls._get_version_output(prefix),
-            ).group(1)
-        )
+        raw = cls._get_version_output(prefix)
+        match = re.match(r"nvc ([0-9]+)\.([0-9]+).*", raw)
+        if not match:
+            raise RuntimeError("Cannot determine NVC version: " + raw)
```
umarcor, line 114: Please, use an f-stringinstead of `+`.

vunit/sim_if/nvc.py
```diff
@@ -261,8 +264,7 @@ def simulate(self, output_path, test_suite_name, config, elaborate_only):
 
         if not elaborate_only:
             cmd += ["--no-save"]
-            if self._supports_jit:
-                cmd += ["--jit"]
+            cmd += ["--jit"]
```
umarcor, line 267: This implies that VUnit can only work with NVC >= 1.9, and that's why you added the runtimeerror above?

Just willing to clarify, because this feels a second change on top of improving the version detection. Might be worth having two commits, even if we merge them at the same time.

nickg: Most of the examples worked with earlier releases but there were also many acceptance test failures so it seems better to specify a minimum version that we know is fully working. But I agree it's a logically separate change so I'll do it in a separate PR after this one is merged.

nickg:
The CI failures don't appear to be caused by this patch.

umarcor (approved):
FTR, this PR modifies `determine_version`. The same function/method exists in the GHDL class, and the returned type is not consistent. Actually, GHDL uses a `float` while this PR is changing NVC to a tuple.
However, that is not a problem because each simulator class uses a different mechanism to detect the version, and we don't currently have a consistent API to "get the version of the tool" independently of the specific tool.

PR #930, opened 2023-04-23
Fix error with NVC and the --gui flag

### LINKED ISSUE #926: Error with NVC and the --gui flag
I just tried the NVC simulator support, which worked well for me (apart from this one issue). NVC seems quite a bit faster than GHDL, so that's awesome!

But there's one error I had to fix when running with the `--gui` flag:

```
  File "/home/odin/git/vunit/vunit/test/runner.py", line 244, in _run_test_suite
    results = test_suite.run(output_path=output_path, read_output=read_output)
  File "/home/odin/git/vunit/vunit/test/list.py", line 105, in run
    test_ok = self._test_case.run(*args, **kwargs)
  File "/home/odin/git/vunit/vunit/test/suites.py", line 72, in run
    results = self._run.run(*args, **kwargs)
  File "/home/odin/git/vunit/vunit/test/suites.py", line 178, in run
    sim_ok = self._simulate(output_path)
  File "/home/odin/git/vunit/vunit/test/suites.py", line 237, in _simulate
    return self._simulator_if.simulate(
  File "/home/odin/git/vunit/vunit/sim_if/nvc.py", line 293, in simulate
    stdout.write(f'{" ".join(cmd)}\n')
TypeError: sequence item 1: expected str instance, PosixPath found
```

The solution I applied was to use `[str(wave_file)]` instead of `[wave_file]` here:

https://github.com/VUnit/vunit/blob/1bc4d060aaeadc80a1f95df1965906870b37e752/vunit/sim_if/nvc.py#L287

0dinD:
Fixes #926

https://github.com/VUnit/vunit/issues/926#issuecomment-1514649824 by  @umarcor 

> For reference, in GHDL the Path is converted to str in
https://github.com/VUnit/vunit/blob/1bc4d060aaeadc80a1f95df1965906870b37e752/vunit/sim_if/ghdl.py#L338

Yeah, the GHDL interface does the string conversion earlier:

https://github.com/VUnit/vunit/blob/fc91e48dbf9daa1029d14758e8395c2836dd81b8/vunit/sim_if/ghdl.py#L330-L340

But the path is still useful (for checking if the file exists and so on), so the code ends up converting back and forth between `Path` and `str` everywhere. Is it not better to wait with the string conversion until a string is required, like what I did in cc5c6073756394ed689256c727595e724885d974?

I could add a commit to refactor the Path handling in the GHDL (or NVC?) interface, if you want it to be more consistent.

### DIFF
--- a/docs/news.d/926.feature.rst
+++ b/docs/news.d/926.feature.rst
@@ -1,0 +1,1 @@
+Add NVC simulator support.
--- a/docs/news.d/930.feature.rst
+++ b/docs/news.d/930.feature.rst
@@ -1,0 +1,1 @@
+Add NVC simulator support.
--- a/vunit/sim_if/nvc.py
+++ b/vunit/sim_if/nvc.py
@@ -286,7 +286,7 @@ class NVCInterface(SimulatorInterface):  # pylint: disable=too-many-instance-att
             status = False
 
         if self._gui and not elaborate_only:
-            cmd = ["gtkwave"] + shlex.split(self._gtkwave_args) + [wave_file]
+            cmd = ["gtkwave"] + shlex.split(self._gtkwave_args) + [str(wave_file)]
 
             init_file = config.sim_options.get(self.name + ".gtkwave_script.gui", None)
             if init_file is not None:


umarcor (approved):

umarcor:
@0dinD, it's ok. I provided the reference to GHDL's implementation for completeness only, to support your point that this change to NVC is correct indeed.

According to your comment/analysis, if you want to make GHDL's implementation cleaner (to reduce the number of changes to/from Path/str), that'd be welcome. Nonetheless, we can have this merged as-is, and deal with that in an upcoming PR.

Thanks for fixing this!

Issue #934, opened 2023-05-01
Use relative paths for submodules


abaebae:
Using relative paths has some advantages
- Users get submodules over their preferred transport
- No need to tweak `.gitmodules` if you mirror VUnit

PR #935, opened 2023-05-01
Make submodule paths relative

abaebae:
- Users get submodules over their preferred transport
- No need to tweak `.gitmodules` if you mirror VUnit


abaebae:
Reolves #934

jobtijhuis:
I would like this to be fixed as well. For me the same benefit applies, in that it makes the repo function when you have a local mirror of it and you want to pull the submodules locally instead of from github. It is also better to not force people to use a certain transport method if not required.

LarsAsplund:
Closes #934

Issue #851, opened 2022-07-12
Unaligned transfers and Endian option for Axi write slave


basedVecna:
It seems that the AXI lite write slave does not support unaligned transfers (If an address in unaligned it is clipped down to the lower boundary), the memory is also only uses little endianness, independent of the endianess set in the memory_t. Should this be implemented or is it a purposeful decision?

crdavis12:
I've also just run into the issue of unaligned transfers not working in the `axi_read_slave`. Haven't gotten there yet but I suspect `axi_write_slave` will be similar. Is there any more discussion on this that I may be missing? I am making some updates to these as I work through my design.

PR #936, opened 2023-05-16
Unaligned addresses with axi_read_slave

crdavis12:
Fixing how `axi_read_slave` handles unaligned addresses.

Let's say your memory looks like this:
    byte 0 = 0x00
    byte 1 = 0x01
    byte 2 = 0x02
    byte 3 = 0x03
    byte 4 = 0x04
    byte 5 = 0x05

AXI4 read request looks like this:
    bus width = 4 bytes
    ARADR = 2
    ARLEN = 0 (1 word)

On RDATA I would expect to see `0x03020100` or maybe `0x0302----` since the lower 2 bytes aren't valid.

Instead you get `0x03020504` for the first transfer, `0x07060908` for the second, etc. This change fixes the rotated data.

Mentioned in #851 but not exactly the same issue.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,10 +1,10 @@
 name: 'coverage'
 
 on:
-  push:
+#  push:
   workflow_dispatch:
-  schedule:
-    - cron: '0 0 * * 5'
+#  schedule:
+#    - cron: '0 0 * * 5'
 
 jobs:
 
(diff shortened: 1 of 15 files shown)


crdavis12:
TODO: Need to update tests

crdavis12:
Any word on when this might be able to be merged? I heard something about breaking verification components out into a different repo -- planning to wait for that first?

crdavis12:
@LarsAsplund -- any chance of getting this released as a 4.7.1 patch? We've had internal tools pointing to this branch for a while now but would be nice to get rid of that workaround.

PR #939, opened 2023-05-27
Add type hints to 'parsing/' package

lavovaLampa:
Initial batch of type hints. Also refactored token kinds into an Enum, as Python 3.6 is no longer supported.

Currently depends on some types provided by package 'typing-extensions', in case this new dependency is unwelcome, I can remove dependent type hints.


umarcor:
CI is failing because the required package is not installed. You can specify it in https://github.com/VUnit/vunit/blob/master/pyproject.toml#L64-L80, since all the tests are executed through tox.

lavovaLampa:
While setuptools is trying to determine dependencies, it has to run the `setup.py` script. It then tries to import other modules, running into an import from `typing_extensions`. AFAIK, setuptools builds packages in isolated environment, meaning this dependency is not present at the time of dependency resolution (makes sense).

lavovaLampa:
Would this solution be OK?

Project metadata could be alternatively moved into `pyproject.toml`. In any case this should solve the issue of dependency definitions during build.

umarcor:
@lavovaLampa could please make https://github.com/VUnit/vunit/pull/939/commits/7cfdaec780e93b11fcd151c02cecb050b5095d54 a separated PR (which we can merge before this)?

lavovaLampa:
Opened #942

PR #945, opened 2023-07-12
Fix issues when trying to parameterize Verilog modules

### LINKED ISSUE #944: Trying to parameterize a Verilog module that name is not lower-case causes error.
When executing a Library.module(), when a Verilog module has non-lower-case name an error occurs.
This is due to the fact that both [Library.entity()](https://github.com/VUnit/vunit/blob/58dcffdb9a811e7520c656c702c074227751c106/vunit/ui/library.py#L452) (VHDL) and [Library.module()](https://github.com/VUnit/vunit/blob/58dcffdb9a811e7520c656c702c074227751c106/vunit/ui/library.py#L467) (Verilog) generate call to [Library.test_bench()](https://github.com/VUnit/vunit/blob/58dcffdb9a811e7520c656c702c074227751c106/vunit/ui/library.py#L481), that makes the name always lower-case.

Above calls trying to parameterize any non-lower-case Verilog module to fail.

piotrva:
This fixes #944 by trying to execute an original version (case-insensitive for VHDL compatibility) with the fall-back for use with Verilog and non-lower-case module names.

### DIFF
--- a/vunit/ui/library.py
+++ b/vunit/ui/library.py
@@ -486,9 +486,12 @@ class Library(object):
         :returns: A :class:`.TestBench` object
         :raises: KeyError
         """
-        name = name.lower()
-
-        return TestBench(self._test_bench_list.get_test_bench(self._library_name, name), self)
+        name_lower = name.lower()
+        
+        try:
+            return TestBench(self._test_bench_list.get_test_bench(self._library_name, name_lower), self)
+        except:
+            return TestBench(self._test_bench_list.get_test_bench(self._library_name, name), self)
 
     def get_test_benches(self, pattern="*", allow_empty=False):
         """


piotrva:
Hi, any chance for having someone look at this PR?
We use VUnit in our company and for now we had to switch to my unofficial branch.

oscargus:
I think it maybe is better to either:
1. Move the case-handling to `get_test_bench`
2. Do try-except on `get_test_bench` only and then only check `KeyError`.

piotrva:
@oscargus I modified it as you suggested.

oscargus:
Thanks! (I should probably have mentioned that I cannot approve or merge anything, but I think/hope that the suggested changes increases the chance of having it merged...)

piotrva:
Can you possibly contact someone who is in charge of such approvals?
This issue is preventing us for use of official VUnit repo in our workflow what is suboptimal at least...

oscargus:
I can try to ping in @LarsAsplund and @kraigher . My impression is that they have been quite busy recently though. (There may be other people that have the correct powers as well.)

piotrva:
Any chance to have this fix merged any time soon?

piotrva:
More than a year passed - is this repository still maintained somehow?

LarsAsplund:
@piotrva Sorry for the delay. VUnit has a lot of users, which is good, but unfortunately we don't have maintainer bandwidth to meet that. Working on solutions...

Regarding your PR. First of all, all tests must be green for a PR to pass. Not sure why they went red the last time since the logs are deleted when they get too old.

I would also recommend a solution that doesn't involve code related to VHDL as this is a Verilog issue. What I recommend is that you don't let `module` call `test_bench` but rather let it return what `test_bench` returns, thereby avoiding the `name = name.lower()` call.

Provided the all tests pass after this change, there should also be a test covering this issue. I would recommend looking into tests/acceptance/artificial/verilog/ to see if those tests can be extended to include this casing issue

piotrva:
Hi, thanks for the response - is there a way the tests can be run again so I can see logs?

PR #947, opened 2023-07-15
Updated common log interface.

LarsAsplund:
Update common log interface as outlined in https://github.com/OSVVM/OSVVM/issues/80

### DIFF
--- a/examples/vhdl/osvvm_log_integration/osvvm_integration/AlertLogPkg.vhd
+++ b/examples/vhdl/osvvm_log_integration/osvvm_integration/AlertLogPkg.vhd
@@ -944,7 +944,7 @@ package body AlertLogPkg is
         if WriteAlertErrorCountVar then
           return ErrorCount + 1;
         else
-          return no_val;
+          return -1;
         end if ;
       end;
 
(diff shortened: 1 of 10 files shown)


umarcor:
@LarsAsplund please add a fragment to https://github.com/VUnit/vunit/tree/master/docs/news.d corresponding to this PR.

LarsAsplund:
@umarcor Actually, it was already added with reference to #776. See https://github.com/VUnit/vunit/blob/master/docs/news.d/776.feature.rst

umarcor:
In those cases, just copy and paste the file with a different number in the filename but exactly the same content. See, for instance, https://github.com/VUnit/vunit/commit/e24d8186a0e876328279e9d95a913d8248a829d3 and http://vunit.github.io/release_notes.html#features.

I would like to contribute an enhancement to towncrier in order to handle these cases better. Yet, for now, this is the solution.

LarsAsplund:
@umarcor Will do.

LarsAsplund:
@umarcor Hm, my latest push with common log updates came with its own branch/PR. Should that have generated yet another news item?

umarcor:
@LarsAsplund yes. in the end, we want all the issues/PRs related to this change/item/feature to show as a single list in http://vunit.github.io/release_notes.html#features.
However, if you are going to push a `docs/news.d: cp 776 949` or `docs/news.d: cp 947 949` only, do push that to master straightaway.

Issue #946, opened 2023-07-12
NVC -M support doesn't work with set_sim_option


mschiller-nrao:
It appears that nvc simulator only supports -M as a global option (eg before the "command" that tells nvc to analyze, elaborate or simulate).  This suggests that it needs to be implemented like "heap_size" is currently implemented so that the -M64m (or whatever to set the max elaboration size) can be set for a design.

This is necessary to support large designs in nvc

Eg this does not work:
```
[...]

PR #948, opened 2023-07-16
sim_if/nvc.py: add a new nvc.global_flags option

nickg:
This can be set for compilation/simulation to pass additional global arguments to the nvc command. See issue #946.

### DIFF
--- a/docs/py/opts.rst
+++ b/docs/py/opts.rst
@@ -27,6 +27,14 @@ The following compilation options are known.
    Extra arguments passed to ModelSim ``vlog`` command when compiling Verilog files.
    Must be a list of strings.
 
+``nvc.a_flags``
+   Extra arguments passed to ``nvc -a`` command during compilation.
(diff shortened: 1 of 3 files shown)


docs/py/opts.rst
```diff
@@ -27,6 +27,14 @@ The following compilation options are known.
    Extra arguments passed to ModelSim ``vlog`` command when compiling Verilog files.
    Must be a list of strings.
 
+``nvc.a_flags``
```
umarcor, line 30: Might we worth adding a "Global Options" or "Common Options" section, either at the beginning or the end of the page? Rather than repeating the `nvc.global_args` in both sections.

umarcor: My bad. It's actually two different options with the same name. Users might want to set "global sim options" but not "global elaboration options", or the other way round.
Therefore, it's good as is.

PR #955, opened 2023-09-05
Fix pip package name in Installing page

Xcodo:
The `pip` command seems to refer to the wrong package name.

Looks like the correct package is https://pypi.org/project/vunit-hdl/.


LarsAsplund:
We can't use ``vunit`` on PYPI because that was already taken so we use ``vunit_hdl`` instead. If that was the name we also used when importing into Python we must use an underscore since dashes are not legal in identifiers. PYPI prefers dashes so to handle this ``pip`` changes underscores to dashes. Either way works.

PR #961, opened 2023-09-28
Added/Fixed Typing for all files in vunit.ui package

g0t00:


vunit/com/codec_vhdl_package.py
```diff
@@ -148,7 +148,7 @@ def _create_enumeration_of_all_msg_types(self):
 
         msg_type_enumeration_types = []
         for record in self.record_types:
-            if record.elements[0].identifier_list[0] == "msg_type":
+            if record.elements[0].identifier_List[0] == "msg_type":
```
oscargus, line 151: Is this really correct? Seems a bit like a search-and-replace-error. `list[` -> `List[`

vunit/test/bench.py
```diff
@@ -13,7 +13,7 @@
 import re
 import bisect
 import collections
-from collections import OrderedDict
+from typing import Any, Iterable, List, Union, OrderedDict
```
oscargus, line 16: Is there any benefit of importing `OrderedDict` from `typing`? Seems more natural to import it from `collections`.

vunit/ui/packagefacade.py
```diff
@@ -9,21 +9,33 @@
 """
 
 from pathlib import Path
+from typing import List, Optional, Union, TYPE_CHECKING
+from vunit.design_unit import DesignUnit
+
```
oscargus, line 14:
```suggestion

from vunit.design_unit import DesignUnit
```
Not sure if any linting rules are applied, but this seems to make more sense either way.

oscargus:
Typing always helps! Some minor comments (I have no powers to approve or merge or whatever, but for what it is worth).

vunit/com/codec_vhdl_package.py
g0t00: Thank you good catch. Thats exactly what happened.

vunit/test/bench.py
g0t00: I kind of assumed that its required for the type generic annotations. Just checked and that not true. Changed back.

vunit/ui/packagefacade.py
g0t00: I do not really catch what your saying. But I did not find clear style re: import statements.
Also after doing all the changes I realized that vunit uses relative imports for most/all internal imports.
I personally do not like relative imports in python, but for consistencies sake I would change. If somebody from the core team can pitch in...

oscargus: Sorry for being unclear, the point was that if there is an empty line, it should probably not happen between the vunit imports.

Now
```
from typing...
from vunit...

from vunit...
```

Better
```
from typing...

from vunit...
from vunit...
```

(Tools like `isort` would probably have done something like the latter anyway.)

g0t00: Ah yes, thank you. Totally agreed.

(normally I just let isort handle imports)

umarcor:
After rebasing on top of branch `master`, the following errors are produced in CI:

# Linting

https://github.com/VUnit/vunit/actions/runs/8236613480/job/22523448088?pr=961#step:5:84
```
 ************* Module vunit.sim_if.__init__
vunit/sim_if/__init__.py:189:0: R0022: Useless option value for 'disable', 'no-self-use' was moved to an optional extension, see https://pylint.readthedocs.io/en/latest/whatsnew/2/2.14/summary.html#removed-checkers. (useless-option-value)
************* Module vunit.ui.preprocessor
vunit/ui/preprocessor.py:25:0: R0022: Useless option value for 'disable', 'no-self-use' was moved to an optional extension, see https://pylint.readthedocs.io/en/latest/whatsnew/2/2.14/summary.html#removed-checkers. (useless-option-value)
```

# Building sdist

https://github.com/VUnit/vunit/actions/runs/8236613480/job/22523449340?pr=961#step:5:43

```
   File "/home/runner/work/vunit/vunit/vunit/configuration.py", line 192, in <module>
    class ConfigurationVisitor(object):
  File "/home/runner/work/vunit/vunit/vunit/configuration.py", line 200, in ConfigurationVisitor
    def get_configuration_dicts(self) -> List[OrderedDict[Any, Configuration]]:
TypeError: 'type' object is not subscriptable
```

---

Furthermore:

- Please add a news fragment to [docs/news.d/](https://github.com/VUnit/vunit/tree/master/docs/news.d). See [vunit.github.io > Contributing > Release Notes Instructions](http://vunit.github.io/contributing.html#release-notes-instructions) and [towncrier.readthedocs.io > Tutorial > Creating News Fragments](https://towncrier.readthedocs.io/en/stable/tutorial.html#creating-news-fragments).
- Does mypy support specifying config options in `pyproject.toml` instead of using a separated file (`mypy.ini`)?

PR #965, opened 2023-10-20
Osvvm memory pkg build based on generic support

g0t00:
Analog to Osvvm included `osvvm.pro` built the correct MemoryPkg files based on simulator support for generic packages


LarsAsplund:
Thank you.

umarcor:
See #997.

PR #971, opened 2023-11-04
Fix check_relation using default checker

kraigher:
I am working on adding unused code warnings to VHDL LS and noticed that the `checker` argument to this procedure was unused.

### DIFF
--- a/vunit/vhdl/check/src/check.vhd
+++ b/vunit/vhdl/check/src/check.vhd
@@ -2263,7 +2263,7 @@ package body check_pkg is
     variable pass : boolean;
   begin
     -- pragma translate_off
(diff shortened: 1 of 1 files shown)


kraigher:
Note that the `run_stability_check`  has an unused `active_clock_edge` parameter which might be a bug: https://github.com/VUnit/vunit/blob/master/vunit/vhdl/check/src/check.vhd#L863C14-L863C31

LarsAsplund:
Nice feature! I like that it can see unused code in several steps. That one line will be unused once you remove another unused line.

I'm thinking such a check should be part of our lint job.

umarcor:
See #997.

PR #975, opened 2023-11-07
sim_if/modelsim: use vmap to find modelsim.ini

bjornfor:
* This creates a single source of truth for finding modelsim.ini. It is
  now possible to use MODELSIM=/path/to/custom/modelsim.ini and both
  ModelSim and vunit will use that. (The VUNIT_MODELSIM_INI environment
  variable can still be used to override modelsim.ini only for vunit.)
* This fixes detecting modelsim.ini in the quartus-prime-lite package
  from https://github.com/nixos/nixpkgs/issues, which exposes binaries
  at $prefix/bin, but not $prefix/modelsim.ini, because the latter is an
  FHS violation.


bjornfor:
@LarsAsplund: Friendly ping?

LarsAsplund:
@bjornfor Sounds like a good idea but there is a problem that needs to be solved. Removing the constraint from the `find_toolchain` call means that it will find the first `vsim` executable on the path. That can be both a Riviera-PRO and an Active-HDL installation. If you look into the simulator interfaces of those simulators and see how they use `find_toolchain`, you will see how they avoid finding each other and ModelSim/Questa. That should give you some ideas on how you can write a `not_aldec_simulator` constraint function.

bjornfor:
I pushed a new version, and fixed the unit tests, which I didn't realize this PR was breaking :sweat_smile:

I don't have a way to test `find_prefix_from_path()` yet (which uses the new `not_aldec_simulator()` function). So beware. (My integration test setup from a year ago is lost. And I don't have access to non-ModelSim/Questa simulators.)

bjornfor:
I think I've fixed the linter issues now.

bjornfor:
Friendly ping.

PR #989, opened 2024-02-23
check_equal support for fixed_pkg resolve #909

### LINKED ISSUE #909: Feature Request: check_equal(...) overloads for ieee.fixed_pkg.ufixed and ieee.fixed_pkg.sfixed
Hi all,

It would be very helpful if VUnit had support for using `check_equal(...)` on the `ieee.fixed_pkg` data types.

Currently I have to do all my testbenches using fixed point arithmetic like this:

```vhdl
check_true(A = B);
```

So that VUnit uses the equality operator from the fixed_pkg.

For verbosity of the test cases it would be better to have:

```vhdl
check_equal(A, B);
```

or even better, if the `check_equal` function could directly convert from the string representation of a fixed point number like this:

```vhdl
check_equal(A, "-13.625");
```

Since the fixed point arithmetic package is part of VHDL-2008 now, I think others might find this useful too.

abaebae:
How do you want to handle the string representation of fixed_pkg types? I used the `to_string` function from the fixed_pkg. Do you prefer the `to_nibble_string` style or a merge of both styles?

### DIFF
--- a/vunit/vhdl/check/run.py
+++ b/vunit/vhdl/check/run.py
@@ -11,9 +11,11 @@ from vunit.check_preprocessor import CheckPreprocessor
 
 import vunit.vhdl.check.tools.generate_check_equal as generate_check_equal
 import vunit.vhdl.check.tools.generate_check_match as generate_check_match
+import vunit.vhdl.check.tools.generate_check_equal_2008p as generate_check_equal_2008p
 
 generate_check_equal.main()
 generate_check_match.main()
+generate_check_equal_2008p.main()
 
 VU = VUnit.from_argv()
 VU.add_vhdl_builtins()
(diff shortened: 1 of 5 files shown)


creiter64:
I'm not an expert in VUnits code, but this looks quite good. About the string representation, I think since `fixed_pkg` already has one and it's part of the VHDL standard, it should be used.

umarcor:
After updating license copyright years and running black, the remaining CI errors are now related to the code in this PR. Check the following format mismatch:

```
Got:

   logger = check

   log_level = pass

   msg = Equality check passed - Got 1010.0101 (1.03125e1).

   file_name:line_num = :0



expected:

   logger = check

   log_level = pass

   msg = Equality check passed - Got 1010.0101 (10.3125).

   file_name:line_num = :0
```

abaebae:
It seems that every simulator had a slightly different default format for the `to_string(real)` implementation. The format is now explicitly set to the format string `"%f"`. That should lead to a consistent output style. The format style is up for debate.

umarcor:
See #997.

PR #990, opened 2024-02-29
Fix regex for VHDL component instantiations

javValverde:
- Add comments to the multiline regex
- Extend unittest for vhdl_parser

### DIFF
--- a/vunit/vhdl_parser.py
+++ b/vunit/vhdl_parser.py
@@ -85,9 +85,16 @@ class VHDLDesignFile(object):  # pylint: disable=too-many-instance-attributes
         )
 
     _component_re = re.compile(
-        r"[a-zA-Z]\w*\s*\:\s*(?:component)?\s*(?:(?:[a-zA-Z]\w*)\.)?([a-zA-Z]\w*)\s*"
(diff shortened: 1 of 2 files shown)


vunit/vhdl_parser.py
```diff
@@ -85,9 +85,16 @@ def parse(cls, code):
...
-        r"(?:generic|port) map\s*\([\s\w\=\>\,\.\)\(\+\-\'\"]*\);",
-        re.IGNORECASE,
+        r"""
+        [a-zA-Z]\w*               # Label
+        \s*\:\s*                  # Semicolon
+        (?:component)?\s*         # Optional component keyword
+        (?:(?:[a-zA-Z]\w*)\.)?    # Optional library name
+        ([a-zA-Z]\w*)\s*          # Capture component name
+        (?:generic|port)\s+map\s* # Generic/port map
+        \(.*?\)\s*;               # Open and closing brackets
```
LarsAsplund, line 95: Rather than allowing anything between the opening and closing brackets I would suggest extending the previously more restrictive expression with what is needed to pass the new test case:

\([\s\w\=\>\,\.\)\(\+\-\'\"\*/]*\)

I cannot recall if the more restrictive expression is there for a reason that is not made explicit in the test cases. For that reason I prefer a more defensive modification.

PR #991, opened 2024-02-29
Add py.typed

LukasVik:
So that code that uses VUnit can be type checked with mypy against VUnit's type hints.

See here for background: https://peps.python.org/pep-0561/

I verified that the file is included in the PyPI releases by running "python3 setup sdist" and checking that it was present in "vunit/py.typed" in the release zip.

### DIFF
--- a/docs/news.d/991.feature.rst
+++ b/docs/news.d/991.feature.rst
@@ -1,0 +1,1 @@
+Add ``py.typed`` file to indicate the existence of Python typing hints.


umarcor:
@LukasVik this PR adds a single empty file. Is that intended?

LukasVik:
@umarcor  Yes this is correct. The file just needs to exist as a marker, its content does not matter.

https://peps.python.org/pep-0561/#packaging-type-information

umarcor:
Nice! Please, add a news fragment and we can have this merged.

LukasVik:
> Nice! Please, add a news fragment and we can have this merged.

@umarcor Done! Thank you!

LukasVik:
I reworded it, is better now.

PR #992, opened 2024-02-29
Minimize imports in an effort to lower simulation setup time

LukasVik:
These changes represent a 21% reduction of execution time for the 460 test cases in hdl-modules repo: https://github.com/hdl-modules/hdl-modules For some small testbenches, the reduction is as much as 42%. Note that this repo contains only small to medium sized testbenches (1-10 seconds, with most around 1-4 seconds).

This is all using the GHDL simulator with GCC backend on Linux. With ModelSim Intel FPGA starter edition 2020.1, the reductions is negligible (1-2%).

This change:

1. Stop using "context" in all VUnit packages. In most packages, importing with context includes a lot more than what is actually used.

2. Stop using "use work.x_pkg.all" in cases where only 1-3 things are used from the package. Instead, one explicit "use" clause for the things that are actually used.

   * Not done for the most common packages (run_pkg, check_pkg, logger_pkg).

   * Not done for includes from very small packages.

3. For packages that have head and body in different files, move all imports to the head file. As they were, split between the files, there was a lot of overlap of imports, and it what hard to get an overview of what is actually imported and what is actually used.

The first point represents roughly 80% of the performance gain.

Note that in order to reach the performance gains listed above, the corresponding changes have to be made to all testbenches and simulation code in the user repo also. If the testbench includes "vunit_context", "com_context" etc, the performance gains will not be realized since everything will be imported anyway.

### DIFF
--- a/vunit/vhdl/check/src/check.vhd
+++ b/vunit/vhdl/check/src/check.vhd
@@ -6,15 +6,6 @@
 --
 -- Copyright (c) 2014-2024, Lars Asplund lars.anders.asplund@gmail.com
 
-library ieee;
-use ieee.std_logic_1164.all;
-use ieee.numeric_std.all;
-use std.textio.all;
-use work.checker_pkg.all;
-use work.string_ops.all;
-use work.location_pkg.all;
-use work.integer_vector_ptr_pkg.all;
-
 package body check_pkg is
   type boolean_vector is array (natural range <>) of boolean;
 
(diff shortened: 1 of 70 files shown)


LukasVik:
Hello Lars!

1. I do not, I did it all by hand. It took a while 🙈. Though I would love to have such a tool available. Something like isort in Python that groups and sorts the import, and some pylint equivalent that checks for unused imports. That would be lovely. Could probably be built on top of rust_hdl if someone wants to put in the work.
2. I agree fully. I think in general in programming keeping the scope clean, i.e. importing as little as possible, is a good thing. So the programming language should help with that. Would be great if VHDL "use" could be done in the compact way that you suggest.
4. Yes that is true. I see no change like this in my commit, but we can keep it in mind for the future!

Thank you!

LukasVik:
I have pushed a fix so that it is now green!

umarcor:
See #997.

PR #994, opened 2024-03-04
Drive invalid in AXI VCs

LukasVik:
When the 'valid' signal is low for an AXI (-Lite) channel, the signals of the channel should be driven with 'X'.
They should not keep their value from the last time the channel was valid.
This can hide errors in the DUT if the DUT samples values in the wrong clock cycle.

This change is largely inspired by the AXI-Stream VCs.

Note that this changes the default behavior of the three VCs, so in some sense it is a breaking change.
This could be avoided by setting the default value ``drive_invalid`` generic to ``false``.
Though having it set to ``true`` is symmetric with the AXI-Stream VCs.
And in my personal opinion it should be enabled in pretty much all cases, to get a stronger test.

### DIFF
--- a/vunit/vhdl/verification_components/src/axi_lite_master.vhd
+++ b/vunit/vhdl/verification_components/src/axi_lite_master.vhd
@@ -19,10 +19,12 @@ context work.com_context;
 context work.vunit_context;
 
 entity axi_lite_master is
-  generic(
-    bus_handle : bus_master_t
+  generic (
+    bus_handle : bus_master_t;
+    drive_invalid : boolean := true;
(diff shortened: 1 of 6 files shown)


LukasVik:
Note that this is a reopen of #802, which I accidentally closed by deleting my fork, not realizing that this would close the PR. Could not find a way to open that again, so I create this new PR instead.

The diff is exactly the same, which has been reviewed by two people in #802. I have added two commits above, however, that fix some minor formatting and comments.

LarsAsplund:
I'm ok with these changes but still some work on the test cases to have them pass. At this point, breaking changes are ok since next release is v5.0.

PR #1002, opened 2024-03-12
Support arbitrary waveform viewer

### LINKED ISSUE #970: Parameters for alternative wave form viewer
We are developing an alternative wave form viewer: https://gitlab.com/surfer-project/surfer and once it is mature enough, it would be nice to add support for it in VUnit. Right now there are the `--gtkwave-fmt` and `--gtkwave-args` arguments that determines which format is used and how GTKWave is called. However, I'd like to open a discussion about what the best way to implement a switch for wave form viewer would be?

My spontaneous idea is to rename `--gtkwave-fmt` to `--wave-fmt` or `--viewer-fmt` (still supporting both, but only document the latter) and then have a `--viewer`-argument (which may default to GTKWave). Possibly also rename `--gtkwave-args` to `--viewer-args` and pass whatever is there to the select viewer?

Any ideas or input regarding this?

oscargus:
Closes #970 

Haven't really had time to test this fully, so putting in draft. But since it seems like 5.0 is upcoming, this may be something to consider.

Basically, there is a `--viewer` argument that lets you select a wave form viewer. Also, all `--gtkwave-*` arguments are renamed `--viewer-*`, but with aliases.

(At least that is the idea once fully working.)

Motivated by https://surfer-project.org/ that is beginning to gain some traction. (Although possible to run in the browser there are also native versions.)

### DIFF
--- a/docs/news.d/1002.feature.rst
+++ b/docs/news.d/1002.feature.rst
@@ -1,0 +1,11 @@
+[GHDL/NVC] Arbitrary waveform viewers are now supported by passing the `--viewer`
+command line arugment. As a consequence, ``ghdl.gtkwave_script.gui`` and
+``nvc.gtkwave_script.gui`` are deprecated in favour of ``ghdl.viewer_script.gui``
+and ``nvc.viewer_script.gui``, respectively. The ``--gtkwave-args`` and
+``--gtkwave-fmt`` command line argument is deprecated in favour of ``--viewer-args``
+and ``--viewer-fmt``, respectively. ``ghdl.viewer.gui`` and ``nvc.viewer.gui`` can
+be used to set the preferred viewer from the run-file. Finally, if the requested
+viewer does not exists, ``gtkwave`` or ``surfer`` is used, in that order. This
+also means that VUnit uses ``surfer`` if ``gtkwave`` is not installed.
+
+[NVC] It is possible to get VCD waveform files by passing ``--viewer-fmt=vcd``.
--- a/docs/news.d/1002.feature.rst
+++ b/docs/news.d/1002.feature.rst
@@ -1,11 +1,11 @@
-[GHDL/NVC] Arbitrary waveform viewers are now supported by passing the `--viewer`
-command line arugment. As a consequence, ``ghdl.gtkwave_script.gui`` and
+[GHDL/NVC] Arbitrary waveform viewers are now supported by passing the ``--viewer``
+command line argument. As a consequence, ``ghdl.gtkwave_script.gui`` and
 ``nvc.gtkwave_script.gui`` are deprecated in favour of ``ghdl.viewer_script.gui``
 and ``nvc.viewer_script.gui``, respectively. The ``--gtkwave-args`` and
-``--gtkwave-fmt`` command line argument is deprecated in favour of ``--viewer-args``
+``--gtkwave-fmt`` command line arguments are deprecated in favour of ``--viewer-args``
 and ``--viewer-fmt``, respectively. ``ghdl.viewer.gui`` and ``nvc.viewer.gui`` can
-be used to set the preferred viewer from the run-file. Finally, if the requested
-viewer does not exists, ``gtkwave`` or ``surfer`` is used, in that order. This
-also means that VUnit uses ``surfer`` if ``gtkwave`` is not installed.
+be used to set the preferred viewer from the run-file. If no viewer is explicitly
+requested, ``gtkwave`` or ``surfer`` is used, in that order. This also means that
+VUnit now uses ``surfer`` if ``gtkwave`` is not installed.
 
 [NVC] It is possible to get VCD waveform files by passing ``--viewer-fmt=vcd``.
--- a/vunit/sim_if/_ossmixin.py
+++ b/vunit/sim_if/_ossmixin.py
@@ -1,0 +1,61 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# Copyright (c) 2014-2024, Lars Asplund lars.anders.asplund@gmail.com
+"""
+Viewer handling for GHDL and NVC.
+"""
+
+
+class OSSMixin:
+    """
+    Mixin class for handling common viewer functionality for the GHDL and NVC simulators.
+    """
+
+    __slots__ = (
+        "_gui",
+        "_viewer",
+        "_viewer_fmt",
+        "_viewer_args",
+        "_gtkwave_available",
+        "_surfer_available",
+    )
+
+    def __init__(self, gui, viewer, viewer_fmt, viewer_args):
+        self._gui = gui
+        self._viewer_fmt = viewer_fmt
+        self._viewer_args = viewer_args
+        self._viewer = viewer
+        if gui:
+            self._gtkwave_available = self.find_executable("gtkwave")
+            self._surfer_available = self.find_executable("surfer")
+
+    def _get_viewer(self, config):
+        """
+        Determine the waveform viewer to use.
+
+        Falls back to gtkwave or surfer, in that order, if none is provided or the provided
+        cannot be found.
+        """
+        viewer = self._viewer or config.sim_options.get(self.name + ".viewer.gui", None)
+
+        if viewer is None:
+            if self._gtkwave_available:
+                viewer = "gtkwave"
+            elif self._surfer_available:
+                viewer = "surfer"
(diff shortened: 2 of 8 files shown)


umarcor:
> But since it seems like 5.0 is upcoming, this may be something to consider.

We cannot deprecate gtkwave arguments in the next release. We need to first release a version with the new feature, announcing that "old" aliases will be removed in a future version. Then, in a future release, we can remove the deprecated keywords/arguments.
Therefore, this PR will not introduce breaking changes, and we can have it merged as soon as it is ready (no matter if it's 5.0 or 5.1).

Other than that, this looks very interesting. So, if it's ready by the time we want to tag 5.0, we can sure have it merged before.

LarsAsplund:
@oscargus Previously, we've handled them explicitly. Here is an example: https://github.com/VUnit/vunit/blob/2c0b75e40aa59a42c51b5f6b2c40767f1b2b83d4/vunit/sim_if/ghdl.py#L255

vunit/sim_if/ghdl.py
```diff
@@ -71,8 +73,8 @@ def from_args(cls, args, output_path, **kwargs):
             output_path=output_path,
             prefix=prefix,
             gui=args.gui,
-            gtkwave_fmt=args.gtkwave_fmt,
-            gtkwave_args=args.gtkwave_args,
+            viewer_fmt=args.viewer_fmt,
+            viewer_args=args.viewer_args,
```
LarsAsplund, line 77: Need to assign the `viewer` argument here:

`viewer=args.viewer`

oscargus:
Now I finally got around to finish this.

The news-item summarizes the changes quite good, but brief.

It seems like the emitted warnings are filtered, but I cannot really figure out how and where...

As a missing waveform viewer now raises after the simulation (which is not ideal, but I couldn't figure out another way to define the waveform viewer in the run-file), I had to remove one test. Let me know if this should be reverted or if there is another way to specify waveform viewer in the run-file.

It is of course possible to just limit the choice to surfer or gtkwave. That would simplify things a bit.

vunit/sim_if/nvc.py
```diff
@@ -299,14 +315,44 @@ def simulate(self, output_path, test_suite_name, config, elaborate_only):  # pyl
...
+    def _get_viewer(self, config):
+        """
+        Determine the waveform viewer to use.
+
+        Falls back to gtkwave or surfer, in that order, if none is provided or the provided
+        cannot be found.
+        """
+        viewer = self._viewer or config.sim_options.get(self.name + ".viewer.gui", None)
+
+        if viewer is None and not self._gtkwave_available and not self._surfer_available:
```
LarsAsplund, line 345: I think these two if statements should be replaced by this:

```
        if viewer is None:
            if self._gtkwave_available:
                viewer = "gtkwave"
            elif self._surfer_available:
                viewer = "surfer"
            else:
                raise RuntimeError("No viewer found. GUI not possible. Install GTKWave or Surfer.")

        elif not self.find_executable(viewer):
            raise RuntimeError(
                f"Cannot find the {viewer} executable in the PATH environment variable. GUI not possible"
            )
```

One reason is that if `viewer` is `None`, the second if statement will fail since the `find_executable` method expects a string. The other reason is that I think we should fail immediately if the given viewer can't be found. If we set the viewer to `smurfer` it's better to fail fast and fix the typo rather than being surprised by GTKWave popping up. Optionally, we can also print the available viewers in the error message.

vunit/sim_if/ghdl.py
```diff
@@ -15,6 +15,7 @@
 import shlex
 import re
 import shutil
+import warnings
```
LarsAsplund, line 18: Remove and replace `warning.warn` with `LOGGER.warning`

vunit/sim_if/nvc.py
```diff
@@ -14,6 +14,7 @@
 import subprocess
 import shlex
 import re
+import warnings
```
LarsAsplund, line 17: Remove

vunit/sim_if/ghdl.py
```diff
@@ -394,6 +409,30 @@ def simulate(self, output_path, test_suite_name, config, elaborate_only):  # pyl
...
+    def _get_viewer(self, config):
+        """
+        Determine the waveform viewer to use.
+
+        Falls back to gtkwave or surfer, in that order, if none is provided or the provided
+        cannot be found.
+        """
+        viewer = self._viewer or config.sim_options.get(self.name + ".viewer.gui", None)
+
+        if viewer is None and not self._gtkwave_available and not self._surfer_available:
```
LarsAsplund, line 421: See comment for NVC

vunit/sim_if/nvc.py
oscargus: Good thinking!

oscargus:
Now I have sorted things out (and passed the lint, wasn't easy with the long warning message...). I ended up factoring out the common viewer code to an `OSSMixin`-class that is used for both GHDL and NVC.

I'm a bit tempted to create a third class of options, maybe `viewer_options` as the viewer will be the same for/independent of ghdl and nvc it will also make sense to have common `viewer`, `viewer_fmt`, etc. This will also allow checking if it exists before actually trying to open the viewer (hence, keeping the previous behavior).

Taking the idea of being able to generate waves without opening the viewer, #1003 and #1042, one may also consider other naming.

vunit/sim_if/_ossmixin.py
```diff
@@ -0,0 +1,61 @@
...
+        self._viewer = viewer
+        if gui:
+            self._gtkwave_available = self.find_executable("gtkwave")
+            self._surfer_available = self.find_executable("surfer")
+
+    def _get_viewer(self, config):
+        """
+        Determine the waveform viewer to use.
+
+        Falls back to gtkwave or surfer, in that order, if none is provided or the provided
```
LarsAsplund, line 38: If the provided isn't found it no longer falls back on gtkwave or surfer.

docs/news.d/1002.feature.rst
```diff
@@ -0,0 +1,11 @@
+[GHDL/NVC] Arbitrary waveform viewers are now supported by passing the `--viewer`
+command line arugment. As a consequence, ``ghdl.gtkwave_script.gui`` and
+``nvc.gtkwave_script.gui`` are deprecated in favour of ``ghdl.viewer_script.gui``
+and ``nvc.viewer_script.gui``, respectively. The ``--gtkwave-args`` and
+``--gtkwave-fmt`` command line argument is deprecated in favour of ``--viewer-args``
+and ``--viewer-fmt``, respectively. ``ghdl.viewer.gui`` and ``nvc.viewer.gui`` can
+be used to set the preferred viewer from the run-file. Finally, if the requested
```
LarsAsplund, line 7: No falling back when requested viewer doesn't exist

vunit/sim_if/ghdl.py
```diff
@@ -382,10 +391,19 @@ def simulate(self, output_path, test_suite_name, config, elaborate_only):  # pyl
         except Process.NonZeroExitCode:
             status = False
 
+        if config.sim_options.get(self.name + ".gtkwave_script.gui", None):
+            warn_str = (
+                "%s.gtkwave_script.gui is deprecated and will be removed " % self.name  # pylint: disable=C0209
```
LarsAsplund, line 396: Is there a reason for why f-strings can't be used so that we don't have to suppress C0209?

vunit/sim_if/_ossmixin.py
```diff
@@ -0,0 +1,61 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# Copyright (c) 2014-2024, Lars Asplund lars.anders.asplund@gmail.com
+"""
+Viewer handling for GHDL and NVC.
+"""
+
+
+class OSSMixin:
```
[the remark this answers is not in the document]
oscargus: Open source software. But wasn't very happy with that to renamed to ViewerMixin.

(I had some idea to merge more of the NVC and GHDL classes, but since it now ended up to be viewer only...)

vunit/sim_if/ghdl.py
oscargus: There were linter warnings about that (but I didn't do it in a proper way either), that I couldn't get rid of (either that or too long lines).

The reason is that one should call loggers with `warning("%s", foo)` so that the string is formatted upon calling warning. 

Reminding me about how it should be done, the correct approach is now used.

oscargus:
All comments should be fixed. Sorry about that!

They are in a separate commit to make it easier to check, but I suggest to squash (I can squash if you do not have it enabled at GitHub).

PR #1005, opened 2024-03-19
Add GHDL backend detection support for mcode JIT

LukasVik:
Ongoing work on the GHDL master branch to use JIT makes VUnit backend detection fail. This makes all GitHub projects that use https://github.com/ghdl/setup-ghdl-ci with VUnit in their CI fail.

This commit adapts the regular expression so it works with newer as well as older GHDL versions with the mcode backend. As for the other backends, it looks like (https://github.com/ghdl/ghdl/commit/4b20cd9e67282b14232d7ff7f6d81df1155fb611#diff-0dc1fb1faeb66cc2780208579bbbe6a87011ad37a31f3d84dd64ff2224111fb6R530) they should all be affected by this change in the "ghdl --version" output. Or at least the LLVM backend?

I don't have access to GHDL built with GCC or LLVM, frankly I don't know how to build them. But I know that this PR fixes the issue for mcode. Perhaps there will be error reports or PRs from other people who use GCC/LLVM in the future.

### DIFF
--- a/docs/news.d/1005.feature.rst
+++ b/docs/news.d/1005.feature.rst
@@ -1,0 +1,1 @@
+[GHDL] Add support for mcode JIT in backend detection.
--- a/tests/unit/test_ghdl_interface.py
+++ b/tests/unit/test_ghdl_interface.py
@@ -87,6 +87,19 @@ warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
         check_output.return_value = version
         self.assertEqual(GHDLInterface.determine_backend("prefix"), "mcode")
 
+        version = b"""\
+GHDL 5.0.0-dev (4.0.0.r9.g77785e49e) [Dunoon edition]
+ Compiled with GNAT Version: 10.5.0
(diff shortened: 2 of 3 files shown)


umarcor:
@LukasVik, as far as I understand, this is just a change in a hardcoded string that is printed by `ghdl version`. That is, "mcode", "JIT", and "mcode JIT" are all the same, and they have been the same for years/decades. Precisely, mcode is an "in-memory just in time" backend, compared to LLVM/GCC which do not support JIT compilation/elaboration yet.
Therefore, the title, description and commit message of this PR are not correct. The issue is "updating the version/backend detection regular expression because the string changed upstream", but VUnit does already support "mcode JIT", support is not being added by this PR.

With regard to having this modification tested with other backends, do not worry about it because it's done in CI (at least LLVM). Since it's green, that means that your changes did not break anything.

LukasVik:
@umarcor Yes, when I wrote "add support" I was only referring to the backend detection regexp. I've updated the commit message and PR title to be more clear.

PR #1007, opened 2024-03-29
Add ghdl jit coverage support

tgingold:
add support for native coverage
    
Coverage was supported for the gcc backend using gcc coverage (option
-fprofile-arcs).  With version 4.0, ghdl also supports coverage for
the mcode backend (using option --coverage).
    
This patch handle supports for this new option.  It currently uses
new options not available in 4.0.  But ghdl 4.1 will be released after
the vunit changes.


LarsAsplund:
Thanks @tgingold. Looking forward to GHDL 4.1!

Issue #1014, opened 2024-05-08
Remove forced `--jit` from NVC


albydnc:
As showed in https://github.com/nickg/nvc/issues/887 , the jit can throw illegal instructions. It would be better if the `--jit` could be toggled off, instead of being forced on.

nickg:
This was a compatibility issue with LLVM 18 and is fixed in the latest 1.12.2 release.

PR #1015, opened 2024-05-08
Add option for disabling NVC `--jit`

albydnc:
Fix for issue https://github.com/VUnit/vunit/issues/1014

### DIFF
--- a/docs/py/opts.rst
+++ b/docs/py/opts.rst
@@ -218,3 +218,7 @@ The following simulation options are known.
 ``nvc.sim_flags``
    Extra simulation flags passed to ``nvc -r``.
    Must be a list of strings.
+
+``nvc.disable_jit``
+   Disable JIT.
+   Must be boolean, default False.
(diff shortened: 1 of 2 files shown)


LarsAsplund:
When it comes to simulator options that are not among the more common ones, we provide `nvc.sim_flags` rather than a more dedicated one. That would work in this case, right?

albydnc:
Yes, but with `nvc.sim_flags` you can only add options, while you can't disable the default ones.
This option needed to be disabled due to a bug in order for nvc to work. See https://github.com/VUnit/vunit/issues/1014#issuecomment-2113479491
Since it is fixed, I will close the PR, but I think there should be a way to remove all default flags.

LarsAsplund:
@albydnc Ah, now I see your point. You're right, there should be such a feature but let's await the next use case. Whatever the solution is, it should probably be common to all simulators.

albydnc:
I agree with you, it should be a generic thing.

PR #1020, opened 2024-05-27
Check 4kByte boundary from aligned address

### LINKED ISSUE #1018: Unaligned accesses to AXI read and write slaves wrongly fail 4kbyte boundary check
When doing unaligned accesses around a 4k boundary, 4k boundary check wrongly fail in AXI read and write slaves.

Example:
**Bus width:** 32 bits
**Start address:** 3841
**Burst length, beats:** 64 beats
**Burst length, bytes:** 255

When doing unaligned write accesses, by in this case starting from 3841, the first byte in the first beat will have it's strobe bit set to zero, i.e not writing that byte, which is expected. 

However, the access is still 64 beats long, as expected, but the `check_4kbyte_boundary` method in `axi_slave_private_pkg` will count the end address of the access from the start address of the burst, not the aligned address.
This will incorrectly assume that the burst ends at address 3841 + 4 * 64 - 1 = 4096, and not at address 4095 (which is what will happen when writing), but the access is still 64 beats long, one byte is just strobed out in the first beat.

A suggested patch here is to count from the starting address which is aligned.
```vhdl
procedure check_4kbyte_boundary(burst : axi_burst_t) is
  variable first_address, last_address : integer;
  variable first_page, last_page : integer;
begin
  first_address := burst.address - (burst.address mod data_size); -- Aligned
  last_address := first_address  + burst.size*burst.length - 1;

  first_page := first_address / 4096;
  last_page := last_address / 4096;

  if first_page /= last_page then
    fail("Crossing 4KByte boundary. First page = "
         & integer'image(first_page) & " (" & to_string(first_address) & "/4096)"
         & ", last page = "
         & integer'image(last_page) & " (" & to_string(last_address) & "/4096)");
  end if;
end procedure;
```

ru551n:
Fixes #1018

### DIFF
--- a/vunit/vhdl/verification_components/src/axi_slave_private_pkg.vhd
+++ b/vunit/vhdl/verification_components/src/axi_slave_private_pkg.vhd
@@ -438,10 +438,10 @@ package body axi_slave_private_pkg is
       variable first_page, last_page : integer;
     begin
       first_address := burst.address - (burst.address mod data_size); -- Aligned
-      last_address := burst.address + burst.size*burst.length - 1;
+      last_address := first_address + burst.size*burst.length - 1;
 
       first_page := first_address / 4096;
-      last_address := first_address + burst.size*burst.length - 1;
+      last_address := last_address / 4096;
 
       if first_page /= last_page then
         fail("Crossing 4KByte boundary. First page = "
--- a/vunit/vhdl/verification_components/src/axi_slave_private_pkg.vhd
+++ b/vunit/vhdl/verification_components/src/axi_slave_private_pkg.vhd
@@ -441,7 +441,7 @@ package body axi_slave_private_pkg is
       last_address := burst.address + burst.size*burst.length - 1;
 
       first_page := first_address / 4096;
-      last_page := last_address / 4096;
+      last_address := first_address + burst.size*burst.length - 1;
 
       if first_page /= last_page then
         fail("Crossing 4KByte boundary. First page = "
--- a/vunit/vhdl/verification_components/src/axi_slave_private_pkg.vhd
+++ b/vunit/vhdl/verification_components/src/axi_slave_private_pkg.vhd
@@ -441,7 +441,7 @@ package body axi_slave_private_pkg is
       last_address := first_address + burst.size*burst.length - 1;
 
       first_page := first_address / 4096;
-      last_address := last_address / 4096;
+      last_page := last_address / 4096;
 
       if first_page /= last_page then
         fail("Crossing 4KByte boundary. First page = "
--- a/vunit/vhdl/verification_components/test/tb_axi_write_slave.vhd
+++ b/vunit/vhdl/verification_components/test/tb_axi_write_slave.vhd
@@ -355,11 +355,10 @@ begin
     
     elsif run("Test unaligned write around 4kbyte boundary") then
       -- Do one beat write at unaligned address starting around 4kB boundary
-      len := 1;
       log_size := log_data_size;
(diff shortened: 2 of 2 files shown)


vunit/vhdl/verification_components/src/axi_slave_private_pkg.vhd
```diff
@@ -441,7 +441,7 @@ package body axi_slave_private_pkg is
       last_address := burst.address + burst.size*burst.length - 1;
 
       first_page := first_address / 4096;
-      last_page := last_address / 4096;
+      last_address := first_address + burst.size*burst.length - 1;
```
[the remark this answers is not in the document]
ru551n: Fixed, your suggestion was what was intended. Don't know what happened when pushing to remote. :/

vunit/vhdl/verification_components/src/axi_slave_private_pkg.vhd
```diff
@@ -438,10 +438,10 @@ package body axi_slave_private_pkg is
       variable first_page, last_page : integer;
     begin
       first_address := burst.address - (burst.address mod data_size); -- Aligned
-      last_address := burst.address + burst.size*burst.length - 1;
+      last_address := first_address + burst.size*burst.length - 1;
 
       first_page := first_address / 4096;
-      last_page := last_address / 4096;
+      last_address := last_address / 4096;
```
LarsAsplund, line 444: Should be `last_page`

ru551n: I apologize, fixed now.

vunit/vhdl/verification_components/test/tb_axi_write_slave.vhd
```diff
@@ -352,6 +352,15 @@ begin
       random_integer_vector_ptr(rnd, data, size * 2, 0, 255);
       buf := allocate(memory, length(data), permissions => no_access);
       transfer(x"2", buf, log_size, data);
+    
+    elsif run("Test unaligned write around 4kbyte boundary") then
+      -- Do one beat write at unaligned address starting around 4kB boundary
+      len := 1;
+      log_size := log_data_size;
+      buf := allocate(memory, 4096 - (2**log_data_size / 8) + 1, permissions => no_access); -- Unaligned address
```
LarsAsplund, line 360: `2**log_data_size / 8` makes more sense if the division by 8 is removed since  `2**log_data_size` is in bytes. Same on next line. The comment should be move to the next buffer allovation.

vunit/vhdl/verification_components/test/tb_axi_write_slave.vhd
```diff
@@ -352,6 +352,15 @@ begin
       random_integer_vector_ptr(rnd, data, size * 2, 0, 255);
       buf := allocate(memory, length(data), permissions => no_access);
       transfer(x"2", buf, log_size, data);
+    
+    elsif run("Test unaligned write around 4kbyte boundary") then
+      -- Do one beat write at unaligned address starting around 4kB boundary
+      len := 1;
```
LarsAsplund, line 358: This line can be removed.

LarsAsplund (commented):
After this, it should be good to merge

LarsAsplund:
There are some unrelated linting issues due to a new version of pylint. I'm fixing that on `master` such that you can rebase this branch.

LarsAsplund:
`master` is now fixed.

vunit/vhdl/verification_components/test/tb_axi_write_slave.vhd
ru551n: Ah now I got the relation of all the variables. 
Fixed

LarsAsplund:
Thanks @ru551n!

PR #1024, opened 2024-05-31
modelsim ini options for GUI mode

alexrayne:
modelsim ini options for GUI mode now can setup with via "modelsim.vsim_ini.gui" sim option.

* usage: 
```
simwave = "wave.do"
UI.set_sim_option("modelsim.vsim_ini.gui", { "ShutdownFile": str(simwave.resolve()) });
```

this provides option in modelsim.ini for GUI-mode simulation:
```
[vsim]
ShutdownFile=wave.do
```
That saves all wave-setup on UI close

### DIFF
--- a/vunit/sim_if/__init__.py
+++ b/vunit/sim_if/__init__.py
@@ -400,6 +400,27 @@ class ListOfStringOption(Option):
         except TypeError:
             fail()
 
+class DictOfStringOption(Option):
+    """
+    Must be a dictinary of strings
+    """
+
+    def validate(self, value):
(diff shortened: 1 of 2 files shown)


alexrayne:
modelsim shutdown do-file saves vsim option, that are overrides vunit one`s. 
Now cleanup that options from do-file before run modelsim, so we can use same do-file on multiple test-cases

LarsAsplund:
Do you know if there is a way to edit the `modelsim.ini` file from TCL?

PR #1025, opened 2024-05-31
Enchance time reporting in report.py and runner.py

tossegus:
This pull request includes the following updates/suggestions:

* Add timestamp at the start of new tests
* Parse time so that time is printed as hours, minutes and seconds instead of only seconds.
  (The breakpoint is at 60 seconds/60 minutes even though it might suffice to allow some slack to the number of seconds/minutes before incrementing them. I.e. 70 seconds could as well be printed as 70 seconds as 1 minute 10 seconds.)

### DIFF
--- a/docs/release_notes/4.8.0.rst
+++ b/docs/release_notes/4.8.0.rst
@@ -1,0 +1,5 @@
+Features
+~~~~~~~~
+
+- Update time reporting from test results.
+- Add timestamps to run tests.
--- a/docs/release_notes/4.8.0.rst
+++ b/docs/release_notes/4.8.0.rst
@@ -1,5 +1,0 @@
-Features
-~~~~~~~~
-
-- Update time reporting from test results.
-- Add timestamps to run tests.
--- a/vunit/test/report.py
+++ b/vunit/test/report.py
@@ -27,22 +27,21 @@ def get_parsed_time(time_in, max_time=0):
       max_time : Longest test time among tests. Optional parameter
     """
     time_str = ""
-    time_max = time_in if max_time == 0 else max_time
 
     (minutes, seconds) = divmod(time_in, 60)
     (hours, minutes) = divmod(minutes, 60)
-    if time_max >= 3600:
+    if max_time >= 3600 or time_in >= 3600:
         # If the longest test took 10 hours or more, pad the string to take this
         # into account.
-        padding = 5 if (time_max / 3600) % 60 >= 10 else 4
+        padding = 5 if (max_time / 3600) % 60 >= 10 else 4
         if hours > 0:
             time_str += f"{int(hours)} h ".rjust(padding)
         else:
             time_str += " " * padding
-    if time_max >= 60:
+    if max_time >= 60 or time_in >= 60:
         # If the longest test took an hour (or more), or the longest test took
         # 10 minutes or more, pad the string to take this into account.
-        padding = 7 if (time_max >= 3600) or ((time_max / 60) % 60 >= 10) else 6
+        padding = 7 if (max_time >= 3600) or ((max_time / 60) % 60 >= 10) else 6
         if minutes > 0:
             time_str += f"{int(minutes)} min ".rjust(padding)
         else:
@@ -50,7 +49,7 @@ def get_parsed_time(time_in, max_time=0):
 
     # If the longest test took a minute (or more), or the longest test
     # took 10 seconds or more, pad the string to take this into account.
-    padding = 6 if (time_max >= 60) or (time_max % 60 >= 10) else 5
+    padding = 6 if (max_time >= 60) or (max_time % 60 >= 10) else 5
     time_str += f"{seconds:2.1f} s".rjust(padding)
 
     return time_str
--- a/vunit/test/report.py
+++ b/vunit/test/report.py
@@ -30,10 +30,11 @@ def get_parsed_time(time_in, max_time=0):
 
     (minutes, seconds) = divmod(time_in, 60)
     (hours, minutes) = divmod(minutes, 60)
+
     if max(time_in, max_time) >= 3600:
         # If the longest test took 10 hours or more, pad the string to take this
         # into account.
-        padding = 5 if (max_time / 3600) % 60 >= 10 else 4
+        padding = len(f"{int(max_time // 3600)} h ")
         if hours > 0:
             time_str += f"{int(hours)} h ".rjust(padding)
         else:
@@ -41,7 +42,7 @@ def get_parsed_time(time_in, max_time=0):
     if max(time_in, max_time) >= 60:
         # If the longest test took an hour (or more), or the longest test took
         # 10 minutes or more, pad the string to take this into account.
-        padding = 7 if (max_time >= 3600) or ((max_time / 60) % 60 >= 10) else 6
+        padding = 7 if (max_time / 60 >= 10) else 6
         if minutes > 0:
             time_str += f"{int(minutes)} min ".rjust(padding)
         else:
@@ -49,7 +50,7 @@ def get_parsed_time(time_in, max_time=0):
 
     # If the longest test took a minute (or more), or the longest test
     # took 10 seconds or more, pad the string to take this into account.
-    padding = 6 if (max_time >= 60) or (max_time % 60 >= 10) else 5
+    padding = 6 if (max_time >= 10) else 5
     time_str += f"{seconds:2.1f} s".rjust(padding)
 
     return time_str
--- a/vunit/test/report.py
+++ b/vunit/test/report.py
@@ -18,6 +18,24 @@ from vunit.color_printer import COLOR_PRINTER
 from vunit.ostools import read_file
 
 
+def get_parsed_time(time_in):
+    """
+    Return string representation of input value
+    in hours, minutes and seconds.
+    """
+    time_str = ""
+    (minutes, seconds) = divmod(time_in, 60)
+    (hours, minutes) = divmod(minutes, 60)
+    if hours > 0:
+        time_str += f"{hours} hours, "
+    if minutes > 0:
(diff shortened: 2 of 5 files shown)


LarsAsplund:
I think there are two important things to consider:
1. It should be easy to get an accurate feel for the duration and that is what this PR addresses. 1 hour 20 minutes is easier to understand than 4800 seconds.
2. Given a long list of tests and their durations it should be easy to spot the slow ones. That is fairly easy today since a slow time is a longer number:

```
4800.0 seconds
48.0 seconds
```

It is trickier to see if we have a long list of this

```
11 minutes 11.1 seconds
2 hours 7.3 seconds
1 hour 51 minutes 43.4 seconds
```

A solution would be to have a more structured format. For example

```
00:11:11.1
02:00:07.3
01:51:43.3
```

It's still a bit hard to spot the longest duration so I think this is even better:

```
   11:11.1
 2:00:07.3
 1:51:43.3
```

We could also sort the summary:

```
   11:11.1
 1:51:43.3
 2:00:07.3
```


I also think some indication of unit of time would improve readability. Using abbreviations means that we don't have to
deal with the presence/absence of the plural s 

```
    11 min 11.1 s
1 h 51 min 43.3 s
2 h 00 min 07.3 s
```

If the longest time decides the number of digits, we can also make faster test suite reports more compact:

```
 7.3 s
11.1 s
43.3 s
```

Without a unit of time such a list would become ambiguous

tossegus:
> I think there are two important things to consider:
> 
>     1. It should be easy to get an accurate feel for the duration and that is what this PR addresses. 1 hour 20 minutes is easier to understand than 4800 seconds.
> 
>     2. Given a long list of tests and their durations it should be easy to spot the slow ones. That is fairly easy today since a slow time is a longer number:
> 
> 
> ```
> 4800.0 seconds
> 48.0 seconds
> ```
> 
> It is trickier to see if we have a long list of this
> 
> ```
> 11 minutes 11.1 seconds
> 2 hours 7.3 seconds
> 1 hour 51 minutes 43.4 seconds
> ```
> 
> A solution would be to have a more structured format. For example
> 
> ```
> 00:11:11.1
> 02:00:07.3
> 01:51:43.3
> ```
> 
> It's still a bit hard to spot the longest duration so I think this is even better:
> 
> ```
>    11:11.1
>  2:00:07.3
>  1:51:43.3
> ```
> 
> We could also sort the summary:
> 
> ```
>    11:11.1
>  1:51:43.3
>  2:00:07.3
> ```
> 
> I also think some indication of unit of time would improve readability. Using abbreviations means that we don't have to deal with the presence/absence of the plural s
> 
> ```
>     11 min 11.1 s
> 1 h 51 min 43.3 s
> 2 h 00 min 07.3 s
> ```
> 
> If the longest time decides the number of digits, we can also make faster test suite reports more compact:
> 
> ```
>  7.3 s
> 11.1 s
> 43.3 s
> ```
> 
> Without a unit of time such a list would become ambiguous

I think those are good points, and suggestions.

The one that resonated the most with me was the last (pair of) suggestion. I.e. using h/m/s and allowing the "largest" time unit to select if h or m is included or not.

LarsAsplund:
Now it looks like a normal PR on my side. I don't think you implemented sorting and that's good. If the list is sorted on time, we don't get the grouping of tests by testbenches which is another good way of sorting the results. For now, I think that should be kept. We can always add optional sorting methods later.

What's needed now is to extend the tests to cover hours and minutes and adding a news fragment for this change, see https://vunit.github.io/contributing.html#release-notes-instructions

LarsAsplund:
I ran an example to see what it looks like in reality. The only thing that I think can improve is how 1 min 7.2 s is aligned in the example below. In the summary, "7" should be aligned to the rightmost digit of the two digits showing seconds but it looks a bit weird with that alignment outside of the summary (see elapsed time and after the completion of the test)

image

tossegus:
> I ran an example to see what it looks like in reality. The only thing that I think can improve is how 1 min 7.2 s is aligned in the example below. In the summary, "7" should be aligned to the rightmost digit of the two digits showing seconds but it looks a bit weird with that alignment outside of the summary (see elapsed time and after the completion of the test)
> 
> image

Well, that looks just wrong. I'll make an update and add a few longer tests for test_test_report.py

vunit/test/report.py
```diff
@@ -18,6 +18,43 @@
...
+      max_time : Longest test time among tests. Optional parameter
+    """
+    time_str = ""
+
+    (minutes, seconds) = divmod(time_in, 60)
+    (hours, minutes) = divmod(minutes, 60)
+    if max(time_in, max_time) >= 3600:
+        # If the longest test took 10 hours or more, pad the string to take this
+        # into account.
+        padding = 5 if (max_time / 3600) % 60 >= 10 else 4
```
LarsAsplund, line 36: I suggest this is generalized to any `max_time`. Something like

`len(f"{max_time // 3600} h ")`

vunit/test/report.py
```diff
@@ -18,6 +18,43 @@
...
+        if hours > 0:
+            time_str += f"{int(hours)} h ".rjust(padding)
+        else:
+            time_str += " " * padding
+    if max(time_in, max_time) >= 60:
+        # If the longest test took an hour (or more), or the longest test took
+        # 10 minutes or more, pad the string to take this into account.
+        padding = 7 if (max_time >= 3600) or ((max_time / 60) % 60 >= 10) else 6
```
LarsAsplund, line 44: `max_time / 60 >= 10` should be sufficient, right?

vunit/test/report.py
```diff
@@ -18,6 +18,43 @@
...
+        if minutes > 0:
+            time_str += f"{int(minutes)} min ".rjust(padding)
+        else:
+            time_str += " " * padding
+
+    # If the longest test took a minute (or more), or the longest test
+    # took 10 seconds or more, pad the string to take this into account.
+    padding = 6 if (max_time >= 60) or (max_time % 60 >= 10) else 5
```
LarsAsplund, line 52: `max_time >= 10`

docs/release_notes/4.8.0.rst
```diff
@@ -0,0 +1,5 @@
+Features
```
LarsAsplund, line 1: Instead of creating a release note you should add a news fragment like this https://github.com/VUnit/vunit/blob/master/docs/news.d/776.feature.rst

The number is the number of this PR and I would call this a feature, i.e. `1025.feature.rst`. The release notes will be automatically generated from the news fragments when we build the next release.

LarsAsplund (commented):
I think the output looks good now. Just some minor coding comments and instructions for creating a news fragment

vunit/test/report.py
tossegus: Nifty! I hadn't seen (nor looked for) that division construct before! I'll use that construction instead.

LarsAsplund:
Thank you for your contribution!

tasgomes:
Sorry to comment on this closed thread. Using the proper time units in the summary report (at the bottom) is a very nice addition. But, is it just me or using the same format to report the time of each test case is now kind of ugly?

The extra spacing in between parenthesis seems really off:
image

Would removing the extra spacing hinder the possibility to spot quickly the slow test cases?

Wouldn't a format like:

```
00:11:11.1
02:00:07.3
01:51:43.3
```

be more suitable?

There wouldn't be unnecessary and odd spacing in between the parenthesis, and the size of the string would be consistent no matter the time each test case takes.

LarsAsplund:
We had a number of discussions regarding the format of the time, see https://github.com/VUnit/vunit/pull/1025#issuecomment-2143612233. However, we didn't really discuss the parenthesis. They serve to emphasize the time column from the others but there are of course several ways of doing that. Ideas? Remove the parenthesis? Remove the internal spacing? Use a separate color? I only have one other example in front of me: pytest. It can be configured to anything but the default is a mix of several things:

image

tasgomes:
I also don't know what looks the best. For sure having the total time of the simulation with a more suitable time unit is good thing. But for the time each test case takes, I personally liked the old way.

> Remove the parenthesis? Remove the internal spacing? Use a separate color?

Another way of solving this would be to use a format that takes the same number of characters no matter the execution time, like `(hh:mm:ss)`.

tossegus:
> I also don't know what looks the best. For sure having the total time of the simulation with a more suitable time unit is good thing. But for the time each test case takes, I personally liked the old way.
> 
> > Remove the parenthesis? Remove the internal spacing? Use a separate color?
> 
> Another way of solving this would be to use a format that takes the same number of characters no matter the execution time, like `(hh:mm:ss)`.

I agree to some degree regarding this. 

However, that would remove the possibility of quickly identifying what tests take the most/much time.
If you want to remove the spaces that aren't needed, you would have to write 00:00:01 for a test taking 1 second, which I'd say is harder to grasp, and looks strange to me.

The question is if an optional time-format flag would be a possible option? In that case the returned time_str could be (I haven't thought it through very thoroughly yet) formatted according to that flag?

LarsAsplund:
It will probably be hard to reach full consensus in this matter since it is rather subjective. I tried to find some research in this area and while it is hard to find definite answers I felt that the time format we changed to was preferred over the alternatives.

A common solution to this is to allow for a configurable time format. If you look at `pytest -h` you will find "millions" of options related to formatting. I don't want to go down that route because it makes the help very bloated.

We had similar discussion for logging from VHDL where we can have a mix of logging styles from various frameworks. It led to the common log interface which allow for transformation of VUnit log streams into whatever format you like, see https://vunit.github.io/logging/user_guide.html#external-logging-framework-integration.

I would be more interested in something similar for Python logging but that would be a new issue.

tasgomes:
> Remove the parenthesis?

I think for now I would just go with this quick fix and keep the same time format as is.

Then, like you say to make this more configurable if there is a need for that then this could be done as another issue.

PR #1039, opened 2024-07-11
Allow automatically determining the number of threads

oscargus:
Pass `-p0` to use all cores (determined by `multiprocessing.cpu_count()`.

### DIFF
--- a/docs/news.d/1039.feature.rst
+++ b/docs/news.d/1039.feature.rst
@@ -1,0 +1,1 @@
+It is possible to use all logical CPUs by passing `-p0` (`--num-threads=0`). See `multiprocessing.cpu_count` for how the number is determined.
--- a/docs/run/src/run.py
+++ b/docs/run/src/run.py
@@ -9,6 +9,7 @@
 from pathlib import Path
 from vunit import VUnit, VUnitCLI
 from io import StringIO
+from multiprocessing import cpu_count
 import sys
 import re
 from tools.doc_support import highlight_code, highlight_log, LogRegistry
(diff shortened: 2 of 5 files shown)


LarsAsplund:
Sounds like a good idea but I would like to it to take hyper threading into account. I typically run 12 threads on my 6 cores. From what I understand there is also `os.cpu_count` that might work differently. I'm on my phone so I can't test now. I've also seen threads indicating it doesn't always work. Differences between Windows and Linux.

oscargus:
I am quite sure that it does that. I've tested a few computers to confirm.

On my laptop (i7 1260P, 4P + 8E): `cpu_count() = 16`
On my workstation (some Xeon, 6 cores, 12 threads): `cpu_count() = 12`
On my compute server (Two 20-core Xeon, 80 threads): `cpu_count() = 80`

I was a bit unsure what to write in the note as the concept of threads from a software perspective is a bit different from the hardware perspective... Maybe one should write "logical CPUs"? (`os.cpu_count()` returns the same btw.)

LarsAsplund:
Ok, then it should be fine. Regarding the use of "threads". I think we should keep it when explaining the `-p` option since the verbose option name is `--num-threads`.

I will look into the PR a bit more when I'm back at my computer but I can't see any major issues with it. I'm not sure how common it is that `cpu_count` fails to determine the number of threads but I think we should test for that case and raise a descriptive error if it happens.

LarsAsplund:
I tested this locally and it works as expected.

I take back my comment on how the `-p` option is described. I'm good with how it is stated and will merge this PR.

oscargus:
> I take back my comment on how the `-p` option is described. I'm good with how it is stated and will merge this PR.

I reformulated it based on your comment, so glad it improved it. :-)

PR #1040, opened 2024-07-11
Use Paths rather than strings where possible

oscargus:
Avoids quite a bit of conversions back and forth.

Also add some typing to make it easier to remember...


oscargus:
The `!s` specifier is not required (unless a special case where one would like to override specifiers in the string). There are 200 or so remaining locations, but I changed where I saw them.

oscargus:
Probably should do this in a more systematic way later, so closing.

PR #1041, opened 2024-07-11
Add SPI verification components

n8tlarsen:
Addresses the SPI subset of #425. #611 seemed stale and claimed to be buggy, so I wrote my own. I've used these components successfully for several designs and figured it would be time to contribute them to the VUnit project.

### DIFF
--- a/vunit/vhdl/verification_components/src/spi_master.vhd
+++ b/vunit/vhdl/verification_components/src/spi_master.vhd
@@ -1,0 +1,115 @@
+library ieee;
+use ieee.std_logic_1164.all;
+
+use work.com_pkg.net;
+use work.com_pkg.receive;
+use work.com_pkg.reply;
+use work.com_types_pkg.all;
+use work.stream_master_pkg.all;
+use work.stream_slave_pkg.all;
+use work.sync_pkg.all;
+use work.spi_pkg.all;
+use work.queue_pkg.all;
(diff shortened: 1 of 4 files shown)


LarsAsplund:
@n8tlarsen Thanks for your contribution, much appreciated! I will get back to this when I have published the VUnit VC standard as mentioned in https://github.com/VUnit/vunit/pull/1017#issuecomment-2186861432. I would like this VC to comply before it gets merged.

n8tlarsen:
Looks like the checks for licensing are failing. I reviewed the contribution guidelines and found the license template, but I am curious why the licensing must be attributed to Lars Asplund. Does the MPL not provide for a redistribution of modified source files?

LarsAsplund:
When we decided on what license to use, we recognized that there were no well-known licenses targeting our domain. Mozilla was selected because its wording was less SW-ish than the others. The single copyright holder was added such that the license can be changed by me alone should a better open source license emerge.

Once we have established the VUnit VC standard, we intend to place VCs in separate repositories. These can live under the VUnit domain on GitHub or elsewhere, they can have their own release cycles, be maintained by others, and have other copyright holders and compatible open-source licenses. Third-party VCs tend to be less trusted than those provided by the main project which is another reason for establishing a standard that comes with some level of quality guarantees.

PR #1055, opened 2024-09-04
Fix syntax error in id_pkg

diedricm:
This PR fixes a bug that resulted in failure during elaboration using riviera pro. I am not sure if the range (0 to 1) is desired or if (1 downto 0) would be more correct. It works like this now though.

### DIFF
--- a/vunit/vhdl/data_types/src/id_pkg.vhd
+++ b/vunit/vhdl/data_types/src/id_pkg.vhd
@@ -22,7 +22,7 @@ package id_pkg is
   end record;
   constant null_id : id_t := (p_data => null_ptr);
   type id_vec_t is array (integer range <>) of id_t;
-  constant null_id_vec : id_vec_t := (1 to 0 => null_id);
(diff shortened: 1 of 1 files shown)


LarsAsplund:
Hi @diedricm,

Vector/arrays with null ranges is a concept used by VUnit to define "no information available", similar to `None` in Python for example. They are typically used as default value for subprogram parameters where the subprogram needs to identify cases where the user didn't provide any value. So the range is intentional. Do you really get an error from Riviera-PRO or just a warning? Can you provide what it says + tool version?

Null ranges are perfectly legal according to the VHDL standard:

_A range specifies a subset of values of a scalar type. A range is said to be a null range if the specified subset
is empty.

The range L to R is called an ascending range; if L > R, then the range is a null range. The range L downto
R is called a descending range; if L < R, then the range is a null range._

diedricm:
Hi @LarsAsplund,
Thank you for bringing that up. I totally forgot about zero ranges. The bug existed with riviera pro 2022.10. I updated to riviera pro 2023.10 and the issue is gone now. My bad.

Issue #1026, opened 2024-06-05
No warning of erroneous architecture when adding configurations


sigmuha:
When adding a generic to a testbench with the "add_config" method, no warning is given to the user if the architecture of the testbench is not correct. 

I experienced this when writing "t_dummy" instead of "tb_dummy" for the entity of the architecture of a testbench, where the entity is called "tb_dummy".  This caused vunit not to create a default configuration, leading to the following error when "add_config" was called:
```
File "C:\Users\dummy\AppData\Roaming\Python\Python312\site-packages\vunit\configuration.py", line 265, in add_config
    config = configs[DEFAULT_NAME].copy()
             ~~~~~~~^^^^^^^^^^^^^^
KeyError: None
```
What caused this error was not obvious at all and took some time to find what caused it. Spelling the architecture correctly fixed the issue. To remedy this, the shown error could be ignored, and continue with the compilation of all design and testbench files, which would highlight the misspelling. 

It is possible that different kinds of errors in the testbench file also can cause this, and not just a misspelling of the entity in the architecture declaration.

LarsAsplund:
Can you provide an [MRE](https://en.wikipedia.org/wiki/Minimal_reproducible_example). I can't seem to reproduce this so maybe I misunderstood your code.

sigmuha:
Sorry for the late reply.

[vunit_issue.tar.gz](https://github.com/user-attachments/files/15957588/vunit_issue.tar.gz) contains two files, run.py and test.vhd. 

Doing ```python run.py``` the same error as written above is reproduced. This happens, because a generic is attempted to be added, when no architecture for the testbench in test.vhd exists. In this case the architecture is misspelled. The user is not notified of this error.

LarsAsplund:
This should be fixed now.

PR #1062, opened 2024-09-25
Collection of bug fixes for a potential v4.7.1

markuspg:
Hi,

this merge request collects fixes from current `master` for a potential `v4.7.1`. I'm using this locally for quite a while and thought maybe it might be useful for someone else too. I went through all commits on `master` since `v4.7.0` and cherry-picked anything looking like a fix (but without breaking behaviour).

The only breaking fix I cherry-picked is [sim_if: continue after compilation failure if '--keep-compiling' is True](https://github.com/VUnit/vunit/commit/7213390e201dc87f62d1e7c282eeff39ac2d9822).


LarsAsplund:
I'm a bit hesitant to make an official 4.7.1 with the purpose of avoiding backward breaking features. I don't want to create an expectation that there will be a long-lived 4.x branch that will be maintained in parallel with 5.x. I rather see that people starts using the pre-releases for version 5.0.0 (https://vunit.github.io/installing.html#using-the-python-package-manager) in order to get access to the latest updates. There will be some breaking changes to address but for most project that is not a major effort that needs to be postponed.

nathanaelhuffman:
To flip this around a bit, it seems like there's appetite for picking up quite a few bugfixes given the length of time since the last stable release. Is there consideration toward making more frequent bugfix releases going forward in the 5.x branches? It would also seem to me that there's no implication that a release now locks the project into supporting more backports to the 4.x branches long-term.

That said, as a long time user I hadn't realized some pre-release versions are available via pip, so that solves my immediate need for a more recent release without making pip point to main, and thanks for referencing it!

markuspg:
@LarsAsplund I perfectly understand your reasoning. At least speaking for me I wouldn't expect a long-lived 4.x branch. My thought was that in the last one and a half years "all" major bugs in 4.7.0 should be found and fixed on 5.0.0. So 4.7.1 would just end out this branch. For my private programming I love riding bleeding edge, for professional projects I prefer tried and tested code.

I'm not familiar with the codebase, but all fixes seemed trivial to me. Maybe the fix for #742 could be dropped, since this is a breaking one. I also didn't cherry-pick the fix for #1026, because I found it difficult to judge the consequences and the issue seemed quite specific to me (contrary to e.g. the memory leak, which is bothering me ;)).

Maybe I'll give 5.0.0 a shot then too. I started this whole endeavour because I'm hunting down a memory leak - that's why I was on the lookout for a bullet proof VUnit release. Yesterday (using the code from this merge request) I found out that the issue is with GHDL actually (anyone else having issues there?).

Anyhow, thank you for your work on VUnit!

LarsAsplund:
It feels to me that the "impression of a version number" is creating problems for us:

1. If we spread several breaking updates over several major releases, we are giving the impression of an "immature project". In reality, the number of breaking changes are the same. The breaking changes we have prepared for v5.0.0 are typically small and insignificant to most users or larger but announced for a long time such that users had time to prepare.
2. Pre-releases are considered experimental. In reality, there is no major difference between commits on the main branch between releases and those with a release tag on them. Development is on feature branches and new features are tested before they are merged.
3. Frequent (feature) releases gives the impression of an active project so some projects do them at fixed time intervals. In reality, it is the pace of commits and the significance of those commits that matters.

Number 1 is the reason for waiting with v5.0.0, but I feel number 2 is a problem as well.

What do you think would be the best approach going forward? My goal is that users incorporate new commits into their projects as they appear in the main branch.  What are the obstacles preventing that?

markuspg:
Now this is becoming complex 😉️ Generally I'm fine with your release schedule and don't want to create additional effort.

My original problem was that I was hunting down a memory leak which I was not sure if it is in VUnit or GHDL. For that reason I wanted an as "bullet-proof" release as possible. With the bugfixes since `v4.7` on `master` it was sure that `v4.7.0` was not it. Going through the changes there don't seem to be really "dangerous" ones for `5.0`, but still there is the risk of new issues popping up. That's why I came up with `v4.7.1`.

Regarding `3)` I like your taking its time release cycle. I'm rather annoyed by project's that bump their version numbers just to seem hip and relevant. The only issue is that there are a few bugfixes the people didn't want to wait for a new major release.

In an ideal world without a lack of time I'd wish for every project to make a final release on the `x-1` branch with all the bugfixes that accumulated until the release of `x`, when `x` is being released.

Setting all this theory aside. I'm still on the hunt for the memory leak, but it does not seem to be in VUnit. For the last few days I went with `v5.0.0-dev.3` and everything worked fine.

So I guess you don't need to "waste" your time with a `v4.7.1` release. Thank you for taking time for the discussion though and your work on VUnit!

LarsAsplund:
@markuspg Thanks for your feedback. Second om my priority list right now is a new verification component standard which means breaking changes on its own but also that verification components are placed in separate repo(s). After that I will review if we're ready for v5.

PR #1070, opened 2024-10-26
Add APB verification components

c-thaler:
I added APB master and slave VCs. The APB master is based on a VC that I used in a few of my designs.

### DIFF
--- a/vunit/vhdl/verification_components/src/apb_master.vhd
+++ b/vunit/vhdl/verification_components/src/apb_master.vhd
@@ -1,0 +1,147 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this file,
+-- You can obtain one at http://mozilla.org/MPL/2.0/.
+--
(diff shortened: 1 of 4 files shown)


LarsAsplund:
Thanks @c-thaler, I will look into it! Please have a look at the VC standard being discussed #1017

c-thaler:
Thanks. I had a look at it and I will start with implementing Rule 13 (single generic).

c-thaler:
I've also implemented rules 10/15 (locking test runner cleanup gate while busy) and 14 (sync interface).

c-thaler:
@LarsAsplund The idea is to use `create_std_cfg` within the constructor of my VC, right? That also means that the logger parameter can be removed from my VC's constructor?

LarsAsplund:
I have two projects where this can be useful. I will have it tested there to see if there are issues to resolve.

LarsAsplund:
@c-thaler Unfortunately not. There are some good news though. VUnit will get funding from [Chips JU](https://www.chips-ju.europa.eu) which will start during the spring. One of our focus areas is verification components which means that there will be time available to address such issues.

PR #1082, opened 2024-12-12
Fix array_axis_vcs example 

### LINKED ISSUE #1081: How to Export Waveforms with QuestaSim/ModelSim?
Hi there!

I’m wondering if any of you have managed to export a waveform to the `vunit_output/..`. path when simulating with QuestaSim/ModelSim.

I assume it would involve adding something like this to `run.py`:

```pyhton
vu.set_sim_option("modelsim.vsim_flags", ["-wlf wave.wlf"])
``` 

However, this option isn’t working. The simulation runs fine, but it doesn’t save any waveform. I checked this by running:

```bash
user@machine:~/.../vunit_out$ find -type f -name *.wlf
```

The Questa User Guide states the following:

issue_vunit

But I’m not sure how VUnit handles Questa simulation outputs.

For example, when running simulations with GHDL, you can use the argument `--gtkwave-fmt vcd` to export the waveform in `.vcd` format to a path accessible through:

```pyhton
str(Path(item.path) / "ghdl" / "wave.vcd"
``` 

My goal is to achieve something similar in order to upload the waveform as an artifact in CI.

Does anyone have any tips or insights on how to resolve this issue? :thinking:

Cheers!

--- 

 /cc @umarcor

Unike267:
Hi there!

Yesterday, I spoke to Unai M. C., and he gave me a hint to resolve #1081 by reviewing the `array_axis_vcs` example. I’ve fixed it to achieve correct simulation with QuestaSim.

### Changes made:

- Replaced `vu` with `VU` on line 32.
- Corrected the instantiation of components in `runall_addwave.do`.

The following images illustrate the errors:

**Error 1**:

vunit_PR_1

**Error 2**:

vunit_PR_2

**Fixed** (with -g, the Questa GUI opens; without -g and with -v, the test information (through `info()` function) is displayed correctly):

vunit_PR_3

vunit_PR_4

Cheers!

---

/cc @umarcor

### DIFF
--- a/examples/vhdl/array_axis_vcs/runall_addwave.do
+++ b/examples/vhdl/array_axis_vcs/runall_addwave.do
@@ -1,5 +1,5 @@
 echo "Running DO File"
-foreach v "vunit_axism vunit_axiss uut uut/fifo" {
+foreach v "uut_vc/vunit_axism uut_vc/vunit_axiss uut_vc/uut uut_vc/uut/fifo" {
   add wave -group "$v" -position insertpoint sim:/tb_axis_loop/$v/*
 }
 run -all;
--- a/examples/vhdl/array_axis_vcs/run.py
+++ b/examples/vhdl/array_axis_vcs/run.py
@@ -29,6 +29,6 @@ SRC_PATH = Path(__file__).parent / "src"
 
 VU.add_library("lib").add_source_files([SRC_PATH / "*.vhd", SRC_PATH / "**" / "*.vhd"])
 
-# VU.set_sim_option('modelsim.init_files.after_load',['runall_addwave.do'])
+VU.set_sim_option("modelsim.init_file.gui", "runall_addwave.do")
 
 VU.main()
--- a/examples/vhdl/array_axis_vcs/run.py
+++ b/examples/vhdl/array_axis_vcs/run.py
@@ -29,6 +29,6 @@ SRC_PATH = Path(__file__).parent / "src"
 
 VU.add_library("lib").add_source_files([SRC_PATH / "*.vhd", SRC_PATH / "**" / "*.vhd"])
 
-# vu.set_sim_option('modelsim.init_files.after_load',['runall_addwave.do'])
+# VU.set_sim_option('modelsim.init_files.after_load',['runall_addwave.do'])
 
 VU.main()


Unike267:
**Note:**

If you run this test bench with QuestaSim, you may need to comment out lines https://github.com/VUnit/vunit/blob/master/examples/vhdl/array_axis_vcs/src/fifo.vhd#L40-L48, as QuestaSim doesn't support PSL (even though PSL is supposed to be supported by any compiler that supports VHDL 2008).

examples/vhdl/array_axis_vcs/runall_addwave.do
```diff
@@ -1,5 +1,5 @@
 echo "Running DO File"
-foreach v "vunit_axism vunit_axiss uut uut/fifo" {
+foreach v "uut_vc/vunit_axism uut_vc/vunit_axiss uut_vc/uut uut_vc/uut/fifo" {
   add wave -group "$v" -position insertpoint sim:/tb_axis_loop/$v/*
 }
 run -all;
```
LarsAsplund, line 5: I would prefer if you remove this line such that the user has to run manually when in GUI mode. When this test is run without the `--gui` option, VUnit will pass the follow script lines to the simulation:

```
set failed [vunit_load]
if {$failed} {quit -code 1}
set failed [vunit_run]
if {$failed} {quit -code 1}
quit -code 0
```

The  `vunit_load` command will load the design and then source `runall_addwave.do`. That script runs the test once with `run -all` until it completes. Then, `vunit_run` will continue to run after the stopping point which will continue forever, just toggling the clock, and the call to `run.py` will never return,

Unike267: Hi @LarsAsplund!

As you mentioned, the content of the script `batch.do`, located at `vunit_out/test_output/lib_name.tb_name.hash/modelsim`, is generated with the following content:

```bash
onerror {quit -code 1}
source "/home/.../vunit/examples/vhdl/array_axis_vcs/vunit_out/test_output/lib.tb_axis_loop.test_260bb5c8c675e898eca5dc9024a4420ede12c0bc/modelsim/common.do"
set failed [vunit_load]
if {$failed} {quit -code 1}
set failed [vunit_run]
if {$failed} {quit -code 1}
quit -code 0
```

However, this is not only the case when you run it without the `-g` option. I’ve performed a couple of tests, and this script remains the same when running with the `-g ` option, with `VU.set_sim_option('modelsim.init_files.after_load', ['runall_addwave.do'])` commented out, or even in other examples I tested. This makes me think that the `batch.do` file is "always" generated with this content. :thinking:

Nevertheless, you are absolutely right, this simulation (even before my change) freezes when you run it without the `-g` option (after the test's verbose output), as shown in the last image of the main comment. In fact, I had to skip the simulation to stop it.

I solved this "freeze" issue by adding `quit -code 1;` at the end of the `runall_addwave.do script. The updated script is as follows:

``` bash
echo "Running DO File"
foreach v "uut_vc/vunit_axism uut_vc/vunit_axiss uut_vc/uut uut_vc/uut/fifo" {
  add wave -group "$v" -position insertpoint sim:/tb_axis_loop/$v/*
}
run -all;
quit -code 1;
``` 

I've verified that with the `-g` option, the Questa GUI opens with the waveform, and without this option, the simulation finishes correctly.

If you have a couple of minutes, could you verify the correct operation of this change? Once confirmed, I will add it to the PR.

Thanks for the feedback, Lars!

Cheers!

Unike267: Correct execution of the test (after adding `quit -code 1;`) without the `-g` option:

vunit_PR_5

LarsAsplund: Yes, `batch.do` is the same in both cases but `vunit_load` defined in `common.do` differs. In one case it calls `runall_addwave.do` but not in the other.

Wouldn't a run with the `-g` option exit immediately after the simulation ended if you add `quit` to the script?

Unike267: > Yes, batch.do is the same in both cases but vunit_load defined in common.do differs. In one case it calls runall_addwave.do but not in the other.

Oh, I see!

> Wouldn't a run with the -g option exit immediately after the simulation ended if you add quit to the script?

Oh, my bad! You are absolutely right. I initially checked that Questa opened, but I closed it immediately. Now, I’ve checked again, and if you leave it open for some time, the GUI does eventually close. :disappointed:

Sorry about that! I’ll look for another solution.

LarsAsplund: Since the DO file is only relevant to the GUI mode it would be better to use this flag instead:

```
VU.set_sim_option("modelsim.init_file.gui", "runall_addwave.do")
```

No need to comment out that line.

Unike267: 🫡

PR #962, opened 2023-10-19
Add Block Design support for simulation

SzymonHitachi:
Closes #963

Added block design parsing and generation based on https://gitlab.com/tsfpga/tsfpga/-/issues/74.

1) Requires creating `tcl` file through `write_bd_tcl` for open block designs
2) Resulting `tcl` should be located together with other `ip_cores`

PR #1086, opened 2024-12-18
Add Vivado Block Design support for simulation

SzymonHitachi:
Reopening https://github.com/VUnit/vunit/pull/962 after cleanup and alignment.

### DIFF
--- a/vunit/vivado/tcl/extract_compile_order.tcl
+++ b/vunit/vivado/tcl/extract_compile_order.tcl
@@ -4,12 +4,45 @@ set output_file [lindex $argv 1]
 open_project ${project_file}
 
 set file_out [open ${output_file} w]
+
+# Generate block designs, and any IPs contained by the block designs
+generate_target {simulation synthesis} [get_files *.bd]
+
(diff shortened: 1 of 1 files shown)


vunit/vivado/tcl/extract_compile_order.tcl
```diff
@@ -4,12 +4,45 @@ set output_file [lindex $argv 1]
 open_project ${project_file}
 
 set file_out [open ${output_file} w]
+
+# Generate block designs, and any IPs contained by the block designs
```
yahya-farhadi: Some of the AMD IPs used in Block design have SystemVerilog Files which is not considered in Librraries. Without this change, errors will be reported for some of the Block Designs.

vunit/vivado/vivado.py line 34
`is_verilog = file_name.endswith(".v") or file_name.endswith(".vp")`

Should be modified as below:

```
# SystemVerilog files should be handled by Verlig Compiler.
# It is required for some AMD IPs used in the block design whose compile order includes SystemVerilog sources.
is_verilog = file_name.endswith(".v") or file_name.endswith(".vp") or file_name.endswith(".sv")
```

PR #1091, opened 2025-01-13
Drop support for Python 3.7.

LarsAsplund:
This version reached end-of-life mid-2023.


LarsAsplund:
Failing tests have other causes and will be fixed in PR #1089 and #1092

PR #1092, opened 2025-01-13
Update setup GHDL action

LarsAsplund:
Bump to the latest setup GHDL action


LarsAsplund:
Failing tests will be fixed in #1089

PR #1095, opened 2025-01-26
Added support for running testbenches depending on a set of user-prov…

LarsAsplund:
…ided source files.

The intended use case is to feed VUnit with files that has been changed since some point in time. The changes are typically provided by the version control system.

The API is:

```
    def run_dependent(self, source_files: List[Union[str, Path]]) -> None:
        """
        Run testbenches depending on the provided source files.

        Test patterns on the command line will add to the depending testbenches.

        :param source_files: List of :class:`str` or :class:`pathlib.Path` items,
                             each representing the relative or absolute path to
                             the source file.
        :returns: None
        """
```

### DIFF
--- a/vunit/ui/__init__.py
+++ b/vunit/ui/__init__.py
@@ -16,9 +16,10 @@ import traceback
 import logging
 import json
 import os
-from typing import Optional, Set, Union
+from typing import Optional, Set, Union, List
 from pathlib import Path
 from fnmatch import fnmatch
+from glob import glob
 
 from ..database import PickledDataBase, DataBase
 from .. import ostools
@@ -111,6 +112,22 @@ class VUnit(object):  # pylint: disable=too-many-instance-attributes, too-many-p
         """
         return cls(args, vhdl_standard=vhdl_standard)
 
+    @staticmethod
+    def _make_test_filter(args, test_patterns):
+        "Create test filter function from test patterns."
+
+        def test_filter(name, attribute_names):
+            keep = any(fnmatch(name, pattern) for pattern in test_patterns)
+
+            if args.with_attributes is not None:
+                keep = keep and set(args.with_attributes).issubset(attribute_names)
+
+            if args.without_attributes is not None:
+                keep = keep and set(args.without_attributes).isdisjoint(attribute_names)
+            return keep
+
+        return test_filter
+
     def __init__(
         self,
         args,
@@ -125,17 +142,7 @@ class VUnit(object):  # pylint: disable=too-many-instance-attributes, too-many-p
         else:
             self._printer = COLOR_PRINTER
 
-        def test_filter(name, attribute_names):
-            keep = any(fnmatch(name, pattern) for pattern in args.test_patterns)
-
-            if args.with_attributes is not None:
-                keep = keep and set(args.with_attributes).issubset(attribute_names)
-
-            if args.without_attributes is not None:
-                keep = keep and set(args.without_attributes).isdisjoint(attribute_names)
-            return keep
-
-        self._test_filter = test_filter
+        self._test_filter = self._make_test_filter(args, args.test_patterns)
         self._vhdl_standard: VHDLStandard = select_vhdl_standard(vhdl_standard)
 
         self._preprocessors = []  # type: ignore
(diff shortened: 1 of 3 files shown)


LarsAsplund:
@LukasVik Is this compatible with the similar function in TSFPGA?

LukasVik:
Yes this would simplify it a lot. Good initiative!

Can a testbench file be passed as argument also? Or is it limited to
“source”/“implementation” files?

Does it work when location/check preprocessing is enabled also? From e.g. a
git diff you would have the original file location, not the preprocessed
file location. Will that work as expected?

So if I understand it correctly, one would call this new “run_dependent()”
method and it adds to the test filter. But one would still call “main()”
after to actually execute the tests? If so I think the verbiage is a little
misleading. Perhaps “add_dependents_to_test_filter”?


Thanks!

LukasVik:
Sounds great!

> Note that there are important changes that VUnit can't detect directly. Let's say you have no changes to your HDL files but there's a stimuli file to `tb_x` that changed. In that case `tb_x` should be simulated. The way to deal with that is to process the changes reported by your version control system. If the stimuli file has changed, `tb_x` is added to the list of files.

Yes makes sense. In the tsfpga class (https://github.com/tsfpga/tsfpga/blob/main/tsfpga/git_simulation_subset.py) we detect changes in the `regs_*.toml/json/yaml` and trigger based on that. There are probably other cases like that but I agree that all that needs to be handled outside of VUnit.

> Preprocessing works. VUnit keeps track of both the original file name and the preprocessed file name and the user-provided files are compared with the original.

Great, a lot of the complexity in the tsfpga class is the handling of original vs processed path. Nice to get rid of that.

> Your understanding of `run_dependent` is correct. Agree that the name is a bit misleading and should be changed. I'm thinking about `include_dependent_in_test_pattern` because it makes it easier to also add a `exclude_dependent_from_test_pattern`.

Cool. I agree with the naming.

vunit/ui/__init__.py
```diff
@@ -1032,6 +1083,19 @@ def add_json4vhdl(self):
         """
         self._builtins.add("json4vhdl")
 
+    def run_dependent(self, source_files: List[Union[str, Path]]) -> None:
```
LukasVik, line 1086: I think it's best if the type hint is `Iterable[...]` instead of list. That way you can pass a set (or tuple, or any iterable class) also. In the case of tsfpga I will pass a set to this function.

Using Iterable is recommended by mypy but I can't find the right page right now. There's some info here: https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html#standard-duck-types

vunit/ui/__init__.py
```diff
@@ -752,6 +762,47 @@ def _main(self, post_run):
...
         return all_ok
 
+    def _update_test_filter(self, dependency_files):
+        """
+        Update test filter to match testbenches depending on user_files
+        """
+        # Find source file objects corresponding to the provided files.
+        # Non-existing files, directories and files not added to the
+        # project are ignored.
+        dependency_paths = []
```
LukasVik, line 772: I think it would be more efficient to use a `set()`. There will be no duplicates in `dependency_paths` even if the user passes duplicates, and the `if a in b` check below is much faster for set than list.

vunit/ui/__init__.py
```diff
@@ -752,6 +762,47 @@ def _main(self, post_run):
...
+        for dependency_file in dependency_files:
+            if isinstance(dependency_file, str):
+                dependency_file = Path(dependency_file)
+            dependency_file = dependency_file.resolve()
+            if dependency_file.exists() and not dependency_file.is_dir():
```
LukasVik, line 777: I think this is clearer, more efficient and more compact:

```suggestion
            if dependency_file.is_file():
```

LukasVik: It includes the check that the path exists https://docs.python.org/3/library/pathlib.html#pathlib.Path.is_file

LukasVik (commented):
Looks overall good to me, it's a good addtion. Some minor comments.

vunit/ui/__init__.py
LarsAsplund: Agree, I usually talk about lists because it makes it more understandable for those not very experienced with Python. However, there is not reason for why we should limit mypy that way when the code works with iterables.

PR #1101, opened 2025-03-05
Add --waves argument to generate waveforms for GHDL and NVC

### LINKED ISSUE #1099: nvc: Waveform files are only generated with --gui option
It seems that when using nvc as simulator, the waveforms set with `--viewer-fmt` format are only generated when using the `--gui` option of the Python runner. With ghdl, this isn't the case, waveforms are generated regardless of the `--gui` option.

I think, the problem is in nvc.py `simulate()` method:

First, `wave_file` is only set to a value if `_gui` is set:
https://github.com/VUnit/vunit/blob/acf7e7f294f9b1b11dc664c2f145b9cb95906532/vunit/sim_if/nvc.py#L260-L265

Later if `wave_file` is set, the nvc `--wave` is added to the command line:
https://github.com/VUnit/vunit/blob/acf7e7f294f9b1b11dc664c2f145b9cb95906532/vunit/sim_if/nvc.py#L295-L296

In ghdl.py, creation of the waveform export command line option is only dependent on `_viewer_fmt`, but not on `_gui`:

https://github.com/VUnit/vunit/blob/acf7e7f294f9b1b11dc664c2f145b9cb95906532/vunit/sim_if/ghdl.py#L368-L375

https://github.com/VUnit/vunit/blob/acf7e7f294f9b1b11dc664c2f145b9cb95906532/vunit/sim_if/ghdl.py#L327-L333

oscargus:
Closes #1099 
Closes #1042 

The approach here is to pass `--waves` if a waveform file should be generated. That is clearer than the "passing `--gtkwave-fmt` "-approach earlier used.

There is a deprecation warning when passing only `--viewer-fmt`/`--gtkwave-fmt` for GHDL, but it still generates the file.

I also added a new alias `--waves-fmt` for consistency.

(I guess that one can also support other simulators generating waveform files. At least for ModelSim/Questa it is possible to generate VCD-files with a few .do-file commands and I guess the same holds for many other simulators.)

### DIFF
--- a/docs/news.d/1101.feature.rst
+++ b/docs/news.d/1101.feature.rst
@@ -1,0 +1,5 @@
+[GHDL/NVC] It is now possible to pass ``--wave`` to generate waveform files without
+opening the GUI. The earlier GHDL-only approach of passing ``--viewer-fmt``
+(or ``--gtkwave-fmt``) is deprecated and will be removed in a later version.
+There is also a new alias for providing the waveform format, ``--wave-fmt``,
+to make it consistent with the new flag.
--- a/vunit/sim_if/ghdl.py
+++ b/vunit/sim_if/ghdl.py
@@ -59,6 +59,7 @@ class GHDLInterface(SimulatorInterface, ViewerMixin):  # pylint: disable=too-man
         group = parser.add_argument_group("ghdl/nvc", description="GHDL/NVC specific flags")
         group.add_argument(
             "--viewer-fmt",
+            "--wave-fmt",
             "--gtkwave-fmt",
             choices=["vcd", "fst", "ghw"],
             default=None,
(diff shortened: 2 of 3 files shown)


oscargus:
I've updated to `--wave` and `--wave-fmt`.

You do not have to specify `--wave-fmt`. The thing is that it is not enough to *only* pass `--wave-fmt` to get a wave form file. You will also have to pass `--wave`. That is something one can think about, but it is at least consistent with the `g` flag. (The defaults are the same since `--wave-fmt` is just an alias for the others so one can actually pass `--wave --gtkwave-fmt=fst`.)

creiter64:
@LarsAsplund Any chance to get this into the next release candidate?

oscargus:
As this was brought up again, I've been thinking a bit about `--wave` vs `--wave-fmt`.

# Options
## `--wave` and `--wave-fmt=fmt`
Pro: no need to change the code
Pro: consistent with `--gui`
Con: you need to type two things to change the format

## `--wave[=fmt]`
Pro: less to type, only a single argument
Con: inconsistent with `--gui`
Con: not obvious how to handle `--wave=vcd --viewer-fmt=fst` 

(For the final point, it is not really obvious what happens when you call `--gui --wave` now anyway, or at least I do not recall it...)

LarsAsplund:
@creiter64 Yes, I expect that it will be part of the next release.

PR #1102, opened 2025-03-06
Fix warning "(vcom-1246) Range 1 to 0 is null" for Modelsim/Questa.

tasgomes:


tasgomes:
I would say that yes it can. At least for me, I was able to solve this kind of warnings for Modelsim and Questa  using this approach.

nselvara:
Thx for the reply. I think then the feature is almost not needed really. Those warnings are for real very annoying...

tasgomes:
I have to disagree on the "almost not needed really", specially if we can fix a warning with very low effort like this one. I would rather have a compilation with zero warnings that having avoidable ones that could "hide" important warnings that shall be addressed.

nselvara:
Yeah, I agree with you - removing those useless warnings improves the debugging experience.
Btw, I actually meant the null vector as a useless feature. However, I'm not sure if yours could be used in every (corner) case.

LarsAsplund:
@tasgomes Thank you for your contribution!

PR #1118, opened 2025-05-30
Added .svp as systemverilog protected source file

dejfson:
.SVP is used for example on QuestaSim/Modelsim to provide encrypted sources as a part of simulation. User can compile them using standard vcom, and they become part of given library.


dejfson:
Duplicate of #1115, I did not notice
.d

Issue #1105, opened 2025-03-12
ghdl compiled with llvm-jit support


csantosb:
When `ghdl` is compiled with the flag `--with-llvm-jit`, the output of `ghdl --version` gives 

    GHDL 5.0.1 (tarball) [Dunoon edition]
     Compiled with GNAT Version: 14.2.0
     static elaboration, LLVM JIT code generator
    Written by Tristan Gingold.

This produces an error in [this line](https://github.com/VUnit/vunit/blob/master/vunit/sim_if/ghdl.py#L152)

    AssertionError: No known GHDL back-end could be detected from running 'ghdl --version'

as the new jit pattern is not recognized.

PR #1121, opened 2025-06-11
fixes error when ghdl is compiled with jit support

csantosb:
See #1105


csantosb:
Tested using testing suite of [open-logic](https://github.com/open-logic/open-logic/tree/main/sim) repository, and up to date ghdl.

PR #1122, opened 2025-06-18
Add option to run testcase pre_config only

robindust-ce:
In my development flow I sometimes need to execute the pre_config of select testcases without actually running them. One example would be regenerating a .mif/.mem file that is loaded by a BRAM without closing my Modelsim GUI. I did not find any existing solutions for this (i.e. vunit_restart will not re-run the testcase pre_config).

As an example I implemented a naive version via a new command line parameter "--pre-config-only" that works for my use case.

Is there any interest in such a feature? Or perhaps I missed a feature which already covers my use-case?

### DIFF
--- a/vunit/ui/__init__.py
+++ b/vunit/ui/__init__.py
@@ -756,6 +756,9 @@ other preprocessors. Lowest value first. The order between preprocessors with th
         """
         Base vunit main function without performing exit
         """
+        if self._args.pre_config_only:
+            return self._main_pre_config_only()
+
         if self._include_in_test_pattern or self._exclude_from_test_pattern:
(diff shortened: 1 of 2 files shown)


tasgomes:
@robindust-ce FYI you can also add custom arguments to VUnit without changing the VUnit source code, in case this is interesting for you to achieve the same functionality.

https://vunit.github.io/py/ui.html#adding-custom-command-line-arguments

LarsAsplund:
I think this use case is a bit too special to break the clean structure we have today. I would rather go with a solution where the functionality of the `pre_config` can be executed separately. Depending on the simulator(s) used you may also call that function from a TCL script that is sourced before the simulation starts (including restarts). See `modelsim.init_files.before_run` (http://vunit.github.io/py/opts.html#simulation-options) for an example

PR #1130, opened 2025-08-12
stop packaging tests folder

glencoe:
Packaging the tests folder occupied a very common
namespace for all clients that would install
vunit.


LarsAsplund:
@glencoe Thank you for contributing!

PR #1133, opened 2025-09-03
#1132 Added warning when attempting to set NVC `--exit-severity` via `nvc.sim_flags`

### LINKED ISSUE #1132: [FEATURE REQUEST] Produce a warning when attempting to set NVC `--exit-severity` via `nvc.sim_flags`
I recently attempted to set the [exit severity](https://www.nickg.me.uk/nvc/manual.html#exit-severity) of an NVC simulation using the `nvc.sim_flags` sim option, but it wasn't changing the exit severity of the simulation. It took me a while to figure out why this was the case, and eventually I realised that this setting is overwritten by VUnit's `vhdl_assert_stop_level` option. It would be good to add a warning in this instance to help future users figure out the problem more easily.

JHertz5:
Resolves #1132.

I'm open to any feedback or changes required to get this merged.

### DIFF
--- a/vunit/sim_if/nvc.py
+++ b/vunit/sim_if/nvc.py
@@ -286,7 +286,17 @@ class NVCInterface(SimulatorInterface, ViewerMixin):  # pylint: disable=too-many
             if self._supports_jit:
                 cmd += ["--jit"]
             cmd += ["-r"]
-            cmd += config.sim_options.get("nvc.sim_flags", [])
+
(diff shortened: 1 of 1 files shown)


oscargus:
Makes sense to me (but I do not have any say). One may consider to change the warning message a bit to more explicitly say "use config.vhdl_assert_stop_level instead" (which I guess is the easiest way to obtain what you want, right?).

LarsAsplund:
@JHertz5 Thank you for contributing!

Issue #1158, opened 2026-02-05
Is "setuptools_scm >= 2.0.0, <3" still required for the buildsystem?


timkoers:
I'm having troubles installing vunit, since setuptools_scm >= 2.0.0 but smaller than 3 is required to install VUnit.
I've been working around this by providing the --no-build-isolation flag to pip when installing, but lately that has been causing some issues too.

setuptools_scm is currently at around version 9.x.x, so if you could re-evaluate if this dependency is even necessary that'd be great

Trooper-X:
> Somehow our CI stopped being able to install VUnit in the last like 3 days. Is this related? Is there a workaround?

Have you tried the `--no-build-isolation` like `timkoers` suggested?
That works for me.

watsonjj:
I believe this is connected to a new release of setuptools ([v82.0.0](https://setuptools.pypa.io/en/stable/history.html#v82-0-0)), which deprecated pkg-resources, and which CIs are likely pulling today. The problem can be bypassed by setting an upper limit on setuptools version in the build-system of vunit's pyproject.toml:

```
[build-system]
requires = [
    "setuptools >= 35.0.2, <82.0.0",
    "setuptools_scm >= 2.0.0, <3"
]
```

felixn:
I'm also affected, same error as Trooper-X

The issue is caused by the upper limit on setuptools_scm <3, while setuptools is unconstrained.
setuptools v82 deprecated pkg_resources
Removing the upper limit on setuptools_scm fixes the issue.

VUnit builds fine with the latest versions of both setuptools and setuptools_scm:
https://github.com/VUnit/vunit/pull/1159

I think a v4.7.1 hotfix would be best!

Markus-Wagner:
I'd also like to support a v4.7.1, as it is the latest released version. Assuming this is the way to go, I can think of 2 different approaches:

1. Fix 4.7.0 and 5.0.0 the same way

```toml
[build-system]
requires = [
    "setuptools >= 35.0.2",
    "setuptools_scm >= 2.0.0"    # maybe increase the version to where it does not require pkg_resources any more
]
```
This would be more future proof, but I don't know if it breaks something else in some elses dependencies

2. Different fixes for 4.7.0 and 5.0.0

Remove the upper limit for v5.0.0 (as used in the example above). I guess we do not need to take care regarding compatibility with other outdated python packages, which require old setuptools_scm versions

[...]

PR #1160, opened 2026-02-09
Restrict setuptools to pre-82.

Trooper-X:
Restrict setup tools to pre-82, as the deprecated pkg_resources is still used.

Issue: #1158


Trooper-X:
As discussed in #1158, for the next release a different strategy is choosen.
That makes this PR obsolete, so closed.

PR #1162, opened 2026-02-10
update build system

nbdy:
Fixes

```
...
      ModuleNotFoundError: No module named 'pkg_resources'
      hint: This error likely indicates that `vunit-hdl@4.7.0` depends on
      `pkg_resources`, but doesn't declare it as a build dependency. If
      `vunit-hdl` is a first-party package, consider adding `pkg_resources`
      to its `build-system.requires`. Otherwise, either add it to your
      `pyproject.toml` under:
      [tool.uv.extra-build-dependencies]
      vunit-hdl = ["pkg_resources"]
      or `uv pip install pkg_resources` into the environment and re-run with
      `--no-build-isolation`.
...
```

The previous setup.py relied on a legacy setuptools environment where pkg_resources was assumed to be available. In modern, slim CI environments using uv, build isolation often lacks this deprecated module, leading to the ModuleNotFoundError. Transitioning to pyproject.toml with hatchling eliminates this legacy dependency and explicitly defines the build requirements, ensuring compatibility with modern Python packaging standards.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,8 +40,9 @@ dynamic = ["version"]
 path = "vunit/about.py"
 
 [tool.hatch.build]
-# Ensure HDL and TCL assets are included in both sdist and wheel
+# Ensure Python source, HDL and TCL assets are included in both sdist and wheel
 include = [
+  "vunit/**/*.py",
   "vunit/**/*.tcl",
   "vunit/vhdl/**",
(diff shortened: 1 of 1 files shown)


LarsAsplund:
@nbdy Are the latest updates to [pyproject.toml](https://github.com/VUnit/vunit/blob/master/pyproject.toml) sufficient for you?

PR #1167, opened 2026-02-19
fix(sim): decode simulator output on Windows

yahya-farhadi:
The old code used a simple data.decode("utf-8") which would throw a raw UnicodeDecodeError when simulator output contained
 non-UTF-8 characters (e.g., Windows legacy code pages like cp1252). The error message was something like:

`UnicodeDecodeError:'utf-8' codec can't decode byte 0xe9 in position 42: invalid continuation byte`

This was confusing because it gave no indication that the problem was with simulator output encoding, not with your source code or VUnit itself.

What the change does:

The new _decode() helper (inside check_output) replaces the old one-liner with a fallback chain of encodings:

```
encodings_to_try = (
    "utf-8",          # Try standard UTF-8 first
    "utf-8-sig",      # UTF-8 with BOM (some Windows tools emit this)
    locale.getpreferredencoding(False) or "utf-8",  # System's default encoding
    "cp1252",         # Common Western European Windows code page
)
```

The change tries each encoding in order. If all fail, it falls back to data.decode("utf-8", errors="replace") which substitutes undecodable bytes with � instead of crashing.

Before: A simulator emitting a single non-UTF-8 byte (e.g., accented character in a file path, or a copyright symbol in a vendor library message) would crash VUnit with an opaque UnicodeDecodeError — both on compile success output and on compile failure output (err.output).

After: The output is decoded gracefully using the most likely encoding, so VUnit continues normally and the user sees the actual compile pass/fail message instead of a decoding traceback.

### DIFF
--- a/vunit/sim_if/__init__.py
+++ b/vunit/sim_if/__init__.py
@@ -368,12 +368,12 @@ def check_output(command, env=None):
         which can raise UnicodeDecodeError if decoded as strict UTF-8.
         """
 
-        encodings_to_try = (
+        # Deduplicate via dict.fromkeys to avoid trying utf-8 twice
+        # when locale.getpreferredencoding() returns "UTF-8".
+        encodings_to_try = dict.fromkeys([
             "utf-8",
-            "utf-8-sig",
-            locale.getpreferredencoding(False) or "utf-8",
-            "cp1252",
-        )
+            locale.getpreferredencoding(False),
+        ])
 
         for encoding in encodings_to_try:
             try:
@@ -381,7 +381,7 @@ def check_output(command, env=None):
             except UnicodeDecodeError:
                 continue
 
-        return data.decode("utf-8", errors="replace")
+        return data.decode("utf-8", errors="backslashreplace")
 
     try:
         output = subprocess.check_output(  # pylint: disable=unexpected-keyword-arg


vunit/sim_if/__init__.py
```diff
@@ -360,14 +361,36 @@ def check_output(command, env=None):
...
+    def _decode(data: bytes) -> str:
+        """Decode tool output robustly across platforms.
+
+        Some simulators on Windows emit output in a legacy code page (e.g. cp1252),
+        which can raise UnicodeDecodeError if decoded as strict UTF-8.
+        """
+
+        encodings_to_try = (
+            "utf-8",
+            "utf-8-sig",
```
LarsAsplund: If the string is encoded with utf-8-sig, utf-8 decoding will pass without errors. You do get some extra characters though:

```
>>> "hello".encode("utf-8-sig").decode("utf-8")
    '\ufeffhello'
```

If utf-8 is decoded as utf-8-sig, it will work most of the time as the decoder recognizes that the initial expected bytes are missing.

```
>>> "hello".encode("utf-8").decode("utf-8-sig")
    'hello'
```

The exception is if the utf-8 string starts with the bytes expected at the beginning of a utf-8-sig string. Rare but fully legal.

To be strict I suggest just using utf-8. Worst case is you get som extra characters being printed

vunit/sim_if/__init__.py
```diff
@@ -360,14 +361,36 @@ def check_output(command, env=None):
...
+            locale.getpreferredencoding(False) or "utf-8",
+            "cp1252",
```
LarsAsplund: cp1252 is a byte encoding so it will never fail and the code will never reach line 384. If the encoding was something else, there will be an information loss when the string is printed. I suggest you remove this line such that line 384 is reached.

I also suggest that you use `errors="backslashreplace"`. That way, there is no information loss in the printed string. The user can copy it, convert back to bytes, and then try different decodings until the correct one is found.

PR #1173, opened 2026-02-27
Feat/vivado systemverilog

yahya-farhadi:
Fix(compile_order): SystemVerilog support in Vivado compile order plus improved subprocess output decoding on Windows

1- Handle SystemVerilog entries in Vivado compile order files and treat .sv as Verilog sources. because some IPs used in block design have systemverilog files, it should be considered unless compile order error will be shown.

2- Improve subprocess output decoding to handle legacy Windows code pages.

The old code used a simple data.decode("utf-8") which would throw a raw UnicodeDecodeError when simulator output contained
non-UTF-8 characters (e.g., Windows legacy code pages like cp1252). The error message was something like:

`UnicodeDecodeError:'utf-8' codec can't decode byte 0xe9 in position 42: invalid continuation byte
`
This was confusing because it gave no indication that the problem was with simulator output encoding, not with your source code or VUnit itself.

What the change does:

The new _decode() helper (inside check_output) replaces the old one-liner with a fallback chain of encodings:

```
encodings_to_try = (
    "utf-8",          # Try standard UTF-8 first
    locale.getpreferredencoding(False),  # System's default encoding
)
```
The change tries each encoding in order. If all fail, it falls back to data.decode("utf-8", errors="replace") which substitutes undecodable bytes with � instead of crashing.

Before: A simulator emitting a single non-UTF-8 byte (e.g., accented character in a file path, or a copyright symbol in a vendor library message) would crash VUnit with an opaque UnicodeDecodeError — both on compile success output and on compile failure output (err.output).

After: The output is decoded gracefully using the most likely encoding, so VUnit continues normally and the user sees the actual compile pass/fail message instead of a decoding traceback.

### DIFF
--- a/vunit/sim_if/__init__.py
+++ b/vunit/sim_if/__init__.py
@@ -359,9 +359,7 @@ def check_output(command, env=None):
 
         encodings_to_try = (
             "utf-8",
-            "utf-8-sig",
-            locale.getpreferredencoding(False) or "utf-8",
-            "cp1252",
+            locale.getpreferredencoding(False),
         )
 
         for encoding in encodings_to_try:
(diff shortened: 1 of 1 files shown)


yahya-farhadi:
We want to merge in our fork and this PR was a mistake.

Issue #1157, opened 2026-02-03
Compilation when --with-attributes returns empty test_list


robindust-ce:
VUnit will compile a testbench that does not set the ".example" attribute for any testcase, when called like this:
`python run.py --with-attributes=.example`

Only after compilation "No tests were run." is reported. Is this a deliberate design decision?
I think it would make sense to skip compilation if the test_list is empty.

My use case, where this would be a great benefit:
The codebase has many run.py's. In the CI/CD pipeline for an FPGA Image A I want to run all testbenches that are tagged to be related to the FPGA Image. The Makefile finds all run.py's and runs them using --with-attributes. However, the time overhead for compiling all testbenches if only a few are actually executed is quite large.

tasgomes:
Did you try with the "-m" (for minimal) flag? That should only compile the minimal list of files necessary to run the tests you enabled via wildcard (and I guess via attribute should also work but I never try).

LarsAsplund:
This is an omission. Currently, at compile time,  we don't make a distinction between "there are no testbenches to simulate because of attributes" and "no testbenches have been specified when calling the run script". The latter is the interpretation which means that we prepare for a full simulation and compile everything that has changed. A CI starting from scratch would compile everything. This can be updated.

PR #1191, opened 2026-04-01
Skip compilation when no tests match filter

robindust-ce:
Suggestion for #1157 

Compilation is skipped when no tests match the active filters (e.g. ``--with-attributes``), avoiding unnecessary work when no tests would run.

### DIFF
--- a/docs/news.d/1157.feature.rst
+++ b/docs/news.d/1157.feature.rst
@@ -1,0 +1,1 @@
+Compilation is now skipped when no tests match the active filters (e.g. ``--with-attributes``), avoiding unnecessary work when no tests would run.
--- a/tests/unit/test_ui.py
+++ b/tests/unit/test_ui.py
@@ -656,7 +656,7 @@ Listed 2 files""".splitlines()
         Test that compilation is skipped when attribute filtering results in
         no matching tests, e.g. --with-attributes=.nonexistent
         """
-        ui = self._create_ui("--with-attributes", ".nonexistent")
+        ui = self._create_ui("--with-attributes", ".nonexistent", "--no-color")
         lib = ui.add_library("lib")
         file_name = str(Path(tempdir) / "tb_skip.vhd")
         create_vhdl_test_bench_file(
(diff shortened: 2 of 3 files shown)


LarsAsplund:
I see what your coming from but I think automatically skipping compilation when there are no matching tests would undermine the purpose of the `-m` flag. The fact that the `-m` flag doesn't prevent compilation when the test list is empty is the real problem. One could argue that the `-m` flag should be default. Why compile anything that isn't needed for running the tests passing the filters? We've chosen the fail fast approach, i.e. if there are syntactic errors we should expose those ASAP. For those that prefer having `-m` as the default, it is always possible to enforce that in the run script.

To fix the current issue with the `-m` flag, I think the `or` expression on [this](https://github.com/VUnit/vunit/blob/1197afd44175df0695eb0bf8d6dd49a4af88b469/vunit/project.py#L551) line should be replaced by:

```
target_files if target_files is not None else self.get_source_files_in_order()
```

Makes sense?

robindust-ce:
Yes that makes sense and would be satisfactory for me. However, I have never considered `-m` because it does not work when component instantiations are bound at elaboration time. In this example dut.vhd (in the library "common") will not be compiled using `-m` even if my testbench depends on it.

```vhdl
library ieee;
use ieee.std_logic_1164.all;

library vunit_lib;

library common;

entity tb_component is
  generic (runner_cfg : string);
end entity;

architecture struct of tb_component is
  component dut is
    port (
      clk  : in  std_logic;
      data : out std_logic
    );
  end component;

  signal clk  : std_logic := '0';
  signal data : std_logic;
begin
  -- Component instantiation: default binding resolves to common.ram_ecc
  i_dut : dut port map (clk => clk, data => data);

  main : process
  begin
    clk <= '1';
    wait for 10 ns;
    check_equal(data, '1', "data should be driven");
    test_runner_cleanup(runner);
  end process;
end architecture;
```

crdavis12:
@robindust-ce one way I've worked around that is manually linking the dependency in python. It wouldn't be fun if you use components a lot, but isn't too bad for a handful.
https://vunit.github.io/py/vunit.html#vunit.ui.source.SourceFileList.add_dependency_on

PR #1202, opened 2026-05-27
Fix overhead in test report for large test counts

std-max:
After every test, `print_latest_status()` calls `_split()`, which iterates over all accumulated results just to display the `P=/S=/F=` counters in the progress line.
This makes per-test bookkeeping O(n) and cumulative bookkeeping O(n^2), and since it runs inside `_stdout_lock` while holding the GIL, worker threads progressively block on it.
It is particularly visible on large runs: throughput drops and CPU utilization collapses well before the end.

This fix uses three running counters and reads them directly instead of calling `_split()`.
The other two callers of `_split()` are left untouched since they need the actual result lists and only run once at the end.


std-max:
@LarsAsplund can you review this PR please?
thanks!

PR #1203, opened 2026-05-27
NVC verilog support

kernmatthias:
NVC continues to improve its verilog support.

This PR intends to extend the nvc interface in vunit to support the compilation of (system)verilog files.

I added nvc to the acceptance tests I found applicable (verilog and also PSL), but the tests still fail with the current version of NVC (1.21.0), so this PR is not quite ready to be merged.


maleiter:
This would make VUnit+NVC the first FOSS **mixed-language** simulation flow. 
That's a huge step!

tasgomes:
I have tested this variant with NVC 1.22.0 (just released) and I was able to successfully simulate a Verilog-generated HLS IP core. @kernmatthias  can you give it a try with this new version? It would be super cool to have this added to VUnit.

PR #1214, opened 2026-08-13
Publish wheels over sdist

UebelAndre:
This change updates CI/CD to publish wheels instead of source distributions to avoid the security concern of arbitrary code execution via `setup.py`


UebelAndre:
@LarsAsplund friendly ping here 🙏

PR #1220, opened 2026-09-13
VHDL to Python interface: python_pkg on a native bridge for NVC, GHDL and Questa

ru551n:
## Summary

This PR brings the VHDL-to-Python interface of the `python_pkg` branch (Lars Asplund, `VUnit/vunit@22af9848`) to `master`, with a new native implementation underneath and a few additions to its API.

**The API of the `python_pkg` branch is the interface.** `python_pkg.vhd`, `python_context.vhd`, the VHPI package and its C sources, `run_script_path`, the `check_result_t` getters and the test cases of `tb_python_pkg.vhd` are taken over; the testbench's 8 error demonstrations, which failed by design, are now negative tests checking the reported message through the mocked logger, so the whole testbench passes on NVC, GHDL and Questa.

Enable with `vu.add_vhdl_builtins()` followed by `vu.add_python()`; in VHDL, `context vunit_lib.python_context;`. `add_python()` builds whatever the simulator needs under the output path, so a run script needs nothing else, and the interpreter starts on first use, so `python_setup`/`python_cleanup` are optional.

## What is underneath

- **NVC and GHDL**: the two VHPIDIRECT applications of the branch and their `compile_*` run-script steps are replaced by the VUnit Python bridge (`vunit/python_bridge`, C in `native/`): compiled on first use on Linux/macOS and cached by a fingerprint of its inputs; prebuilt MSVC DLLs per CPython version on Windows (built in CI, shipped in releases, never committed). Loading is automatic (`--load` for NVC, the search path for GHDL, per backend).
- **Questa/ModelSim**: the same bridge through an FLI front end (`native/fli.c`), built and cached by `add_python()` like the others (the system C compiler on Linux, the MinGW compiler bundled with Questa on Windows), so the API behaves identically on NVC, GHDL and Questa. Verified on Questa Altera Starter FPGA Edition 2025.3 (Linux). The branch's FLI application is therefore not included.
- **Riviera-PRO/Active-HDL**: the branch's VHPI application, now built by `add_python()` as well and starting the interpreter on first use. Untested (no licence available); its run-script helper `compile_vhpi_application` is not needed and not included.
- The interpreter is started with `PyConfig` as the Python running VUnit (virtual environments and editable installs resolve), the GIL is taken on every entry, only scalars and fixed-size chunks cross the FFI (no length limit, same ABI on all simulators), and Python errors are reported through `python_logger` (`vunit_lib:python`), so they can be tested with `mock`/`check_only_log`.

## Added to the API, and why

- **Sessions**: a trailing, defaulted `session` parameter on every operation; a session is a VUnit object (a record around an `integer_vector_ptr_t`, created with `new_session(name)` or, from an existing identity, `new_session(id)`, with `get_id`/`name` accessors) whose name is an identity under `vunit_lib:python`; every session logs on the logger of its identity, `python_logger` being the parent. Two Python models that both define `model` or `config` can live in separate namespaces of the same interpreter.
- **More `arg`/`kwarg` value types**: `std_ulogic` (→ `bool`), `arg_unsigned`/`arg_signed` and their `kwarg` forms (any width, exact Python integers), `real_vector`, `integer_vector_ptr_t`, `integer_array_t`. The typed names for `unsigned`/`signed` exist because an `arg` overload would make a string literal argument such as `arg("Hello")` ambiguous. A value that cannot be converted (metavalues) is reported on `python_logger` and becomes an argument that raises in Python, so a call never proceeds with a missing argument. They came out of migrating a real testbench that hands 32-bit CSR counters to a Python checker.
- **Argument groups**: `kwarg("a", 1) & kwarg("b", 2)` becomes one argument (`**dict(...)`), `arg(1) & arg(2)` becomes `*(1, 2,)`, and mixed groups follow Python's rules; this lifts the 10-argument limit of `call` and lets arguments be built in loops (`null_arg` is the identity), without touching `arg_t` or `to_call_str`.
- **`integer_array_t` as NumPy arrays**, in both directions, with shape and word size preserved (`get(a, x, y)` is `a[y, x]`): bulk data such as images and weight tensors are not viable as source text.
- **More result types**: `eval_boolean`/`call_boolean`, `eval_std_ulogic`, width-checked `std_ulogic_vector`/`signed`/`unsigned` results, `call_string`, `call_real_vector`, `call_integer_vector_ptr`, `eval_integer_array`/`call_integer_array`.
- **`exec_file`**: executes a Python file (a relative name is relative to the testbench directory, `tb_path`) into the session namespace, so a verification component can load its own model from a generic without the testbench knowing that Python is involved.
- The string-argument forms of `to_call_str` and `call` are removed; arguments are always built with `arg`/`kwarg`.
- `real` values are not range-checked against single precision on the bridge (VHDL `real` is a double on these simulators); any finite value is exchanged exactly and a non-finite one is reported as a failure.

The embedded Python example gained cases for each of these, in the style of the existing ones, including a minimal verification component whose behaviour is a Python function.

## NumPy as a dependency

`numpy` is added to the package dependencies. `integer_array_t` values are exchanged as NumPy arrays and a user of the feature should not have to discover that from an error message; the runtime itself only imports NumPy lazily, so nothing changes for users who never call `add_python()`.

## Tests and CI

- `vunit/vhdl/python/run.py`: the branch's `tb_python_pkg.vhd` plus `tb_python_pkg_bridge.vhd` (75 cases for the additions), on NVC, GHDL (mcode, llvm-jit, llvm, gcc) and Questa.
- `python_bridge.yml`: Linux with CPython 3.10/3.14 × NVC 1.22.1 and every GHDL backend on the latest release; canaries (allowed to fail) on NVC built from master and the GHDL nightly builds; macOS with NVC and GHDL; the MSVC DLL build (reusable `python_bridge_dlls.yml`) and DLL-fed Windows tests on NVC and GHDL with CPython 3.10/3.12/3.14; a packaging check. `push.yml` only gains the DLL download in the release job.
- Unit tests for the build/cache logic, the generated packages and the simulator hooks; documentation in `docs/python_bridge/user_guide.rst`.

## Review

The first review round is addressed in one commit per comment (24035bef … 91ff756d, sessions completed in 6f34c380), each answered on its thread.

## Disclosure

This work was done with substantial use of an AI assistant (Claude Code, Anthropic): the design, the C, Python and VHDL code, the tests, the workflows and this description were produced with it, under my direction and review, and verified by running the test suites on the simulators listed above. Please review it with that in mind.


docs/news.d/+python_bridge.feature.rst
```diff
@@ -0,0 +1,13 @@
+VHDL testbenches can now execute Python code and call Python functions, for
```
LarsAsplund: It feel like this should be re-written to cover the larger picture. It feels like Claude (?) focused on the additions such as std_logic support.

examples/vhdl/embedded_python/tb_example.vhd
```diff
@@ -0,0 +1,770 @@
...
+      end loop;
+
+      return result;
+    end;
+
+    -- The model of the device used in some of the examples below is the Python
+    -- module device_model.py, next to this testbench
+    procedure import_device_model is
+    begin
+      import_module_from_file(tb_path(runner_cfg) & "device_model.py", "device_model");
```
LarsAsplund: VUnit provides `join(a, b)` to concatenate path segments in a way that works regardless if the path separator is present in `a` or not

examples/vhdl/embedded_python/tb_example.vhd
```diff
@@ -0,0 +1,770 @@
...
+        -- A boolean result can be used directly as a condition
+        if eval("device_model.is_busy()") then
+          info("The device reports that it is busy");
+        else
+          check_failed("The device should report that it is busy");
+        end if;
+        check_true(call_boolean("device_model.is_busy"));
+
+        -- A single bit and a vector of bits, std_logic characters on the Python side
+        check_equal(call_std_logic("device_model.parity"), '1', result("for the parity of the status"));
```
LarsAsplund: Other "typed" VUnit functions/procedures use `_std_ulogic(_vector)`

examples/vhdl/embedded_python/tb_example.vhd
```diff
@@ -0,0 +1,770 @@
...
+        exec("from my_run_script import hello_world");
+        exec("hello_world()");
+
+      elsif run("Test exec_file and import_module_from_file") then
+        -- Any Python file can be executed as it is, or imported as a module by its path.
+        -- The difference shows for a file that has a state of its own.
+
+        -- exec_file executes the file in the namespace we have been using all along,
+        -- which puts counter and bump there. The file name is relative to the run script.
+        exec_file("bump.py");
```
LarsAsplund, line 309: I think we should be explicit rather than implicit regarding the path. That follows the pattern we've used before. That way we don't have to decide what the best root is. `tb_path` is probably equally good as default.

examples/vhdl/embedded_python/tb_example.vhd
```diff
@@ -0,0 +1,770 @@
...
+        check_equal(integer'(call("bump")), 1);
+        check_equal(integer'(call("bump")), 2);
+
+        -- Executing it a second time executes counter = 0 again, which is easy to miss
+        exec_file("bump.py");
+        check_equal(integer'(call("bump")), 1, result("for the counter after re-executing the file"));
+
+        -- import_module_from_file gives us the module object, which keeps its state
+        import_module_from_file(tb_path(runner_cfg) & "bump.py", "bumper");
```
LarsAsplund: `join`

vunit/vhdl/python/src/python_pkg.vhd
```diff
@@ -0,0 +1,1064 @@
...
+
+  impure function call_real(
+    identifier : string; arg1, arg2, arg3, arg4, arg5, arg6, arg7, arg8, arg9, arg10 : arg_t := null_arg;
+    session : python_session_t := default_session
+  ) return real;
+  alias call is call_real[
+    string, arg_t, arg_t, arg_t, arg_t, arg_t, arg_t, arg_t, arg_t, arg_t, arg_t, python_session_t return real];
+
+  -----------------------------------------------------------------------------
+  -- More argument values of call
```
LarsAsplund: Collect all in one place

vunit/vhdl/python/src/python_pkg.vhd
```diff
@@ -0,0 +1,1064 @@
...
+  -- their explicit names, not as eval overloads, which keeps
+  -- check_equal(eval("17"), 17) and length(eval("[1, 2]")) unambiguous.
+  -- signed and unsigned results are only available in the procedure form
+  -- since a function cannot know the width of the result.
+  impure function eval_boolean(
+    expr : string; session : python_session_t := default_session
+  ) return boolean;
+  alias eval is eval_boolean[string, python_session_t return boolean];
+
+  impure function eval_std_logic(
```
LarsAsplund: `eval/call_std_ulogic(_vector)`

vunit/vhdl/python/src/python_pkg.vhd
```diff
@@ -0,0 +1,1064 @@
...
+  begin
+    return eval(to_call_str(identifier, arg1, arg2, arg3, arg4, arg5, arg6, arg7, arg8, arg9, arg10), session);
+  end;
+
+  -----------------------------------------------------------------------------
+  -- More argument values of call
+  -----------------------------------------------------------------------------
+  -- The Python source text of an argument value. The operation is only used to
+  -- report a value that cannot be converted, which gives an empty string.
+  function p_arg_value(value : real_vector; operation : string) return string is
```
LarsAsplund: `operation` not used in functions. Maybe an extra comment that this is intentional or rename it to something making the same point (unless these functions are called with `operation` as a keyword parameter)

vunit/vhdl/python/src/python_pkg.vhd
```diff
@@ -0,0 +1,1064 @@
...
+    end if;
+    failure(
+      python_logger,
+      operation & " cannot convert " & std_logic'image(value) & "; expected '0', '1', 'L' or 'H'"
+    );
+    return "";
+  end;
+
+  -- True when the vector has an element other than 0, 1, L and H
+  function p_has_metavalue(value : std_logic_vector) return boolean is
```
LarsAsplund: There is a standard function `is_x` for this

vunit/vhdl/python/src/python_pkg.vhd
```diff
@@ -0,0 +1,1064 @@
...
+      else
+        return args & ", " & arg;
+      end if;
+    end;
+  begin
+    return identifier & "(" & append(append(append(append(append(append(append(append(append(append(
+      "", arg1), arg2), arg3), arg4), arg5), arg6), arg7), arg8), arg9), arg10) & ")";
+  end;
+
+  procedure call(
```
LarsAsplund: call with string arguments should be remove. This was my initial approach

vunit/vhdl/python/src/python_pkg.vhd
```diff
@@ -0,0 +1,1064 @@
...
+    end if;
+    -- Resizing before negating keeps the magnitude of signed'low representable
+    magnitude := -resize(folded, value'length + 1);
+    return p_to_hex_literal(std_logic_vector(magnitude), true);
+  end;
+
+  function arg(value : real_vector) return arg_t is
+    constant text : string := p_arg_value(value, "arg");
+  begin
+    if text = "" then
```
LarsAsplund: Should this be an error instead? If the user calls `arg` and it becomes a `null_arg`, it will completely disappear from the Python call

vunit/vhdl/python/src/python_pkg.vhd
```diff
@@ -0,0 +1,1064 @@
...
+    constant text : string := p_arg_value(value, "kwarg_signed");
+  begin
+    if text = "" then
+      return null_arg;
+    end if;
+    return (kw, text);
+  end;
+
+  -----------------------------------------------------------------------------
+  -- Keyword argument groups
```
LarsAsplund: Can positional arguments be grouped in the same way?

vunit/vhdl/python/src/python_pkg.vhd
```diff
@@ -0,0 +1,1064 @@
...
+    if value.name = p_positional_arg then
+      -- A group: drop the **dict( prefix and the trailing )
+      return text(p_group_prefix'length + 1 to text'length - 1);
+    end if;
+    return value.name & "=" & value.value;
+  end;
+
+  impure function "&"(l, r : arg_t) return arg_t is
+  begin
+    if l.name = p_ignore_arg then
```
LarsAsplund: Is grouping will `null_arg` a real valid use case?

vunit/vhdl/python/src/python_ffi_pkg_bridge.vhd
```diff
@@ -0,0 +1,432 @@
...
+  -- Logger used to report Python errors, for example exceptions with their traceback
+  constant python_logger : logger_t := get_logger("vunit_lib:python");
+
+  -- A session is a named Python namespace. Sessions are created on first use
+  -- and are isolated from each other, for example
+  --
+  --   constant golden_model : python_session_t := "golden_model";
+  --
+  -- The default session is the __main__ namespace.
+  type python_session_t is array (positive range <>) of character;
```
LarsAsplund: `session_t` should probably be an object that we can build upon, i.e. a record type with a `p_data` field of type `integer_vector_ptr_t`. It is created with `new_session` function. I would like to see that its name is an identity (`id_t`) and that log calls are made to a logger with that identity. Going forward, I think we may see message passing into actors within the session namespaces. If so, actors in that namespace will be children of the session identiy. As an initial step, it would be sufficient to place the session name in such an record type but leave everything else as is.

LarsAsplund (commented):
I'm not done reviewing but this is a start. As a general comment it looks very good with nice additions. I do see that this will be a feature with a lot of development going forward. For that reason, I think most of it should be external to the VUnit core, for example in repo vunit/python_bridge. That way it can have its own life-cycle with more frequent major releases. I have started an embryo of a package system for VUnit that should be used. See how json4vhdl became a VUnit package that can be downloaded from pypi and imported as Python package (https://pypi.org/project/vunit-json-for-vhdl)

docs/news.d/+python_bridge.feature.rst
ru551n: Fixed in 24035bef: rewritten as an announcement of the feature itself.

examples/vhdl/embedded_python/tb_example.vhd
ru551n: Fixed in b9430da1: `join(tb_path(runner_cfg), "...")` everywhere in the example.

examples/vhdl/embedded_python/tb_example.vhd
ru551n: Fixed in d7a4009a: `eval_std_ulogic`, `eval_std_ulogic_vector`, `call_std_ulogic`, `call_std_ulogic_vector` and `arg`/`kwarg(std_ulogic)`, with `std_ulogic`/`std_ulogic_vector` parameter and result types.

examples/vhdl/embedded_python/tb_example.vhd
ru551n: Fixed in 4ce146ee: a relative file name is resolved from `tb_path` (in VHDL, from `runner_state`), an absolute path is used as given, and the run-script-relative resolution is gone.

examples/vhdl/embedded_python/tb_example.vhd
ru551n: Fixed in b9430da1.

vunit/vhdl/python/src/python_pkg.vhd
ru551n: Fixed in 6644db12: all `arg`/`kwarg` constructors are in one block, declarations as well as bodies.

vunit/vhdl/python/src/python_pkg.vhd
ru551n: Fixed in d7a4009a.

vunit/vhdl/python/src/python_pkg.vhd
ru551n: Fixed in 59860e3f: the parameter is dropped from the overloads that cannot fail.

vunit/vhdl/python/src/python_pkg.vhd
ru551n: Fixed in bbb18477.

vunit/vhdl/python/src/python_pkg.vhd
ru551n: Fixed in 4d5c3990: both string forms are removed; the example uses `arg(...)`.

vunit/vhdl/python/src/python_pkg.vhd
ru551n: Agreed and fixed in d33a0676: after the failure is logged, the argument becomes `__vunit__.error("<message>")`, so the call raises in Python with the same message instead of proceeding with a missing argument (only observable when the logger is mocked, but that is exactly when it matters).

vunit/vhdl/python/src/python_pkg.vhd
ru551n: Yes. Done in e4ffa820: `arg(1) & arg(2)` becomes one argument `*(1, 2,)`, a mixed group `arg(1) & kwarg("a", 2)` becomes `*(1,), **dict(a=2)`, and appending a positional argument after keyword arguments is an error, as in Python.

vunit/vhdl/python/src/python_pkg.vhd
ru551n: Yes, it is the identity element of `&`, and three things rely on it: (1) building a group in a loop, `args := null_arg; for i in regs'range loop args := args & kwarg(names(i).all, regs(i)); end loop;`, which otherwise needs a special first iteration; (2) optional arguments, since VHDL has no conditional expression the idiom is a helper that returns `null_arg` when the argument does not apply, and the caller just writes `base & optional_kwarg(...)`; (3) an empty group, `call("f", args)` with `args = null_arg`, is a valid call with no arguments, consistent with the unused slots of `call`. It also costs nothing: after the change in d33a0676 a failed conversion no longer degrades to `null_arg`, so it is only ever produced deliberately. Documented in the comment and the user guide in 9e0cdfdb.

vunit/vhdl/python/src/python_ffi_pkg_bridge.vhd
ru551n: Done in 91ff756d: `python_session_t` is a record with `p_data : integer_vector_ptr_t`, created with `new_session(name)`; the name is an identity under `vunit_lib:python` (the identity of `python_logger`), `name(session)` returns it, and the operations of a session log on the logger of that identity, a child of `python_logger` (the default session logs on `python_logger` itself). Sessions are compared by identity, so `new_session("golden")` twice is the same session.

ru551n:
> I'm not done reviewing but this is a start. As a general comment it looks very good with nice additions. I do see that this will be a feature with a lot of development going forward. For that reason, I think most of it should be external to the VUnit core, for example in repo vunit/python_bridge. That way it can have its own life-cycle with more frequent major releases. I have started an embryo of a package system for VUnit that should be used. See how json4vhdl became a VUnit package that can be downloaded from pypi and imported as Python package (https://pypi.org/project/vunit-json-for-vhdl)

Regarding the packaging, I have created #1221 as a means to support passing simulator flags (hooks into sim_if) as well as running setup hooks (compiling the C python bridge).

I have also created [vunit-python-bridge](https://github.com/ru551n/vunit-python-bridge) which will be a package similar to vunit-json-for-vhdl, We will merge this later to a repo under the VUnit banner when this feature is completed.

For the sake of context I propose that we keep reviewing here, then sync to the vunit-python-bridge repo when the review is done.  When we are satisified we can create a repo under the VUnit group and merge my personal repo to there.

ru551n:
~Regarding docs, this is quite complex feature, and a user manual would be nice (in the scope as we already have now). The question is though where it should live. Either it can live here, with an added maintanability issue (syncing with the package repo), or it can live in it's own package repo and get linked to from VUnit main repo.~

~Maybe add a list of available packages and links to their respective documentation?~

As per discussion, as much as possible should reside in the package itself.

ru551n:
~Also I feel that the examples should be in the VUnit core. They should fail with an exact error message if the package is not installed.~

As per discussion, move examples to package.

vunit/vhdl/python/src/python_ffi_pkg_bridge.vhd
ru551n: Now exactly as you describe, in 6f34c380: `python_session_t` is a record with `p_data : integer_vector_ptr_t`; `new_session(name)` creates its identity under `vunit_lib:python` and, following `new_actor`, `new_session(id : id_t)` takes an existing identity so that a session can sit anywhere in the identity tree (and actors can be its children later); `get_id(session)` and `name(session)` are the accessors; and every session, the default one included, logs on `get_logger(get_id(session))` — the earlier exception where the default session logged on `python_logger` is gone; `python_logger` is the logger of the parent identity. The Python namespace is keyed by the full identity name, so sessions with the same leaf name under different parents are distinct.

docs/python_bridge/user_guide.rst
```diff
@@ -0,0 +1,711 @@
...
+
+    ...
+
+    exec("import numpy as np");
+    result := eval("int(np.sum([1, 2, 3]))");
+
+Requirements
+------------
+
+* NVC, GHDL or Questa/ModelSim, through the VUnit Python bridge described
```
ru551n, line 42: Needs to be tested on Questa.

docs/python_bridge/user_guide.rst
```diff
@@ -0,0 +1,711 @@
...
+    expected := eval("model(x)", session => golden);
+    got := call("model", arg(x), session => fixed_point);
+
+Since ``session`` comes after up to 10 positional arguments in ``call``, it is
+normally given by name, as above. Both sessions in the example can define
+``model`` without interfering with each other.
+
+A session is a VUnit object with an :ref:`identity <id_user_guide>` of its own,
+returned by ``get_id(session)``. ``new_session(name)`` creates the identity
+``name`` under ``vunit_lib:python``, the identity of ``python_logger``, while
```
ru551n, line 131: With the packaging, the python package is probably have it's own library.

docs/python_bridge/user_guide.rst
```diff
@@ -0,0 +1,711 @@
...
+
+    constant default_logger : logger_t := get_logger(get_id(default_session));
+    constant golden_logger : logger_t := get_logger(get_id(golden));
+
+    ...
+
+    mock(golden_logger, failure);
+
+On Riviera-PRO/Active-HDL (VHPI), only the default session is supported:
+passing any other session fails with a clear error.
```
ru551n, line 165: Probably not gonna be true later.

docs/python_bridge/user_guide.rst
```diff
@@ -0,0 +1,711 @@
...
+
+When the result type is not clear from the context (for example when it is
+passed as an argument, or logged with ``info``), the explicit ``eval_<type>``
+name selects the overload:
+
+.. code-block:: vhdl
+
+    info("pi = " & to_string(eval_real("pi")));
+
+On NVC, GHDL and Questa, where the API is implemented by the VUnit Python
```
ru551n, line 259: Same, subject to change.

docs/python_bridge/user_guide.rst
```diff
@@ -0,0 +1,711 @@
...
+``real``                   ``float``
+``boolean``                ``bool``
+``string``                 ``str``
+``integer_vector``         ``list`` of ``int``
+``real_vector``            ``list`` of ``float``
+``integer_vector_ptr_t``   ``list`` of ``int``
+``std_ulogic``              ``bool``, 1 and H give ``True``, 0 and L give ``False``
+``unsigned``               ``int``, passed as ``arg_unsigned``/``kwarg_unsigned``
+``signed``                 ``int``, passed as ``arg_signed``/``kwarg_signed``
+``integer_array_t``        NumPy array, not on Riviera-PRO/Active-HDL
```
ru551n, line 327: Subject to change

ru551n:
Closing in favour of VUnit/vunit-python-bridge#1. With the package support of #1221 merged, the VHDL to Python interface no longer needs any code in VUnit core: it is now the vunit-python-bridge package, added with `vu.add_package("vunit-python-bridge", allow_setup=True)`. That pull request keeps the history of this one and includes the fixes made here since, such as the gcc fallback for the Windows builds and the 64-bit simulator check.

PR #1221, opened 2026-09-14
Let packages run a setup step and register simulator hooks

ru551n:
This adds two generic, opt-in capabilities to the package system introduced in #1166, for packages that do more than list a set of static sources.

## A setup function with a context

A package can point out a setup function with a new `setup` key in its `vunit_pkg.toml`:

```toml
[package]
requires-vunit = ">=5.0.0"
requires-vhdl = ">=2008"
library = "foo_lib"
setup = "foo.vunit_setup:setup"

[[package.sources]]
include = ["src/*.vhd"]
```

VUnit imports the module, validates the format and calls the function with a `PackageContext` once the sources listed in the manifest have been added:

```python
def setup(context):
    library = build_native_library(context.package_root / "c", context.output_path / "foo")
    context.add_source_files("foo_lib", context.output_path / "foo" / "*.vhd")
```

The context gives the package its root directory, its library, the VHDL standard used for its sources, the VUnit output path, the path of the run script, the selected simulator and methods to add libraries and source files. Errors raised by the setup function are reported as a `RuntimeError` naming the package.

Running the function takes both the package declaring it and the project allowing it: `add_package` has an `allow_setup` argument defaulting to `False`, and adding a package that declares a setup function without `allow_setup=True` is an error naming the function that would have run. A package of nothing but VHDL never runs code of its own, and the Python code a project does allow to run is then visible in its run script rather than in manifests the project never reads.

The run script path is determined as the first call stack frame outside the `vunit` package when the `VUnit` object is created, falling back to `sys.argv[0]` and then the current working directory.

## Simulator hooks

A package whose HDL code depends on something built outside the simulator, a native library or generated files, also needs the simulator to be given extra options. The setup function can register such options for a given simulator:

```python
def setup(context):
    library = build_native_library(context.package_root / "c", context.output_path / "foo")

    context.register_simulator_hooks(
        "nvc",
        run_flags=lambda simulator_interface: [f"--load={library}"],
    )

    context.register_simulator_hooks(
        "ghdl",
        elab_flags=lambda simulator_interface: [f"-Wl,{library}"],
    )
```

The registry lives in `vunit/sim_if/hooks.py` and supports extra elaboration flags, extra simulation flags and a transform of the simulation environment. A hook is called with the simulator interface instance so its result can depend on how the simulator was configured, for example on the GHDL backend. GHDL, NVC and ModelSim/Questa use the elaboration hook, all five vsim/command based interfaces (GHDL, NVC, ModelSim/Questa, Riviera-PRO, Active-HDL) use the simulation flags hook and GHDL and NVC use the environment hook. The registry is cleared when a `VUnit` object is created, so registrations do not leak between projects.

## Why

These are the generic parts needed by any package that builds something or has to reach the simulator. They are opt-in: a package without the `setup` key behaves exactly as before, a package still gets its own library through the `library` key just like `vunit-json-for-vhdl` does, and no existing behaviour or test changes.

Concretely, this is the core part needed to move the VHDL-to-Python interface of https://github.com/VUnit/vunit/pull/1220 into a separate package, as requested in its review. That package would build its native bridge in its setup function, ship its VHDL in its own library and register the `--load`/link flags the simulator needs.

## Tests and docs

* `tests/unit/test_builtins.py`: the setup function is called with the expected context (package root, library, VHDL standard, output path, run script path, simulator name, class, prefix and backend), for packages with and without sources, its delegating methods, invalid `setup` strings and types, a missing module, a missing function, a failing setup function and hook registration through the context.
* `tests/unit/test_sim_if_hooks.py`: the registry itself (per simulator, order, clearing, validation, environment copying) and mock-based tests that the GHDL, NVC, ModelSim, Riviera-PRO and Active-HDL interfaces use the hooks. No simulator needed.
* The acceptance fixture package `tests/acceptance/artificial/vunit_pkg` gets a setup function writing a marker under the output path, checked in `tests/acceptance/test_artificial.py`. The setup module lives outside the `foo` package so the existing assertion that its `__init__.py` is never imported still holds. Verified with GHDL and NVC.
* `docs/hdl_libraries.rst` gets a `Packages` section documenting `vunit_pkg.toml`, the setup function and the simulator hooks, the `add_package` docstring documents the manifest keys, `PackageContext` is added to the Python interface reference and there are news fragments.

`pytest tests/unit` (587 passed), `pytest tests/lint` and the docs build all pass.

## Also included

Four further generic additions, all with the same motivation: a package that builds something against the simulator should need nothing private from VUnit. The VHDL-to-Python interface of #1220 is the first user, and https://github.com/ru551n/vunit-python-bridge is the package built on this contract, but nothing here is specific to Python.

* **`run_script_path` in `runner_cfg`, and `check_result_t` accessors.** The test runner passes the run script through a new `"run script path"` key of `runner_cfg` and `run_pkg` exposes it as `run_script_path(runner_cfg)`. A package whose HDL calls into another language needs it so that side can resolve what the project states relative to the run script, the way the script itself does, without the testbench being told where the project lives. `check_pkg` gets `is_pass`, `get_checker`, `get_msg`, `get_log_level`, `get_line_num` and `get_file_name` on `check_result_t`, so such a package can report a check result produced by its own language binding through the VUnit checker instead of reimplementing the logging of a failed check.
* **A public `GHDLInterface.backend`.** The GHDL backend decides how a native library is bound to the design: `llvm` and `gcc` link it ahead of time at elaboration, `mcode` and `llvm-jit` load it when the simulation starts. The interface knew its backend only in a private attribute, leaving a package to find a prefix itself and repeat the detection VUnit had already done. A read-only `backend` property gives a hook, which is called with the interface, the answer directly: `elab_flags(interface)` can branch on `interface.backend`.
* **A `process_flags` hook, and `run_env` for the vsim based simulators.** The existing `run_flags` hook reaches the `vsim` command inside the generated do-file, which is where the options selecting a design and a native library belong. Some options are not honoured there because they decide how the `vsim` process itself is set up before any design is loaded; Questa's `-noautoldlibpath`, which keeps the runtime libraries Questa bundles out of the dynamic library search path, is one. `register_simulator_hooks(name, process_flags=...)` extends the command line of the `vsim` process VUnit starts, both the persistent one and the batch one, and `run_env` now also gives those interfaces the environment of that process. Both are read when the interface is constructed, so they have to be registered before that, which a package's setup function is: it runs from `add_package`, called before `main` creates the interface. The existing hooks are unchanged; the `Packages` documentation gains a table of the four hook kinds and which simulators use each.
* **`simulator_prefix` and `simulator_backend` on the context, and a public `SimulatorInterface.prefix`.** A package building something against the simulator installation itself has to know which one in its setup function, long before any simulator interface exists to ask. `context.simulator_prefix` is `find_prefix()` of the selected simulator class, `None` when no simulator was found, and `context.simulator_backend` is how the installation found there was built, which is simulator specific: `determine_backend(prefix)` for a simulator class having such a method, which is GHDL and its code generator, and `None` otherwise. The backend follows from the prefix rather than the class, so a classmethod alone would not have answered it. A hook is called with the interface rather than the context and had no way to the prefix either, so `SimulatorInterface` gets a read-only `prefix` property returning the prefix the instance was built with, the same value the setup function saw.

Tests and docs follow the same pattern as above: unit tests for each addition (`tests/unit/test_simulator_interface.py`, `tests/unit/test_ghdl_interface.py`, `tests/unit/test_sim_if_hooks.py`, `tests/unit/test_builtins.py`), the VHDL tests and news fragments that come with the `run_pkg`/`check_pkg` change, and the `Packages` section of `docs/hdl_libraries.rst` extended with the prefix and backend the context provides, the `backend` and `prefix` properties and the hook kinds table. `pytest tests/unit` is 587 passed, `pytest tests/lint` and the docs build pass, the package acceptance test passes on GHDL and NVC and `tb_magic_paths` and the 257 `check_pkg` testbench tests pass on NVC.

This work was done with substantial use of an AI assistant (Claude Code, Anthropic), under my direction and review.


docs/hdl_libraries.rst
```diff
@@ -72,6 +72,161 @@ Moreover, multiple approaches are supported for using `OSVVMLibraries <https://g
...
+   def setup(context):
+       build_dir = context.output_path / "foo"
+       build_foreign_library(context.package_root / "c", build_dir)
+       context.add_source_files("foo_lib", build_dir / "*.vhd")
+
+The context provides the package root, the library created for the package, the VHDL standard used for
+its sources, the VUnit output path, the path of the run script and the selected simulator.
+Errors raised by the setup function are reported as an error for the package.
+
+Foreign Language Interfaces
```
ru551n: Note: Remove this section, it is too specific for that use case.

docs/hdl_libraries.rst
```diff
@@ -72,6 +72,161 @@ Moreover, multiple approaches are supported for using `OSVVMLibraries <https://g
...
+     - ``VHPIDIRECT_GHDL``
+   * - NVC
+     - ``VHPIDIRECT_NVC``
+
+The two VHPIDIRECT flavours are named apart because a package cannot treat them as one: GHDL and NVC
+differ in how the attribute names the library and in how the library is loaded.
+Simulators not listed support no foreign language interface and return an empty set, so a package
+requiring one can fail with a clear message instead of generating VHDL that will not elaborate.
+
+Simulator Hooks
```
ru551n, line 152: Note: Write this section to be more generic and not explicitly mention foreign language interfaces.

tests/unit/test_simulator_interface.py
```diff
@@ -234,6 +239,29 @@ def tearDown(self):
             rmtree(self.output_path)
 
 
+class TestSupportedForeignLanguageInterfaces(unittest.TestCase):
```
ru551n: Note: This doesn't belong here, it belongs in the vunit-python-bridge

vunit/sim_if/__init__.py
```diff
@@ -94,6 +94,13 @@ def supports_vhdl_call_paths(cls):
         """
         return False
 
+    @classmethod
+    def supported_foreign_language_interfaces(cls):
```
ru551n: Note: If possible, move this to the vunit-python-bridge package

vunit/sim_if/activehdl.py
```diff
@@ -60,6 +61,13 @@ def supports_vhdl_call_paths(cls):
         """
         return True
 
+    @classmethod
+    def supported_foreign_language_interfaces(cls):
```
ru551n: Note: If possible, move this to the vunit-python-bridge package

vunit/sim_if/ghdl.py
```diff
@@ -199,6 +212,13 @@ def supports_vhdl_package_generics(cls):
         """
         return True
 
+    @classmethod
+    def supported_foreign_language_interfaces(cls):
```
ru551n: Note: If possible, move this to the vunit-python-bridge package

vunit/sim_if/modelsim.py
```diff
@@ -159,6 +160,13 @@ def supports_vhdl_call_paths(cls):
         """
         return True
 
+    @classmethod
+    def supported_foreign_language_interfaces(cls):
```
ru551n: Note: If possible, move this to the vunit-python-bridge package

vunit/sim_if/nvc.py
```diff
@@ -147,6 +148,13 @@ def supports_vhdl_call_paths(cls):
         """
         return True
 
+    @classmethod
+    def supported_foreign_language_interfaces(cls):
```
ru551n: Note: If possible, move this to the vunit-python-bridge package

vunit/sim_if/rivierapro.py
```diff
@@ -100,6 +101,13 @@ def supports_vhdl_call_paths(cls):
         """
         return True
 
+    @classmethod
+    def supported_foreign_language_interfaces(cls):
```
ru551n: Note: If possible, move this to the vunit-python-bridge package

vunit/vhdl/check/src/check.vhd
```diff
@@ -131,6 +131,36 @@ package body check_pkg is
     log(check_result);
   end;
 
+  impure function is_pass(check_result : check_result_t) return boolean is
```
ru551n, line 134: These are minor API additions that make the API cleaner towards external packages.

docs/hdl_libraries.rst
ru551n: Done: the Foreign Language Interfaces section is removed together with the method it documented.

docs/hdl_libraries.rst
ru551n: Done: the Simulator Hooks section is rewritten generically, about HDL depending on something built outside the simulator, with no mention of foreign language interfaces.

tests/unit/test_simulator_interface.py
ru551n: Done: the test is gone with the method, which now belongs to the vunit-python-bridge package.

vunit/sim_if/__init__.py
ru551n: Done: supported_foreign_language_interfaces() is dropped from core and moved to the vunit-python-bridge package.

vunit/sim_if/activehdl.py
ru551n: Done: supported_foreign_language_interfaces() is dropped from core and moved to the vunit-python-bridge package.

vunit/sim_if/ghdl.py
ru551n: Done: supported_foreign_language_interfaces() is dropped from core and moved to the vunit-python-bridge package.

vunit/sim_if/modelsim.py
ru551n: Done: supported_foreign_language_interfaces() is dropped from core and moved to the vunit-python-bridge package.

vunit/sim_if/nvc.py
ru551n: Done: supported_foreign_language_interfaces() is dropped from core and moved to the vunit-python-bridge package.

vunit/sim_if/rivierapro.py
ru551n: Done: supported_foreign_language_interfaces() is dropped from core and moved to the vunit-python-bridge package.

LarsAsplund:
I will have to look at this in detail later. Just a small comment regarding the list of source files. The reason for having support for a simple list is to speed-up the acceptance of third-party packages in companies that have more rigorous qualification processes. If the package is pure VHDL and the addition of the package in a project (`add_package`) doesn't execute any Python code, the risk of bad things happening is reduced. In cases where a package is more advanced and requires the execution of Python code, it should be explicitly allowed in the `add_package` call. If a package provides pure Python packages, these are imported like any other Python package and nothing is hidden for the user.

ru551n:
> I will have to look at this in detail later. Just a small comment regarding the list of source files. The reason for having support for a simple list is to speed-up the acceptance of third-party packages in companies that have more rigorous qualification processes. If the package is pure VHDL and the addition of the package in a project (`add_package`) doesn't execute any Python code, the risk of bad things happening is reduced. In cases where a package is more advanced and requires the execution of Python code, it should be explicitly allowed in the `add_package` call. If a package provides pure Python packages, these are imported like any other Python package and nothing is hidden for the user.

Understandable, added `allow_setup` argument to the `add_package` call, which is defaulted to `False`.

ru551n:
Agreed, and this is now implemented: `add_package` takes `allow_setup`, defaulting to `False`, so adding a package never runs Python code of the package unless the call asks for it.

A package that declares a setup function without `allow_setup=True` fails with an error naming the `module:function` it would have run, which keeps what a project allows to run visible in the run script rather than in a manifest the project never reads.

docs/hdl_libraries.rst
```diff
@@ -72,6 +72,164 @@ Moreover, multiple approaches are supported for using `OSVVMLibraries <https://g
...
+.. code-block:: toml
+   :caption: vunit_pkg.toml
+
+   [package]
+   requires-vunit = ">=5.0.0"
+   requires-vhdl = ">=2008"
+   library = "foo_lib"
+
+   [[package.sources]]
+   include = ["src/*.vhd"]
```
LarsAsplund: I'd prefer the path `hdl/src/*.vhd` as I will probably recommend a directory structure for package repos.

docs/hdl_libraries.rst
```diff
@@ -72,6 +72,164 @@ Moreover, multiple approaches are supported for using `OSVVMLibraries <https://g
...
+
+.. code-block:: toml
+   :caption: vunit_pkg.toml
+
+   [package]
+   library = "foo_lib"
+   setup = "foo.vunit_setup:setup"
+
+   [[package.sources]]
+   include = ["src/*.vhd"]
```
LarsAsplund: hdl/src/*.vhd

docs/hdl_libraries.rst
```diff
@@ -72,6 +72,164 @@ Moreover, multiple approaches are supported for using `OSVVMLibraries <https://g
...
+of the ``vsim`` process VUnit starts, which is where an option affecting how that process itself is
+set up has to go. Questa's ``-noautoldlibpath``, which keeps the bundled runtime libraries out of the
+dynamic library search path, is only honoured there.
+
+``process_flags`` and ``run_env`` are read when VUnit constructs the simulator interface, so they have
+to be registered before that happens. A package registering them from its setup function is in time:
+the setup function runs when :meth:`add_package() <vunit.ui.VUnit.add_package>` is called on the VUnit
+object, which is before :meth:`main() <vunit.ui.VUnit.main>` creates the interface.
+
+The simulator interface a hook is called with also tells how the simulator was found.
```
LarsAsplund: This sentence is a bit strange

vunit/ui/__init__.py
```diff
@@ -1277,10 +1294,53 @@ def add_vhdl_builtins(self, external=None, use_external_log=None):
...
+                             PyPI package names, equivalent to underscores.
+        :param allow_setup: Allow the setup function of the package to run. Defaults to ``False``,
+                            which makes a package declaring a setup function an error such that no
+                            Python code of a package runs unless the run script asks for it.
+
+        :example:
+
+        .. code-block:: python
+
+            VU.add_package("vunit-json-for-vhdl", allow_setup=True)
```
LarsAsplund: I think `VU.add_package("vunit-json-for-vhdl")` is sufficient since that package doesn't use a `setup` function. Also, remove the TOML file. It's not how the TOML file for that package is written and it's also more related to how a package is created. The API should only describe how a package is added/used.

docs/hdl_libraries.rst
ru551n: Done: both package examples now use `hdl/src/*.vhd`, in 087a5453d13dde3b4322b7318e41c47bf12dc881.

docs/hdl_libraries.rst
ru551n: Done: changed together with the example above, in 087a5453d13dde3b4322b7318e41c47bf12dc881.

docs/hdl_libraries.rst
ru551n: Done: reworded in 49476bf7242a6368e625ccaf80b07eb5895aae46 to "A hook can also look up where and how the simulator was installed through the simulator interface it is called with."

vunit/ui/__init__.py
ru551n: Done in 4bbc0116d2e73367e7ae71acc3a6d89e96ddb0c3: the example is now `VU.add_package("vunit-json-for-vhdl")` and the TOML example is removed. I also removed the list of `vunit_pkg.toml` keys from the docstring, since it describes how a package is created, and pointed to the Packages section for that instead.

ru551n:
The Windows CI jobs failed on this PR for a reason unrelated to its changes: MSYS2 no longer builds `jq` for the MINGW64 environment, so `ghdl/setup-ghdl` stopped with `error: target not found: mingw-w64-x86_64-jq` in its "Setup GHDL" step before any job ran. The last scheduled run on master (18 September) still passed, so master will fail the same way.

316f95b9f4e709d9074b71aea27e5f844c310c95 moves the Windows jobs to the UCRT64 environment, which MSYS2 recommends and `setup-ghdl` supports. I can move it to a separate pull request if you prefer.

